Issue #40, opened 2019-04-29
Provide checks to use pluralized variables names when accessing collections


andrueastman:
Looking at this sample
```cs
GraphServiceClient graphClient = new GraphServiceClient( authProvider );
var user = await graphClient.Users
    .Request()
    .Select( e => new {
             e.DisplayName,
             e.GivenName,
             e.PostalCode 
             })
    .GetAsync();
```

The variable name used is in singular but we are actually returning a collection of users. 

We can use this library https://www.nuget.org/packages/Pluralize.NET.Core/ to pluralize variable names when we need to.

MIchaelMainer:
Are you able to determine whether the query targets an entityset? If so, I think we want to just use the name of the entityset for the variable name. 

        <EntitySet Name="organization" EntityType="microsoft.graph.organization" />
        <EntitySet Name="groupSettings" EntityType="microsoft.graph.groupSetting" />
        <EntitySet Name="groupSettingTemplates" EntityType="microsoft.graph.groupSettingTemplate" />
        <EntitySet Name="subscribedSkus" EntityType="microsoft.graph.subscribedSku" />
        <EntitySet Name="users" EntityType="microsoft.graph.user">

var users = await graphClient.**Users**
var organization = await graphClient.**Organization**

In the case of the navigation, we just use the name of navigation property.

Eventually, we should show the reader the type of object returned. Let's not address this now.
UserCollectionPage users = await graphClient.Users

andrueastman:
Closed with PR #41

PR #41, opened 2019-05-07
Fix for incorrect pluralization of variable names

andrueastman:
This pull request proposes fixes to the pluralization of variable names generated by the snippet generator. (#40 )

Currently, the API detects that the ODataPathSegment is of a collection type and therefore blindly looks for the name in the ElementType and therefore cause the generated variable name to be always be in singular. 

However, this is incorrect as the returned name from the ODataPathSegment are suited just fine (NavigationProperty, EntitySet, NavigationLinkProperty or a KeySegment) as they will have they right pluralized form. 

The code has therefore been refactored to cater for the case of creating a new entity which will involve a POST method call to a possibly pluralized entity needing a singular version of the entity being created.

A number of unit tests have also been added to validate the fix proposed. 

Docs preview of the snippets generated by this branch can be viewed [here](https://review.docs.microsoft.com/en-us/graph/api/user-list?branch=andrueastman%2Fsnippets-preview&view=graph-rest-1.0&tabs=cs)

### DIFF
--- a/CodeSnippetsReflection.Test/SnippetsModelShould.cs
+++ b/CodeSnippetsReflection.Test/SnippetsModelShould.cs
@@ -164,7 +164,100 @@ namespace CodeSnippetsReflection.Test
             Assert.Equal("",snippetModel.RequestBody);
         }
 
+        #region Test ResponseVariableNames
+        [Fact]
+        public void SetAppropriateVariableNameForPeopleEntity()
+        {
+            //Arrange
+            var requestPayload = new HttpRequestMessage(HttpMethod.Get, "https://graph.microsoft.com/v1.0/me/people/");
+
+            //Act
+            var snippetModel = new SnippetModel(requestPayload, ServiceRootUrl, _edmModel);
+
+            //Assert that the variable name is "people" for a collection
+            Assert.Equal("people", snippetModel.ResponseVariableName);
+        }
+
+        [Fact]
+        public void SetAppropriateVariableNameForUsersList()
+        {
+            //Arrange
+            var requestPayload = new HttpRequestMessage(HttpMethod.Get, "https://graph.microsoft.com/v1.0/users");
+
+            //Act
+            var snippetModel = new SnippetModel(requestPayload, ServiceRootUrl, _edmModel);
+
+            //Assert that the variable name is "users" for the users collection
+            Assert.Equal("users", snippetModel.ResponseVariableName);
+        }
 
 
+        [Fact]
+        public void SetAppropriateVariableNameForSingleUser()
+        {
+            //Arrange
+            var requestPayload = new HttpRequestMessage(HttpMethod.Get, "https://graph.microsoft.com/v1.0/users/{id|userPrincipalName}");
+
+            //Act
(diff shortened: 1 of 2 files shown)


CodeSnippetsReflection.Test/SnippetsModelShould.cs
```diff
@@ -164,7 +164,100 @@ public void PopulateEmptyStringOnEmptyRequestBody()
...
+        public void SetAppropriateVariableNameForChildrenItemsInDrive()
+        {
+            //Arrange
+            var requestPayload = new HttpRequestMessage(HttpMethod.Get, "https://graph.microsoft.com/v1.0//drives/{drive-id}/items/{item-id}/children");
+
+            //Act
+            var snippetModel = new SnippetModel(requestPayload, ServiceRootUrl, _edmModel);
+
+            //Assert that the variable name is "children" for the collection returned
+            Assert.Equal("children", snippetModel.ResponseVariableName);
```
MIchaelMainer: GET /users/{id} returns a **User** so we call the variable `user`, this makes sense. 

GET /drives/{drive-id}/items/{item-id}/children returns a **DriveItemCollectionPage**, we should call the variable `driveItems`.

CodeSnippetsReflection.Test/SnippetsModelShould.cs
```diff
@@ -164,7 +164,100 @@ public void PopulateEmptyStringOnEmptyRequestBody()
...
+        public void SetAppropriateVariableNameForPeopleEntity()
+        {
+            //Arrange
+            var requestPayload = new HttpRequestMessage(HttpMethod.Get, "https://graph.microsoft.com/v1.0/me/people/");
+
+            //Act
+            var snippetModel = new SnippetModel(requestPayload, ServiceRootUrl, _edmModel);
+
+            //Assert that the variable name is "people" for a collection
+            Assert.Equal("people", snippetModel.ResponseVariableName);
```
MIchaelMainer: Comment only: calling the variable `people` does not demonstrate well that this is a PersonCollectionPage object. This is good enough for now but at some point, we should output `PersonCollectionPage people` instead of `var people`.

CodeSnippetsReflection/SnippetModel.cs
```diff
@@ -53,20 +53,25 @@ public SnippetModel(HttpRequestMessage requestPayload, string serviceRootUrl, IE
         /// <summary>
         ///Get a string showing the name variable to be manipulated by the segment
         /// </summary>
-        /// <param name="oDataPathSegment">The pathe segment in question</param>
+        /// <param name="oDataPathSegment">The path segment in question</param>
+        /// <returns> string to be used as return variable name in this call.</returns>
```
MIchaelMainer: Can we use the type to build the variable name?

andrueastman:
@MIchaelMainer Getting the actual return type name is possible but may not be very straightforward by inferring from the metadata as there exists a number of cases where the type does not match the what is present in the metadata. (e.g. sdk may return a page/interface like IMailFolderDeltaCollectionPage depending on the specific case).

Does it make sense for us to create a separate issue for this to be properly looked at and addressed?

MIchaelMainer:
Separate issue created.

PR #51, opened 2019-05-27
Support for Custom Query Options

### LINKED ISSUE #42: Add support for custom query options
Add support for the custom query options that maybe present in requests. 

An example request is found [here](https://docs.microsoft.com/en-us/graph/api/calendar-list-calendarview?view=graph-rest-1.0&tabs=cs).

### LINKED ISSUE #48: Accept MailTipsType enum as second parameter to GetMailTips for C#
[GetMailTips snippet](https://docs.microsoft.com/en-us/graph/api/user-getmailtips?view=graph-rest-1.0&tabs=cs#sdk-sample-code) should accept `MailTipsType` enum and not a string for C#.

Expected snippet should look like this:
```cs

var emailAddressesList = new List<String>();
emailAddressesList .Add( "danas@contoso.onmicrosoft.com" );
emailAddressesList .Add( "fannyd@contoso.onmicrosoft.com" );

// Use MailTipsType enum.
var mailTipsOptions = MailTipsType.AutomaticReplies | MailTipsType.MailboxFullStatus;

await graphClient.Me
	.GetMailTips(emailAddressesList, mailTipsOptions)
	.Request()
	.PostAsync()
```

andrueastman:
This PR proposes the following changes.

- Added support for custom query options in JS and csharp snippets (closes #42 ) View [example](https://review.docs.microsoft.com/en-us/graph/api/calendar-list-calendarview?view=graph-rest-1.0&branch=andrueastman%2Fsnippets-preview&tabs=cs)
- Added support for multiple enums in csharp (with common handling) (closes #48 ) View [here](https://review.docs.microsoft.com/en-us/graph/api/user-getmailtips?view=graph-rest-1.0&branch=andrueastman%2Fsnippets-preview&tabs=cs).
- Updated string function calls to explicitly use a declared comparison type
- Added unit tests :)

### DIFF
--- a/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
+++ b/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
@@ -33,6 +33,9 @@ namespace CodeSnippetsReflection.LanguageGenerators
                 snippetModel.SelectFieldList = snippetModel.SelectFieldList.Select(UppercaseFirstLetter).ToList();
                 var actions = CommonGenerator.GenerateQuerySection(snippetModel, languageExpressions);
 
+                //append any custom queries present
+                snippetBuilder.Append(GenerateCustomQuerySection(snippetModel));
+
                 if (snippetModel.Method == HttpMethod.Get)
                 {
                     var extraSnippet = "";
@@ -69,8 +72,8 @@ namespace CodeSnippetsReflection.LanguageGenerators
                                 foreach (var (key, jToken) in testObj)
                                 {
                                     var jsonString = JsonConvert.SerializeObject(jToken);
-                                    snippetBuilder.Append($"var {key} = ");
-                                    snippetBuilder.Append(CSharpGenerateObjectFromJson(segment, jsonString, new List<string> { key }));
+                                    snippetBuilder.Append($"var {LowerCaseFirstLetter(key)} = ");
+                                    snippetBuilder.Append(CSharpGenerateObjectFromJson(segment, jsonString, new List<string> { LowerCaseFirstLetter(key) }));
                                 }
                             }
                             snippetBuilder.Append(GenerateRequestSection(snippetModel, $"{actions}\n\t.PostAsync();"));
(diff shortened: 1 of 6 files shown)


CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -429,6 +432,80 @@ private static string GeneratePropertySectionSnippet(SnippetModel snippetModel)
...
+        /// <param name="snippetModel">Snippet model built from the request</param>
+        private static string GenerateCustomQuerySection(SnippetModel snippetModel)
+        {
+            if (!snippetModel.CustomQueryOptions.Any())
+            {
+                return string.Empty;//nothing to do here
+            }
+
+            StringBuilder stringBuilder = new StringBuilder();
+            stringBuilder.Append("var queryOptions = new List<QueryOption>()\r\n");
```
MIchaelMainer: I'm sorry to do this, but we should expand the scope of this work to include **headers** since the .Net client uses the same mechanism to add query parameters as to add custom headers . You'd have to refactor this code in the future to support setting request headers so we might as well do it now.

https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/dev/src/Microsoft.Graph.Core/Requests/BaseRequest.cs#L44
 
Line 448 really should be `stringBuilder.Append("var queryOptions = new List<Option>()\r\n");`

`GenerateCustomQuerySection(SnippetModel snippetModel)` should really be something like `GenerateCustomHeadersAndQuerySection(SnippetModel snippetModel)`. I don't like that name and you'll need to refactor anyways to accommodate HeaderOption as well.

andrueastman: @MIchaelMainer The snippet generator already supports adding of headers in CSharp. It however uses the extension method in the SDK found [here](https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/dev/src/Microsoft.Graph/Requests/Helpers/HeaderHelper.cs#L22).
This in turn generates snippets that look like [this](https://docs.microsoft.com/en-us/graph/api/user-delta?view=graph-rest-1.0&tabs=cs#sdk-sample-code-2).

Is this sufficient to address the concern about supporting headers ?

MIchaelMainer: That totally works. Thank you.

PR #55, opened 2019-07-01
Support for odata $ref in Csharp

### LINKED ISSUE #53: Add support for $ref scenarios in Csharp
Snippet generator should support $ref scenarios in C# by adding the "Reference" section to the Resources path.

The code refererenced in these issues should therefore look like this.

https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/494
https://github.com/microsoftgraph/microsoft-graph-docs/issues/4842

```await graphClient.Groups[""].Owners[""].Reference.Request().DeleteAsync();```

### PR DESCRIPTION
**Pull request by andrueastman**
This PR closes #53 by the following actions. 
- Added support for odata $ref (ReferenceSegement)
- Added test to validate the behavior

Preview fix [here](https://docs.microsoft.com/en-us/graph/api/group-delete-owners?view=graph-rest-1.0&tabs=csharp#request)

### DIFF
--- a/CodeSnippetsReflection.Test/CSharpGeneratorShould.cs
+++ b/CodeSnippetsReflection.Test/CSharpGeneratorShould.cs
@@ -267,5 +267,80 @@ namespace CodeSnippetsReflection.Test
             //Assert the snippet generated is as expected
             Assert.Equal(AuthProviderPrefix + expectedSnippet, result);
         }
+
+        [Fact]
+        //This tests asserts that we can generate snippets with $ref that adds/creates references
+        public void GeneratesSnippetsAddingReferencesToObject()
+        {
+            //Arrange
+            LanguageExpressions expressions = new CSharpExpressions();
+            const string messageJsonObject = "{\r\n  \"@odata.id\": \"https://graph.microsoft.com/v1.0/users/{id}\"\r\n}";
+            var requestPayload = new HttpRequestMessage(HttpMethod.Post, "https://graph.microsoft.com/v1.0/groups/{id}/owners/$ref")
+            {
+                Content = new StringContent(messageJsonObject)
+            };
+
+            var snippetModel = new SnippetModel(requestPayload, ServiceRootUrl, _edmModel);
+            
+
+            //Act by generating the code snippet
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: CodeSnippetsReflection.Test/CSharpGeneratorShould.cs [hunk 1]
```diff
@@ -245,5 +245,27 @@ public void GeneratesSnippetsWithCustomQueryOptions()
             //Assert the snippet generated is as expected
             Assert.Equal(AuthProviderPrefix + expectedSnippet, result);
         }
+
+        [Fact]
+        //This tests asserts that we can generate snippets with $ref operator inside it
+        public void GeneratesSnippetsWithReferenceSegmentInPath()
```
MIchaelMainer: We need to support adding references as well.

Here's an example of how references are added with a POST and the body containing a reference to a user to add. Note that the example is bad since there is a singular {id} token while there should be one for group and one for the user (I don't know if it makes a difference for snippet generation.
https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/v1.0/api/group-post-owners.md.

The POST MUST have a body that references the canonical identifier of the referenced entity.

andrueastman: @MIchaelMainer Thanks for pointing this out. Please check the updates for the same with unit tests added.

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -277,6 +285,18 @@ private static string CSharpGenerateObjectFromJson(ODataPathSegment pathSegment,
                             var newPath = path.Append(key).ToList();//add new identifier to the path
                             if (key.Contains("@odata"))
                             {
+                                var additionalDataString = $"{tabSpace}\tAdditionalData = new Dictionary<string, object>()\r\n{tabSpace}\t{{\r\n";
```
MIchaelMainer: The only time that I know that this could be a problem is when we add an @odata property. For example, I know at some point, we will automatically set the @odata.type property for some descendents of abstract types and that we need to always have that property set.

The risk is that we generate a snippet that sets the property twice. I don't think we need to worry about that now, we'll cross that when we need to.

PR #60, opened 2019-07-12
Support for Java snippet generation

### LINKED ISSUE #32: Java Code Generation

andrueastman:
This PR closes #32 with the following actions

- Support for all HTTP methods in java
- Object construction in java for inputs with request bodies
- Made operationsegment parameter generation common to Csharp and Java
- Updated readme with java examples
- Added unit tests

You can view snippets generated in the docs [here](https://review.docs.microsoft.com/en-us/graph/api/user-list?view=graph-rest-1.0&branch=andrueastman%2Fsnippets-preview&tabs=java).


CodeSnippetsReflection/LanguageGenerators/JavaGenerator.cs
```diff
@@ -0,0 +1,576 @@
...
+        /// a recursive call is made to sort out the needed object.
+        /// </summary>
+        /// <param name="pathSegment">Odata Function/Entity from which the object is needed</param>
+        /// <param name="jsonString">Json string from which the information of the object to be initialized is held</param>
+        /// <param name="path">List of strings/identifier showing the path through the Edm/json structure to reach the Class Identifier from the segment</param>
+        /// <param name="usedVarNames">List to keep track of variable names used to prevent use of the same variable name</param>
+        private static string JavaGenerateObjectFromJson(ODataPathSegment pathSegment, string jsonString, List<string> path, List<string> usedVarNames = null)
+        {
+            var stringBuilder = new StringBuilder();
+            var jsonObject = JsonConvert.DeserializeObject(jsonString);
```
darrelmiller: It is possible that JToken.Parse() is closer to what you are looking for here, however, that is not going to return a String for primitives, it is going to return a JValue.  In generate DeserializeObject is a much more expensive operation, although it may have been optimized for the scenario that you are doing here.

PR #71, opened 2019-07-29
Adding GET request for fetching all Sample Queries from JSON file

irvinesunday:
This PR adds the following features and closes tasks #67 and #65 

1.  A new `GraphExplorerSamplesService` library that:
- Provides a `DiskFileUtility` class that impements an `IFileUtility` interface which provides an abstraction for reading file contents from a given file path.

- Provides a static `SamplesService` class that handles deserializations from a given JSON string to a list of `SampleQueryModel` objects.

- A `SampleQueryModel` class with property validations that provides the base structure of a sample query object. 

2. A `SamplesService.Test` test project that provides the following test services:
- Tests for the `SampleQueryModel` properties that need validation.
-  Tests for the `SampleService` class to test different deserialization scenarios from JSON to a list of `SampleQueryModel` objects.


GraphExplorerSamplesService/SampleQueryModel.cs
```diff
@@ -0,0 +1,78 @@
...
+        public HttpMethods Method { get; set; }
+        
+        [JsonProperty(Required = Required.Always)]                   
+        public string HumanName
+        {
+            get { return _humanName; }
+            set
+            {
+                if (value.Length > 64)
+                { throw new ArgumentOutOfRangeException(nameof(HumanName), "The maximum length allowed is 64 characters."); }
```
andrueastman: [nitpick] Please format this appropriately so as not to have the entire block statement in one line.

GraphExplorerSamplesService/DiskFileUtility.cs
```diff
@@ -0,0 +1,43 @@
...
+        public async Task<string> ReadFromFile(string filePathSource)
+        {
+            try
+            {
+                using (StreamReader streamReader = new StreamReader(filePathSource))
+                {
+                    return await streamReader.ReadToEndAsync();
+                }
+            }
+            catch(DirectoryNotFoundException dirException)
```
andrueastman: As we are not really handling the exceptions in this code (we are simply re throwing them), I would think that the try catch blocks might not be really needed here.
Also, we are losing stack trace information when we do this
`throw dirException;` 
as compared to a simple 
`throw`.
This might possibly make it a bit harder to find the root cause of the failure from the response/logs.

GraphExplorerSamplesService/SamplesService.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                throw new ArgumentNullException(nameof(jsonString), "The JSON string to be deserialized cannot be null.");
+            }
+
+            try
+            {
+                var sampleQueriesList = JsonConvert.DeserializeObject<SampleQueriesList>(jsonString);
+                return sampleQueriesList;
+            }
+            catch(JsonReaderException readerException)
```
andrueastman: Similar sentiment on exception handling expressed in previous comment applies here as well.

SamplesService.Test/SamplesServiceShould.cs
```diff
@@ -0,0 +1,146 @@
...
+                    Assert.Equal("my profile", item.HumanName);
+                    Assert.Equal("/v1.0/me/", item.RequestUrl);
+                },
+                item =>
+                {
+                    Assert.Equal(Guid.Parse("F1E6738D-7C9C-4DB7-B5EC-1C92DADD03CB"), item.Id);
+                    Assert.Equal("Users", item.Category);
+                    Assert.Equal(SampleQueryModel.HttpMethods.POST, item.Method);
+                    Assert.Equal("create user", item.HumanName);
+                    Assert.Equal("/v1.0/users", item.RequestUrl);
```
andrueastman: We could probably at least assert that post body is not null here.

GraphExplorerSamplesService/DiskFileUtility.cs
irvinesunday: Agreed! This has been resolved

SamplesService.Test/SamplesServiceShould.cs
irvinesunday: Added! Including the other test points.

GraphExplorerSamplesService/SamplesService.cs
irvinesunday: Resolved

GraphExplorerSamplesService/SampleQueryModel.cs
irvinesunday: Resolved

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            try
+            {
+                // Get the file contents
+                var jsonFileContents = await _fileUtility.ReadFromFile(_configuration["SampleQueriesFilePathName"]);
+
+                // Get a list of the sample queries from the file contents
+                var sampleQueriesList = SamplesService.GetSampleQueriesList(jsonFileContents);
+
+                if (sampleQueriesList != null)
```
MIchaelMainer: `samplesQueriesList` List could be zero length. We should return NoContent for null or zero length.

GraphExplorerSamplesService/SampleQueryModel.cs
```diff
@@ -0,0 +1,82 @@
...
+        public Guid Id { get; set; }
+
+        [JsonProperty(Required = Required.Always)]
+        public string Category { get; set; }
+
+        [JsonProperty(Required = Required.Always)]
+        public HttpMethods Method { get; set; }
+        
+        [JsonProperty(Required = Required.Always)]                   
+        public string HumanName
```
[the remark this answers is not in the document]
irvinesunday: @MIchaelMainer 
Query Title is the header name used on the samples queries csv file. Graph Explorer maps this name to `humanName` as shown in the snapshot below; hence it will be expecting a similar name from the API.
@cwahome please confirm this.

MicrosoftTeams-image

GraphExplorerSamplesService/IFileUtility.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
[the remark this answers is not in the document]
irvinesunday: `IFileUtility` is a contract that is used to guarantee that whatever implements it will always return a JSON string read from a file source. This file source could be on a disk, a remote or local database on the server or even a different API. Whichever the source, now or in the future, we want to guarantee that the implementation of getting a JSON string back will not change with the change of the source of the file.

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
irvinesunday: Added the extra check.

GraphExplorerSamplesService/SampleQueriesList.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
```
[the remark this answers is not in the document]
irvinesunday: Done

GraphExplorerSamplesService/DiskFileUtility.cs
```diff
@@ -0,0 +1,25 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace GraphExplorerSamplesService
+{
+    public class DiskFileUtility : IFileUtility
```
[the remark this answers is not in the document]
irvinesunday: We might have for example, in the future, a `DatabaseFileUtility` class which will guarantee to our `GraphExplorerSamplesController` class that it implements the `IFileUtility` which reads a file from whichever database source  and returns the contents back in a consistent way without needing to change the current implementation.

GraphExplorerSamplesService/SampleQueryModel.cs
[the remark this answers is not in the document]
MIchaelMainer: That is not a good mapping of names. This should be fixed in Graph Explorer.

GraphExplorerSamplesService/DiskFileUtility.cs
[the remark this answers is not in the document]
MIchaelMainer: Excellent.

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
```diff
@@ -0,0 +1,66 @@
...
+                if (string.IsNullOrEmpty(search))
+                {
+                    // No query string values provided, return entire list of sample queries
+                    return Ok(sampleQueriesList);
+                }
+
+                // Search by Category
+                List<SampleQueryModel> sampleQueriesByCategory = sampleQueriesList.SampleQueries.FindAll(x => x.Category.ToLower() == search.ToLower());
+
+                if (sampleQueriesByCategory == null || sampleQueriesByCategory.Count == 0)
```
andrueastman: You could always also use `! sampleQueriesByCategory.Any()` instead of `sampleQueriesByCategory.Count == 0`. Just an FYI though.

MIchaelMainer:
Make sure you squash your commits.

GraphExplorerSamplesService/SampleQueryModel.cs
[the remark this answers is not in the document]
cwahome: It actually refers to the csv file headers that are being imported. The key is the header in the csv while the value is the object property it is mapping to.

Since the api is here. We will not use the csv. hence no need to change much @MIchaelMainer

PR #78, opened 2019-08-07
Adding DELETE, PUT and POST request methods, utility functions, property validations and tests

### LINKED ISSUE #69: Implement PUT request method

irvinesunday:
This PR closes #69 #64 #66 #74 #73 #70 

The following features are added in this PR:

1. `POST` request - That adds new sample queries to the list of sample queries and orders the new sample query according to the hierarchical order of the defined list of categories and the existing categories already available in the list of sample queries. Categories are now defined in a hierarchical order in a static class using a `Linked List` data structure.
2. `DELETE` request - That removes a sample query from the list of sample queries.
3. `PUT` request - That updates a sample query in the list of sample queries.
4. Adds property validation logic for `Category`, `RequestUrl` properties in the `SampleQueryModel` class.
5. Adds utility functions and code documentation in the `SamplesService` for handling serializing, adding, updating and removing sample queries.
6. Refactors the `GraphExplorerSamplesController` by extracting common code into a single function and extending the search feature to include properties: `HumanName` and `Tip`.
7. Adds an extension method that helps in formatting a JSON string into a document-readable JSON-styled string. This is called before passing a JSON string for writing to a JSON file.
8. Moves classes into respective namespaces so as to group similar functions together and to simplify and make the architecture of the library more modular.
9. Adds tests that validate the `SamplesService` utility class functions and the `SampleQueryModel` class property validations logic.
10. Updates the `SampleQueriesList` constructor to initialize a new list of `SampleQueryModel` on instantiation.


GraphExplorerSamplesService/Services/SamplesService.cs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        /// <summary>
+        /// Adds a <see cref="SampleQueryModel"/> object into an instance of a <see cref="SampleQueriesList"/>.
+        /// </summary>
+        /// <param name="sampleQueriesList">The instance of a <see cref="SampleQueriesList"/> which the <see cref="SampleQueryModel"/> 
+        /// object will be added into.</param>
+        /// <param name="sampleQueryModel">The <see cref="SampleQueryModel"/> object to be added.</param>
+        /// <returns>The instance of a <see cref="SampleQueriesList"/> with the newly added <see cref="SampleQueryModel"/> object.</returns>
+        public static SampleQueriesList AddToSampleQueriesList(SampleQueriesList sampleQueriesList, ref SampleQueryModel sampleQueryModel)
```
andrueastman: Lets not pass the model here by reference. We could cause unintended side effects when this function is modded in the future. We can just assign the id before calling this function and pass in a model with an already assigned id.

SamplesService.Test/SampleQueryModelShould.cs
```diff
@@ -1,32 +1,123 @@
...
+        {
+            // Arrange
+            SampleQueryModel sampleQueryModel = new SampleQueryModel();
+
+            // Act
+            sampleQueryModel.Category = "   Users    "; // leading and trailing whitespaces
+
+            // Assert
+            Assert.False(sampleQueryModel.Category.StartsWith(" "));
+            Assert.False(sampleQueryModel.Category.EndsWith(" "));
```
andrueastman: For this and similar tests lets add
`Assert.Equal("Users", sampleQueryModel.Category);`
just to verify that the value is exactly what we want.

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
```diff
@@ -42,25 +42,201 @@ public async Task<IActionResult> Get(string search)
...
+                    return Ok(sampleQueriesList); 
+                }
+
+                // Search sample queries
+                List<SampleQueryModel> filteredSampleQueries = sampleQueriesList.SampleQueries.
+                    FindAll(x => (x.Category != null && x.Category.ToLower().Contains(search.ToLower())) || 
+                                 (x.HumanName != null && x.HumanName.ToLower().Contains(search.ToLower())) ||
+                                 (x.Tip != null && x.Tip.ToLower().Contains(search.ToLower())));
+                
+                if (filteredSampleQueries == null || filteredSampleQueries.Count == 0)
```
andrueastman: FindAll will not return null but an empty list. Makes sense to remove `filteredSampleQueries == null`

PR #97, opened 2019-09-02
Fix for parameter order by optionality in csharp

### LINKED ISSUE #87: Parameter sorting in functions not always correct
Looking at the snippet below, found [here](https://docs.microsoft.com/en-us/graph/api/tablecollection-add?view=graph-rest-1.0&tabs=csharp#request)

```
var address = "Sheet1!A1:D5";
var hasHeaders = true;
await graphClient.Me.Drive.Items["{id}"].Workbook.Tables
	.Add(address,hasHeaders)
	.Request()
	.PostAsync();
```

The parameters for the `Add` function are in the wrong order. and should be called as so 
```
.Add(hasHeaders,address)
```

Also, the nullability of the parameters should be handled if appropiate.

andrueastman:
This PR closes #87 .

Parameters for odata actions would be sorted in the incorrect order in CSharp snippets as the SDK generator sorts the parameters by first the required parameters then by the optional parameters making some snippets not to compile.

Tests have been added to validate that CSharp snippets are ordered by the required parameters first.


CodeSnippetsReflection.Test/CommonGeneratorShould.cs
```diff
@@ -492,12 +492,81 @@ public void GetParameterListFromOperationSegment_ShouldReturnStringWithDoubleQuo
...
+            var result = CommonGenerator.GetParameterListFromOperationSegment(operationSegment, snippetModel);
 
             //Assert the string parameter is now double quoted
             Assert.Equal("\"A1:B2\"", result.First());
         }
 
+        [Fact]
+        public void GetParameterListFromOperationSegment_ShouldReturnParameterListOrderedByOptionality()
+        {
+            //Arrange
```
irvinesunday: Remove one of the duplicate comment

CodeSnippetsReflection.Test/CommonGeneratorShould.cs
```diff
@@ -492,12 +492,81 @@ public void GetParameterListFromOperationSegment_ShouldReturnStringWithDoubleQuo
...
+            //Act
+            var result = CommonGenerator.GetParameterListFromOperationSegment(operationSegment, snippetModel,"",false).ToList();
+
+            //Assert the parameters are ordered as expected
+            Assert.Equal("address", result[0]);
+            Assert.Equal("hasHeaders", result[1]);
+        }
+
+        [Fact]
+        public void GetParameterListFromOperationSegment_ShouldNullNotProvidedParametersInList()
```
irvinesunday: Can we improve the name of this test; it is unclear exactly what is being validated in this test from the name.

CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
```diff
@@ -320,5 +337,43 @@ public static IEnumerable<string> GetParameterListFromOperationSegment(Operation
...
+        /// Helper function to add parameters from list of operationSegment parameters. This is validated from a list of parameters given
+        /// from the request to prevent unnecessary population of optional params.
+        /// </summary>
+        /// <param name="initialParameterList">Initial list of parameters that we need to add parameters to</param>
+        /// <param name="edmOperationParameterList">List of parameters for the operation segment</param>
+        /// <param name="parametersProvided">List of parameters that are present in the request</param>
+        /// <param name="collectionSuffix">Suffix to add in collection parameters</param>
+        /// <returns></returns>
+        private static List<string> AddValidParameterItemsFromIEdmOperationParameterList(
+            List<string> initialParameterList ,
```
irvinesunday: [nitpick] Fix formatting - comma out of place.

CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
```diff
@@ -320,5 +337,43 @@ public static IEnumerable<string> GetParameterListFromOperationSegment(Operation
...
+            IEnumerable<IEdmOperationParameter> edmOperationParameterList, 
+            List<string> parametersProvided, 
+            string collectionSuffix)
+        {
+            foreach (var parameter in edmOperationParameterList)
+            {
+                if ((parameter.Name.ToLower().Equals("bindingparameter"))
```
irvinesunday: [nitpick] Remove unnecessary parentheses in this conditional **if** block.

CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
```diff
@@ -320,5 +337,43 @@ public static IEnumerable<string> GetParameterListFromOperationSegment(Operation
...
+        /// <returns></returns>
+        private static List<string> AddValidParameterItemsFromIEdmOperationParameterList(
+            List<string> initialParameterList,
+            IEnumerable<IEdmOperationParameter> edmOperationParameterList, 
+            List<string> parametersProvided, 
+            string collectionSuffix)
+        {
+            foreach (var parameter in edmOperationParameterList)
+            {
+                if (parameter.Name.ToLower().Equals("bindingparameter")
```
MIchaelMainer: This will cause an issue in the future when the binding parameter name doesn't match these arbitrary names (not standardized for Graph). The first parameter is always the binding parameter when bound=true. We should first check that the action or function is bound, and if it is, then skip the first parameter defined in the metadata.

andrueastman:
The behavior in the Csharp generator where the optional parameters are populated after the required parameters is found here. 
https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/29d268b7fb11ca4dcfe53c0f71f84be023f6b08e/Templates/CSharp/Base/IRequestBuilder.Base.template.tt#L110

CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
andrueastman: Thanks for pointing out that they are not standardized. This should be now resolved.

PR #118, opened 2019-09-17
Adding the permissions endpoint

### LINKED ISSUE #117: Create GraphExplorerPermissionsController endpoint with the GET method.
This will contain a GET method that will receive a request url argument, together with the http verb and scope type arguments and return a list of scopes that match the parameter values.

irvinesunday:
This PR closes #117 

Proposes:
- Creating a `GraphExplorerPermissionsController` with a `GET` action method that takes in a `ScopeRequest` object with property values bound from the query string parameters and returns a list of permission scopes for the requested parameters, or `NotFound` if the result is `null`.
- Defining a `ScopeRequest` class whose instance will be bound to the query string parameters during a `GET` request. The `RequestUrl` property has been decorated with a `Required` attribute, ensuring that the `requesturl` query string parameter in a `GET` request always has value. The rest of the model properties have default values.

The below will all be valid calls to this endpoint:
 
- This will get the scopes for the specified request url for its `GET` method and its `Application` scope type: **/api/GraphExplorerPermissions?requesturl=/security/alerts/{alert_id}**
 
- This will get the scopes for the specified request url for its `PATCH` method and its `Application` scope type: **/api/GraphExplorerPermissions?requesturl=/security/alerts/{alert_id}&method=PATCH**
 
- This will get the scopes for the specified request url for its `PATCH` method and its `DelegatedPersonal` scope type: **/api/GraphExplorerPermissions?requesturl=/security/alerts/{alert_id}&method=PATCH&scopetype=DelegatedPersonal**


cwahome:
How would the url look?

irvinesunday:
> How would the url look?

Something like this:
`/security/alerts/{alert_id}`
`/me/events/{id}/attachments/{id}`

GraphWebApi/Models/ScopeRequest.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Defines the model for a permission scope request.
+    /// </summary>
+    public class ScopeRequest
+    {
+        [Required(ErrorMessage = "The RequestUrl parameter is required in the query string.")]
+        public string RequestUrl { get; set; }
+        public string HttpVerb { get; set; } = "GET";
+        public string ScopeType { get; set; } = "Application";
```
cwahome: we'd also need to pass the scope type?

irvinesunday: Yes; there are three types of scopes provided for each `RequestUrl` under a given `HttpVerb`
However, the default scope that is inferred by the API is `Application` and the default `HttpVerb` is `GET`

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public GraphExplorerPermissionsController(IPermissionsStore permissionsStore)
+        {
+            _permissionsStore = permissionsStore;
+        }
+
+        // Gets the permission scopes for a request url
+        [HttpGet]
+        [Produces("application/json")]
+        public IActionResult GetPermissionScopes([FromQuery] ScopeRequest scopeRequest)
```
darrelmiller: Is there a reason that this is modeled as an object instead of 3 parameters?  There is no standard for serializing objects in the query string and therefore this makes a the contract dependent on implementation details.

GraphWebApi/Models/ScopeRequest.cs
darrelmiller: Currently Graph Explorer only supports Delegated Access.  It would make more sense for that to be the default.

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
irvinesunday: I think using a model makes validating the `RequestUrl` parameter easier as well as defining the default values for the `ScopeType` = "Application" and `HttpMethod` = "GET" in cases where they have not been specified in the call.
Also, capturing the parameters from the model makes calling the `GetScopes` method a one-liner, versus having multiple conditional statements checking against each query string parameter on whether they have a value or not.

GraphWebApi/Models/ScopeRequest.cs
irvinesunday: Okay. I can switch that up

irvinesunday: @darrelmiller we currently have `DelegatedWork` and `DelegatedPersonal` in our permissions; which of the two should we pick for our default?

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
darrelmiller: How about this?
```
public IActionResult GetPermissionScopes([FromQuery]string RequestUrl, [FromQuery]string method = "GET", [FromQuery]string scopeType = "Delegated")
```
From the controller action it is immediately apparent which properties have defaults and which don't, the parameter defaults are immediately visible and the call to GetScopes is still a one-liner.  As far as the Required attribute is concerned, I'm not using that is a good idea.  What status code does it return if it is missing?  Is RequestUrl="" considered present?  What if RequestUrl is not actually a valid URL?  I think if we are going to validate that value, we should validate it properly and explicitly return a 400.

GraphWebApi/Models/ScopeRequest.cs
darrelmiller: @irvinesunday  That's a great question and I don't know how to determine if someone is logged in using a personal account or a work account.  However, we should default to DelegatedWork, because that's what most people will use and what the demo account uses.  We should investigate how to determine if someone has logged in with a MSA account and use DelegatedPersonal in that case.  This will be an excellent enhancement.

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
darrelmiller: If you are interested in more background on trying to serialize objects in the query string, consider the case where we might want to pass multiple ScopeType values, so the property becomes an array of ScopeType.  Now how does that appear in the request URL.  Here's a ridiculously long thread on the topic https://github.com/OAI/OpenAPI-Specification/issues/1706

irvinesunday: I'm in agreement with your suggestion

PR #120, opened 2019-09-17
Fixing null evaluations and empty scopes to return nulls when getting permission scopes

### LINKED ISSUE #119: Fix permissions service to return nulls for missing search values or for empty scopes.
Non-existing search values in the permissions json file passed in the parameters of the GetScopes need to be evaluated to null results. The same goes for an empty scopes list for a given scope type.

irvinesunday:
This PR closes #119 

Proposes:
- Adding null-conditional operators in the LINQ query that searches the permissions and scopes json object to evaluate to `null` instead of throwing `Object reference exceptions` or `ArgumentExceptions` as is the current situation in the `GetScopes` method of the `PermissionStore.cs` class. The culprit target values are `httpVerb` and `scopeType`, throwing the above exceptions respectively if missing from the json object.
- Adding an evaluation check for returned empty scopes for a given scope type. This needs to evaluate to a null return value as well. There are some permissions with empty list of scopes in the permissions and scopes json file; these need to be accounted for.
- Adding validation tests to validate the above bug fixes.

### DIFF
--- a/GraphExplorerPermissionsService/Interfaces/IPermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Interfaces/IPermissionsStore.cs
@@ -6,6 +6,6 @@ namespace GraphExplorerPermissionsService.Interfaces
 {
     public interface IPermissionsStore
     {
-        string[] GetScopes(string requestUrl, string method = "GET", string scopeType = "Application");
+        string[] GetScopes(string requestUrl, string httpVerb = "GET", string scopeType = "Application");
     }
 }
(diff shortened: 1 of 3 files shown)


andrueastman:
👍 on validating the fix with the tests.

GraphExplorerPermissionsService/Interfaces/IPermissionsStore.cs
```diff
@@ -6,6 +6,6 @@ namespace GraphExplorerPermissionsService.Interfaces
 {
     public interface IPermissionsStore
     {
-        string[] GetScopes(string requestUrl, string method = "GET", string scopeType = "Application");
+        string[] GetScopes(string requestUrl, string httpVerb = "GET", string scopeType = "Application");
```
darrelmiller: I'm going to be annoying here and ask that we change this back to method.  httpVerb is wrong, despite what it says in Saurabh's document.

irvinesunday: Okay :-)

Issue #132, opened 2019-10-18
Return empty array for permissions with unsupported scopes.


irvinesunday:
Instead of returning null and consequently error code 404 if a permission has an unsupported scope, we would want to return an empty array instead with 200 OK. Nulls signify that the specific request url of the permission requested for does not exist in the list of permissions 
[AB#7035](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7035)

irvinesunday:
Resolved

PR #133, opened 2019-10-18
Support for processing multiple permission files

### LINKED ISSUE #131: Add support for processing multiple permission files.
We would want to load both V1.0 and Beta files into the in-memory collection of permissions for a single point of access when fetching scopes.

irvinesunday:
Closes #131`#132

This PR proposes the following:
- Processing multiple permission files (v1.0 and Beta) and extracting out unique permissions from both and storing them in memory. In this case we have 2 versions of V1.0 and 1 copy of Beta version. `ver1` of the V1.0 file contains accurate scopes for permissions contained within, though the list of permissions are incomplete and with some duplicated. `ver2` of the V1.0 file has all of the permissions from `ver1` (with some degree of inaccuracy of scopes), however, it has all of the missing permissions from `ver1` with accurate scopes. The Beta version has all scopes accurate for the `beta` endpoint, though it has incomplete list of permissions and with some duplicated as well. The intention is to load up all of the unique permissions across all three files in-memory. In total we have 1866 unique permissions between the three files.
- Adding tests to validate the modifications. 
- Adding tests permission files with a few permissions for validation purposes. Some of the permissions have been intentionally duplicated across the files.
- Replaced `moq` framework objects with concrete implementations of interface instances in the test project.


PermissionsService.Test/PermissionsService.Test.csproj
```diff
@@ -19,4 +50,10 @@
     <ProjectReference Include="..\GraphExplorerPermissionsService\GraphExplorerPermissionsService.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
MIchaelMainer: Where is Microsoft.IdentityModel.Protocols used? I think you can remove this.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -33,24 +41,35 @@ private void SeedTables()
...
+                        JToken apiPermissions = permissionsObject.First.First;
+                        
+                        foreach (JProperty property in apiPermissions)
+                        {
+                            // Remove any '(...)' from the request url and set to lowercase for uniformity
+                            string requestUrl = Regex.Replace(property.Name.ToLower(), @"\(.*?\)", string.Empty);
 
-                        // Add the request url
-                        _urlTemplateTable.Add(count.ToString(), new UriTemplate(property.Name));
+                            if (uniqueRequestUrlsTable.Add(requestUrl))
```
MIchaelMainer: Is it possible that there are duplicate input URLs with difference scopes associated with them? In that case, we would still want to add the scope despite the fact that there already exists a URL in the `uniqueRequestUrlsTable`

irvinesunday: Doing that might be counterproductive, as the duplicates might have incorrect scopes in them. The order of the files are such that the degree of accuracy cascades downwards, e.g. all the possible duplicates of `v1.0_ver1` requestUrls found in the `Beta` or `v1.0_ver2` files will be ignored, as we have a high degree of confidence that the scopes from the original requestUrl in `v1.0_ver1` are accurate. This cascades down to the subsequent files in that order, adding only unique permissions not added from the previous file(s).

GraphExplorerPermissionsService/Interfaces/IPermissionStore.cs
```diff
@@ -1,8 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-
-namespace GraphExplorerPermissionsService.Interfaces
+﻿namespace GraphExplorerPermissionsService.Interfaces
```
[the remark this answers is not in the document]
irvinesunday: Done. Will update for the rest of the project in a separate PR.

PR #158, opened 2019-11-08
Replaces default password with randomized Guid value

### LINKED ISSUE #156: Bug: Randomize the Create User password
**Describe the bug**
There's need to randomize the password provided under the User sample on the create user query

**To Reproduce**
1. Go to https://developer.microsoft.com/en-us/graph/graph-explorer/preview/ and login
2. Click on Users sample
3. Scroll down to POST query to Create User
4. See default password under Request Body

**Expected behavior**
It should be a random number with every user

**Screenshots**
image

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

irvinesunday:
closes #156 

Proposes:
- Replacing the default password in the _postBody_ template of the **Users** sample query with a randomized Guid value.
- Creating a test to validate the above functionality.

### DIFF
--- a/GraphExplorerSamplesService/Services/SamplesService.cs
+++ b/GraphExplorerSamplesService/Services/SamplesService.cs
@@ -226,11 +226,16 @@ namespace GraphExplorerSamplesService.Services
 
             if (sampleQuery != null && !string.IsNullOrEmpty(sampleQuery.PostBody))
             {
-                JObject postBodyObject = JObject.Parse(sampleQuery.PostBody);
+                try
+                {
+                    JObject postBodyObject = JObject.Parse(sampleQuery.PostBody);
 
-                postBodyObject["passwordProfile"]["password"] = Guid.NewGuid();
+                    postBodyObject["passwordProfile"]["password"] = Guid.NewGuid();
 
-                sampleQuery.PostBody = postBodyObject.ToString();
+                    sampleQuery.PostBody = postBodyObject.ToString();
+                }
+                catch
+                { }    
             }
         }
 
@@ -241,7 +246,7 @@ namespace GraphExplorerSamplesService.Services
         /// <returns>An instance of <see cref="SampleQueriesList"/> that has been preprocessed with the established business rules.</returns>
         private static SampleQueriesList SanitizeSampleQueries(SampleQueriesList sampleQueriesList)
         {
-            var orderedSampleQueries = OrderSamplesQueries(sampleQueriesList);
+            SampleQueriesList orderedSampleQueries = OrderSamplesQueries(sampleQueriesList);
 
             GenerateUniqueUserPassword(orderedSampleQueries);
 
--- a/GraphExplorerSamplesService/Services/SamplesService.cs
+++ b/GraphExplorerSamplesService/Services/SamplesService.cs
@@ -1,5 +1,6 @@
 using GraphExplorerSamplesService.Models;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


GraphExplorerSamplesService/Services/SamplesService.cs
```diff
@@ -188,30 +189,63 @@ private static int GetNewSampleQueryIndex(SampleQueriesList sampleQueriesList, S
...
+        /// needs to be replaced with a random Guid value.</param>
+        private static void GenerateUniqueUserPassword(SampleQueriesList sampleQueriesList)
+        {
+            SampleQueryModel sampleQuery = sampleQueriesList.SampleQueries.Find(s => s.Category == "Users" && s.Method == SampleQueryModel.HttpMethods.POST);
+
+            if (sampleQuery != null && !string.IsNullOrEmpty(sampleQuery.PostBody))
+            {
+                JObject postBodyObject = JObject.Parse(sampleQuery.PostBody);
+
+                postBodyObject["passwordProfile"]["password"] = Guid.NewGuid();
```
andrueastman: If a sample exists of the category users and post method but it does not have the "password" property, we will find ourselves adding the property when we shouldn't. I think we should check if the property exists before attempting to do anything.

GraphExplorerSamplesService/Services/SamplesService.cs
```diff
@@ -188,30 +189,63 @@ private static int GetNewSampleQueryIndex(SampleQueriesList sampleQueriesList, S
...
         }
+
+        /// <summary>
+        /// Replaces the default password with a random Guid value in the postBody template of the Users sample query.
+        /// </summary>
+        /// <param name="sampleQueriesList">An instance of <see cref="SampleQueriesList"/> with the Users sample query whose password in the postBody template 
+        /// needs to be replaced with a random Guid value.</param>
+        private static void GenerateUniqueUserPassword(SampleQueriesList sampleQueriesList)
+        {
+            SampleQueryModel sampleQuery = sampleQueriesList.SampleQueries.Find(s => s.Category == "Users" && s.Method == SampleQueryModel.HttpMethods.POST);
```
andrueastman: I think there is a possible future where there might be a password property in a different category. Or even on updating the user(PATCH method). 
Its something we might want to think about. But this is cool for now. 👍

GraphExplorerSamplesService/Services/SamplesService.cs
irvinesunday: Ideally what would happen is we'll have an exception if we indexed a property path that doesn't exist. A `try catch` would be more ideal I think; where we would "do nothing" when such an error is caught.

irvinesunday:
> In GE v3, there are tokens in the sample files that were replaced when the sample was loaded. Do we have that same functionality in GE v4? If not, is it on the roadmap? We really should be implementing that for this. Where do we store the samples? I'm sorry, I'm behind on the state of these things.

@MIchaelMainer the samples are stored in a JSON file in the API project directory. 
@bettirosengugi are you able to find out what other tokens in the GE v4 samples need to be replaced when the samples are loaded? Then we can plan to action them.

GraphExplorerSamplesService/Services/SamplesService.cs
andrueastman: Okay. 
Lets add the ignored comment though as the catch statement is blank.
 
```cs
                catch
                {
                    // ignored
                }
```

cwahome:
> In GE v3, there are tokens in the sample files that were replaced when the sample was loaded. Do we have that same functionality in GE v4? If not, is it on the roadmap? We really should be implementing that for this. Where do we store the samples? I'm sorry, I'm behind on the state of these things.

The work in GE v4 is handled [in this PR](https://github.com/microsoftgraph/microsoft-graph-explorer-v2/pull/287)

PR #171, opened 2019-12-09
Migrates the OpenApi service into GE API

irvinesunday:
Closes https://github.com/microsoftgraph/msgraph-openapi-introspection/issues/16

Proposes: 
- Migrating the OpenApi service project into the GE API project. 
- Updating the OpenApi spec. doc. with the endpoint definitions from the OpenApi service.

**Notes**
- Suggestions are welcomed on how we can rename the two new controllers (`HomeController` and `IndexController`) that were migrated over from the OpenApi service.


GraphWebApi/Controllers/HomeController.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.FileProviders;
+using System.IO;
+using System.Reflection;
+
+namespace GraphWebApi.Controllers
+{
+    [Route("")]
```
andrueastman: This route looks like it does the same job as the existing  route "/OpenApi.yaml". If there is no extra value in adding it i think we can remove it.

GraphWebApi/Controllers/IndexController.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Threading.Tasks;
+using GraphWebApi.Models;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Graph.OpenAPIService;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Services;
+
+namespace GraphWebApi.Controllers
+{
+    [Route("list")]
```
andrueastman: What do you think about moving this code to the `OpenApiController` and create a route called `openapi/operations` ?

irvinesunday: Makes sense

GraphWebApi/Controllers/HomeController.cs
irvinesunday: I agree.

GraphWebApi/Properties/launchSettings.json
```diff
@@ -12,7 +12,7 @@
     "IIS Express": {
       "commandName": "IISExpress",
       "launchBrowser": true,
-      "launchUrl": "api/GraphExplorerSnippets",
+      "launchUrl": "swagger/index.html",
```
andrueastman: 👍

PR #172, opened 2019-12-17
Fetches OpenApi description from URL

irvinesunday:
Closes https://github.com/microsoftgraph/msgraph-openapi-introspection/issues/18

Proposes:
- Adding a `url` query parameter that will be used to slice the Graph OpenApi service and return the OpenApi spec for the operation ids that match the respective url path.
- Changing the startup page to point to the swagger ui of the API.


OpenAPIService/OpenApiService.cs
```diff
@@ -139,6 +177,35 @@ public static OpenApiDocument CreateFilteredDocument(OpenApiDocument source, str
...
+        /// Populates the _uriTemplateTable with the Graph url paths and the _openApiOperationsTable 
+        /// with the respective OpenApiOperations for these urls paths.
+        /// </summary>
+        /// <param name="graphVersion">Version of Microsoft Graph.</param>
+        /// <param name="forceRefresh">Don't read from in-memory cache.</param>
+        private static void PopulateReferenceTables(string graphVersion, bool forceRefresh)
+        {
+            HashSet<string> uniqueUrlsTable = new HashSet<string>(); // to ensure unique url path entries in the UriTemplate table
+
+            _source = GetGraphOpenApiDocument(graphVersion, forceRefresh).GetAwaiter().GetResult();
```
andrueastman: We should change this to use `async await` pattern to have something like this.

```cs
_source = await GetGraphOpenApiDocument(graphVersion, forceRefresh);
```

GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -61,7 +69,7 @@ public class OpenApiController : ControllerBase
             }           
         }
 
-        private static void WriteIndex(string baseUrl, string graphVersion, OpenApiDocument graphOpenApi, Stream stream)
+        private void WriteIndex(string baseUrl, string graphVersion, OpenApiDocument graphOpenApi, Stream stream)
```
andrueastman: I think this could remain static.

OpenAPIService/OpenApiService.cs
```diff
@@ -126,11 +138,37 @@ public static OpenApiDocument CreateFilteredDocument(OpenApiDocument source, str
...
                     predicate = (o) => o.Tags.Any(t => tagsArray.Contains(t.Name));
                 }
             }
+            else if (url != null)
+            {                
+                /* Extract the respective Operation Id(s) that match the provided url path */
+
+                if (!_openApiOperationsTable.Any() || forceRefresh)
+                {
+                    PopulateReferenceTables(graphVersion, forceRefresh);
```
andrueastman: We should await this. 
Otherwise, we may read off an empty `_uriTemplateTable` or `_openApiOperationsTable`.

OpenAPIService/OpenApiService.cs
```diff
@@ -139,6 +177,35 @@ public static OpenApiDocument CreateFilteredDocument(OpenApiDocument source, str
             return predicate;
         }
 
+        /// <summary>
+        /// Populates the _uriTemplateTable with the Graph url paths and the _openApiOperationsTable 
+        /// with the respective OpenApiOperations for these urls paths.
+        /// </summary>
+        /// <param name="graphVersion">Version of Microsoft Graph.</param>
+        /// <param name="forceRefresh">Don't read from in-memory cache.</param>
+        private static async void PopulateReferenceTables(string graphVersion, bool forceRefresh)
```
andrueastman:
```suggestion
        private static async Task PopulateReferenceTables(string graphVersion, bool forceRefresh)
```

OpenAPIService/OpenApiService.cs
andrueastman: 👍

PR #188, opened 2020-02-06
Adding support for OpenApi doc. inline referencing

irvinesunday:
Proposes
- Bumping up version of package `Microsoft.OpenApi.Readers` to _1.2.0-preview_ to add support for inline referencing of OpenApi doc. 
- Adding direct reference to `Microsoft.OpenApi.OData.Readers` library so as:
  - to get the most recent bug fixes;
  - enable us get direct method reference in the `Microsoft.OpenApi.OData.Readers` library


irvinesunday:
Closing to fix breaking project references

PR #190, opened 2020-02-20
Adding support for OpenApi doc. inline referencing of PowerPlatform style

### LINKED ISSUE #184: Add navigation property descriptions to OpenAPI

### LINKED ISSUE #186: Enable inlining of OpenAPI descriptions for PowerPlatform style

irvinesunday:
Closes #186 

Proposes

- Bumping up version of package `Microsoft.OpenApi.Readers` to `1.2.0-preview` to add support for inline referencing of OpenApi descriptions of PowerPlatform style.
- Adding direct reference to `Microsoft.OpenApi.OData.Readers` library so as:
  - to get the most recent bug fix that closes #184 ;
  - enable us get direct method reference in the `Microsoft.OpenApi.OData.Readers` library


darrelmiller:
I'm not sure I'm ok with pulling in the OData.Reader source. Let's talk.

OpenAPIService/OpenApiService.cs
```diff
@@ -337,9 +337,12 @@ private static OpenApiDocument FixReferences(OpenApiDocument document)
         {
             // This method is only needed because the output of ConvertToOpenApi isn't quite a valid OpenApiDocument instance.
             // So we write it out, and read it back in again to fix it up.
-            var sb = new StringBuilder();
-            document.SerializeAsV3(new OpenApiYamlWriter(new StringWriter(sb)));
-            var doc = new OpenApiStringReader().Read(sb.ToString(), out var diag);
+
+            var outputString = new StringWriter(CultureInfo.InvariantCulture);
+            var writer = new OpenApiYamlWriter(outputString, new OpenApiWriterSettings { ReferenceInline = ReferenceInlineSetting.InlineLocalReferences });
```
darrelmiller: Inlining references should only happen for PowerPlatform style, not for PowerShell or Plain.

GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -43,6 +43,8 @@ public OpenApiController(IConfiguration configuration)
                     return new BadRequestResult();
                 }
 
+                OpenApiService.openApiStyle = style;
```
andrueastman: I think it would be better to pass in the style as a parameter to where its needed rather than using the static class variable. 
Potentially, a race condition could occur if this endpoint is called with different styles in parallel modifying/reading the same `OpenApiService.openApiStyle` variable.

OpenAPIService/OpenApiService.cs
```diff
@@ -313,7 +314,7 @@ private static OpenApiDocument Clone(OpenApiDocument subsetOpenApiDocument)
             return reader.Read(stream, out OpenApiDiagnostic diag);
         }
 
-        private static async Task<OpenApiDocument> CreateOpenApiDocumentAsync(Uri csdlHref)
+        private static async Task<OpenApiDocument> CreateOpenApiDocumentAsync(Uri csdlHref, OpenApiStyle style = new OpenApiStyle())
```
andrueastman: Since OpenApiSyle is an enum,i think it would be great if we explicitly set a default value rather than using the new keyword

PR #219, opened 2020-04-06
Provides a modular means of styling OpenAPI documents

### LINKED ISSUE #213: Update OpenApiService to specify default query param values with the default values defined for the respective styles
The OpenApiService need to be updated to capture the below design requirement:

https://microsoft.sharepoint-df.com/teams/GraphTooling/_layouts/15/Doc.aspx?sourcedoc=%7B986B15D3-9AEA-4B94-BFB6-3A29841705FF%7D&file=OpenApiService%20Style%20Design%20Doc..docx&nav=eyJjIjoxMTk5NDcxNTA2fQ&action=default&mobileredirect=true

irvinesunday:
Closes #213 

Proposes:
- Creating a `StyleOptions` class that defines the different styling options for an OpenAPI document.
- Creating an instance of this class for each call to the `OpenApiController` with the supplied query params (with values or otherwise - only the `OpenApiStyle` has a default value) and passing this instance as a method argument when fetching and transforming the OpenAPI document.
- Fixing the `WriteIndex()` method in the `OpenApiController` to be able to properly render OpenAPI descriptions for _tags_; and updating the same method with new query params so as to be able to appropriately capture the query param values in the subsequent _operationIds_.
- Refactoring the `OpenApiService` method `CreateOpenApiDocumentAsync()` so that it is able to dynamically set `PowerShell` specific settings whether they are defined or not in the incoming `styleOptions` parameter value. [This has been specified as an optional value so as not to break any existing client code that uses this method].

### DIFF
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -121,13 +121,13 @@ namespace GraphWebApi.Controllers
 
             foreach (var item in indexSearch.Index)
             {
-
-                var target = $"{baseUrl}/openapi?tags={item.Key.Name}&openApiVersion=3&graphVersion={graphVersion}";
-                sw.WriteLine($"<li>{item.Key.Name} [<a href='../../openapi?tags={target}'>OpenApi</a>]   [<a href='/swagger/index.html#url={target}'>Swagger UI</a>]</li>");
+                var target = $"{baseUrl}/openapi?tags={item.Key.Name}&openApiVersion={openApiVersion}&graphVersion={graphVersion}&format={format}&style={style}";
+                sw.WriteLine($"<li>{item.Key.Name} [<a href='{target}'>OpenApi</a>]   [<a href='/swagger/index.html#url={target}'>Swagger UI</a>]</li>");
                 sw.WriteLine("<ul>");
                 foreach (var op in item.Value)
                 {
-                    sw.WriteLine($"<li>{op.OperationId}  [<a href='../../openapi?operationIds={op.OperationId}'>OpenAPI</a>]</li>");
+                    sw.WriteLine($"<li>{op.OperationId}  [<a href='../../openapi?operationIds={op.OperationId}&openApiVersion={openApiVersion}&graphVersion={graphVersion}" +
+                        $"&format={format}&style={style}'>OpenAPI</a>]</li>");
                 }
                 sw.WriteLine("</ul>");
             }
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -56,7 +56,7 @@ namespace GraphWebApi.Controllers
 
                 OpenApiDocument source = await OpenApiService.GetGraphOpenApiDocumentAsync(graphUri, forceRefresh, styleOptions);
 
-                var subsetOpenApiDocument = OpenApiService.CreateFilteredDocument(source, title, graphVersion, predicate);
+                var subsetOpenApiDocument = OpenApiService.CreateFilteredDocument(source, title, styleOptions.GraphVersion, predicate);
 
                 subsetOpenApiDocument = OpenApiService.ApplyStyle(styleOptions, subsetOpenApiDocument);
 
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -4,6 +4,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Services;
 using OpenAPIService;
+using OpenAPIService.Common;
 using System.Globalization;
 using System.IO;
 using System.Threading.Tasks;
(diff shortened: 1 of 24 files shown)


GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -1,9 +1,10 @@
-﻿using GraphWebApi.Models;
+using GraphWebApi.Models;
```
andrueastman, line 1: Lets add the Copyright header

OpenAPIService/OpenApiService.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Microsoft.OData.Edm.Csdl;
+using Microsoft.OData.Edm.Csdl;
```
andrueastman, line 1: Lets add the Copyright header

OpenAPIService/OpenApiService.cs
```diff
@@ -218,15 +223,15 @@ private static async Task PopulateReferenceTablesAync(string graphUri, bool forc
...
         {
             var stream = new MemoryStream();
             var sr = new StreamWriter(stream);
             OpenApiWriterBase writer;
 
-            if (format == "yaml")
+            if (styleOptions.OpenApiFormat == Constants.OpenApiConstants.Format_Yaml)
             {
-                if (style == OpenApiStyle.PowerPlatform)
+                if (styleOptions.Style == OpenApiStyle.PowerPlatform)
```
darrelmiller, line 234: Once we have converted Style to the options class we shouldn't use Style again.

OpenAPIService/Common/OpenApiStyleOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+        public string OpenApiFormat { get; private set; }
+        public bool InlineLocalReferences { get; private set; } = false;
+        public bool EnablePagination { get; private set; } = false;
+        public bool EnableDiscriminatorValue { get; private set; } = false;
+        public bool EnableDerivedTypesReferencesForRequestBody { get; private set; } = false;
+        public bool EnableDerivedTypesReferencesForResponses { get; private set; } = false;
+
+        public OpenApiStyleOptions(OpenApiStyle style, string openApiVersion = null, string graphVersion = null, string openApiFormat = null)
+        {
+            Style = style;
```
darrelmiller, line 24: I don't think we should store Style in StyleOptions.  Having options derive from style, but then still using style as a parameter is confusing.

OpenAPIService/OpenApiService.cs
```diff
@@ -331,18 +336,22 @@ private static OpenApiDocument Clone(OpenApiDocument subsetOpenApiDocument)
...
             var settings = new OpenApiConvertSettings() {
-                 EnableKeyAsSegment = true,
-                 EnableOperationId = true,
-                 PrefixEntityTypeNameBeforeKey =true,
-                 TagDepth = 2                  
+                EnableKeyAsSegment = true,
+                EnableOperationId = true,
+                PrefixEntityTypeNameBeforeKey = true,
+                TagDepth = 2,
+                EnablePagination = styleOptions == null ? false : styleOptions.EnablePagination,
```
darrelmiller, line 351: We should making pagination always true.  It is just extra ignored metadata if a tool doesn't support it.

OpenAPIService/OpenApiService.cs
```diff
@@ -331,18 +336,22 @@ private static OpenApiDocument Clone(OpenApiDocument subsetOpenApiDocument)
...
+                EnableDiscriminatorValue = styleOptions == null ? false : styleOptions.EnableDiscriminatorValue,
```
darrelmiller, line 352: Same for discriminator.  This isn't a style thing, it is related to having an accurate representation of OData

irvinesunday: What about showing the derived types references for base types $refs in the responses and requestBody, do we want to enable this for all the `OpenApiStyle` options?

OpenAPIService/Common/OpenApiStyleOptions.cs
irvinesunday: I don't see how the class would make sense without a `Style` property. It is the key identifier property for an instance of this class; it makes the class fully descriptive. What other option did you have in mind?

OpenAPIService/OpenApiService.cs
irvinesunday: This specific example has been remedied in a subsequent [PR](https://github.com/microsoftgraph/microsoft-graph-explorer-api/pull/225); however there are other instances where we make decisions based on the style. e.g. [here](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/d6dcb6eabc1b0c9b2fd4e297b37bb0c3c3f1fbc5/OpenAPIService/OpenApiService.cs#L311) and [here](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/d6dcb6eabc1b0c9b2fd4e297b37bb0c3c3f1fbc5/OpenAPIService/OpenApiService.cs#L297) and as in the above linked PR, where we perform some actions based on whether the style is `GEAutocomplete` for the Graph Explorer autocomplete feature.

My thoughts: If we removed the Style property, then we'd lose the flexibility of making such decisions dynamically based on the inherent instance property `Style` and resort to either using the query parameter style or, even worse, have to add properties in the modal `StyleOptions.cs` class to act as decision flags.

PR #224, opened 2020-04-14
Return all permissions

### LINKED ISSUE #221: Add support to return all permissions
Allow DevX API to return all permissions when requested by a client, e.g Graph Explorer.

thewahome:
Returns all permissions when the permissions endpoint is called with no parameters

Closes #221

### DIFF
--- a/GraphExplorerPermissionsService/Interfaces/IPermissionStore.cs
+++ b/GraphExplorerPermissionsService/Interfaces/IPermissionStore.cs
@@ -9,6 +9,6 @@ namespace GraphExplorerPermissionsService.Interfaces
 {
     public interface IPermissionsStore
     {
-        List<ScopeInformation> GetScopes(string requestUrl, string method = "GET", string scopeType = "DelegatedWork");
+        List<ScopeInformation> GetScopes(string requestUrl = null, string method = "GET", string scopeType = "DelegatedWork");
     }
 }
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -38,7 +38,7 @@ namespace GraphExplorerPermissionsService
 
             SeedTables();
         }
-        
+
         private void SeedTables()
         {
             try
@@ -49,12 +49,12 @@ namespace GraphExplorerPermissionsService
                  * with confidence since permissions are already seeded. */
 
                 SeedPermissionsTables();
-                SeedScopesInfoTables();                
+                SeedScopesInfoTables();
             }
             catch
             {
                 // Do nothing; the tables will just be empty
-            }                  
+            }
         }
 
         /// <summary>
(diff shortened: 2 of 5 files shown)


GraphExplorerPermissionsService/Interfaces/IPermissionStore.cs
```diff
@@ -9,6 +9,6 @@ namespace GraphExplorerPermissionsService.Interfaces
 {
     public interface IPermissionsStore
     {
-        List<ScopeInformation> GetScopes(string requestUrl, string method = "GET", string scopeType = "DelegatedWork");
+        List<ScopeInformation> GetScopes(string requestUrl = null, string method = "GET", string scopeType = "DelegatedWork");
```
irvinesunday, line 12: Since param `method` is not required for this use case of fetching ALL scopes, you can initialize it to `null` as well.

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
```diff
@@ -26,12 +26,11 @@ public GraphExplorerPermissionsController(IPermissionsStore permissionsStore)
         // Gets the permission scopes and info for a request url
```
irvinesunday, line 26: This comment will need updating, since the functionality of the method has expanded

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
```diff
@@ -26,12 +26,11 @@ public GraphExplorerPermissionsController(IPermissionsStore permissionsStore)
...
         [HttpGet]
         [Produces("application/json")]
-        public IActionResult GetPermissionScopes([FromQuery]string requestUrl, [FromQuery]string method = "GET", [FromQuery]string scopeType = "DelegatedWork")
+        public IActionResult GetPermissionScopes([FromQuery]string requestUrl = null, [FromQuery]string method = "GET", [FromQuery]string scopeType = "DelegatedWork")
```
irvinesunday, line 29: Likewise, the `method` param here can be initialized to `null` since it is not required when fetching ALL scopes.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -125,43 +125,65 @@ private void SeedScopesInfoTables()
...
+                    {
+                        var result = (JArray)scope;
+
+                        string[] permissions = result.FirstOrDefault()?
+                        .SelectToken(scopeType)?
+                        .Select(s => (string)s)
+                        .ToArray();
+
+                        if (permissions != null)
+                            permissionsList.AddRange(permissions);
```
irvinesunday, line 156: I think it's prudent to enclose this line of code within a block of curly braces. Best practice 🙂

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -125,43 +125,65 @@ private void SeedScopesInfoTables()
...
+                    }
+                    if (permissionsList.Count > 0)
+                        scopes = permissionsList.Distinct().ToArray();
```
irvinesunday, line 159: Same applies here.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -125,43 +125,65 @@ private void SeedScopesInfoTables()
...
                 }
+                else
+                {
```
irvinesunday, line 162: If you update the `method` arg. to have a default value of `null` you can add a nullability check here.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -49,12 +49,12 @@ private void SeedTables()
...
+                SeedScopesInfoTables();
             }
             catch
             {
                 // Do nothing; the tables will just be empty
-            }                  
+            }
         }
 
         /// <summary>
```
irvinesunday, line 60: Update the summary comment to reflect the expanded functionality of the method.

GraphWebApi/wwwroot/OpenApi.yaml
```diff
@@ -233,7 +233,7 @@ paths:
       parameters:
```
irvinesunday, line 233: If you update the param `method = null`, update the corresponding required value here to be `false`

PermissionsService.Test/PermissionsStoreShould.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Extensions.Configuration;
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
irvinesunday, line 12: Remove unnecessary using directive.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -125,43 +125,65 @@ private void SeedScopesInfoTables()
...
-                    .Select(s => (string)s)
-                    .ToArray();
+                    if (resultMatch == null)
+                    {
+                        return null;
+                    }
+                    resultValue = (JArray)_scopesListTable[int.Parse(resultMatch.Key)];
 
-                List<ScopeInformation> scopesList = null;
+                    string[] permissions = resultValue.FirstOrDefault(x => x.Value<string>("HttpVerb") == method)?
```
ddyett, line 175: >string[] permissions [](start = 20, length = 20)

why not just assign scopes directly?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -125,43 +125,65 @@ private void SeedScopesInfoTables()
...
+                        .SelectToken(scopeType)?
+                        .Select(s => (string)s)
+                        .ToArray();
+
+                    scopes = permissions;
+                }
+
+                List<ScopeInformation> scopesList = new List<ScopeInformation>();
```
ddyett, line 183: >List<ScopeInformation> scopesList [](start = 16, length = 33)

declared before the if but the value is not used outside of the if statement.

GraphExplorerPermissionsService/Interfaces/IPermissionStore.cs
thewahome: Defaulting it like this ensures we do not need to do null checks

GraphExplorerPermissionsService/Services/PermissionsStore.cs
thewahome: Because 
1. the scopes list fetch for all scopes may contain duplicates which i need to remove before adding to the array
2. You can not add to an array directly simply. so it becomes easier to add to a list, then convert to an array

GraphExplorerPermissionsService/Services/PermissionsStore.cs
thewahome: I do not understand the use for the method... help

GraphExplorerPermissionsService/Services/PermissionsStore.cs
ddyett: scopes is a string[] and permissions is a string[].  below it the only action is to assign scopes=permissions.  i'm not questioning the query, i'm asking why the permissions variable is needed.

---
In reply to: [408306119](https://github.com/microsoftgraph/microsoft-graph-explorer-api/pull/224#discussion_r408306119) [](ancestors = 408306119)

thewahome: I see... I was not aware you were looking at the second assignment.. My bad.. Fixed it. Nice catch

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: I think "Retrieves permission scopes." would suffice.

Issue #204, opened 2020-03-17
Move samples queries file into blob storage


irvinesunday:
The _SamplesQueries.json_ file is currently stored in a folder directory. The problem with this is:
- Updating the web app will overwrite any changes made to the _SamplesQueries.json_
- If we decide to scale out, we will have two copies of the file.
- We need to support localization of these resources, so some refactoring/normalization of this data is needed.

This file needs to be moved to an Azure blob storage.

[AB#4592](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4592)

ddyett:
Part of AB#4592

Issue #198, opened 2020-02-28
Add Accept-Language support for Samples and CapabilitiesAPI/PermissionsAPI


darrelmiller:
The DevX API should return descriptive content that respects the requested language.

[AB#4592](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4592)

ddyett:
Part of AB#4592

PR #228, opened 2020-04-20
Fetches localized samples and permissions files from Azure Blob Storage

irvinesunday:
Closes: #204 , #198 , [AB#4592](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4592)

**Proposes**:
- Creating a utility class that is responsible for fetching blobs from an Azure storage account. The `Write` method implementation of the interface inherited by this class is not implemented (and will further be refactored out in a subsequent issue/branch/PR).
- Adding `Accept-Language` capabilities for the Samples and Permissions endpoints which will enable fetching of the respective localized files.
- Updating and refactoring Permissions tests to integrate and validate the above changes.
- Refactoring Samples and Permissions endpoints and the `PermissionStore.cs` class to integrate with and consume the above changes.
- Updating the `appsettings.json` to incorporate settings relevant for connecting and fetching files from the Azure blob storage. 
- Refactoring the `GetScopes` method of the  `IPermissionStore.cs` interface and the `PermissionStore.cs` class accordingly to be appropriately descriptive in their method parameter definitions. Likewise the OpenAPI description for the calling endpoint - `PermissionsController.cs`.

**Notes**:
- The Azure blob storage account connection string needs to be supplied in the KeyVault during deployment.
- Local caching of the samples and permissions have not been implemented; this will need a separate issue/branch/PR process.
- Business requirements has since required that the API only provides read access to both the Samples and Permissions endpoint. This action requires a separate issue/branch/PR. In the meantime, any references to write/delete actions in the Samples endpoint have been made ineffective but the code is still intact.


FileService/Common/FileServiceHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// </summary>
+        /// <param name="filePathSource">The complete file path.</param>
+        /// <param name="directoryName">The expected directory name.</param>
+        /// <param name="fileName">The expected file name.</param>
+        public static void RetrieveFilePathSourceValues(string filePathSource, out string directoryName, out string fileName)
+        {
+            directoryName = null;
+            fileName = null;
+
+            if (filePathSource.IndexOf(FileServiceConstants.DirectorySeparator) > 0)
```
zengin, line 26: >filePathSource [](start = 16, length = 14)

This can throw null pointer exception. "public" methods should validate their input. You can either change access type, or validate input.

FileService/Common/FileServiceHelper.cs
```diff
@@ -0,0 +1,106 @@
...
+    /// </summary>
+    public static class FileServiceHelper
+    {        
+        /// <summary>
+        /// Retrieves the directory and file names from a given a file path source.
+        /// </summary>
+        /// <param name="filePathSource">The complete file path.</param>
+        /// <param name="directoryName">The expected directory name.</param>
+        /// <param name="fileName">The expected file name.</param>
+        public static void RetrieveFilePathSourceValues(string filePathSource, out string directoryName, out string fileName)
```
zengin, line 21: >out string directoryName, out string fileName [](start = 79, length = 45)

consider using tuple return type instead of multiple out parameters: [https://docs.microsoft.com/en-us/dotnet/csharp/tuples](https://docs.microsoft.com/en-us/dotnet/csharp/tuples)

FileService/Services/AzureBlobStorageUtility.cs
```diff
@@ -0,0 +1,70 @@
...
+                    throw new IOException($"The '{_containerName}' container doesn't exist.");
+                }
+            }
+
+            throw new IOException("Failed to connect to the blob storage account.");
+        }
+
+        public async Task WriteToFile(string fileContents, string filePathSource)
+        {
+            // Not implemented
```
zengin, line 67: >// Not implemented [](start = 12, length = 18)

`throw new NotImplementedException();` so that you can see if anyone ever calls this no-op method.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -98,9 +103,17 @@ private void SeedPermissionsTables()
         /// <summary>
         /// Populates the delegated and application scopes information tables. 
         /// </summary>
-        private void SeedScopesInfoTables()
+        private void SeedScopesInfoTables(string localCode = null)
```
zengin, line 106: >localCode [](start = 49, length = 9)

locale?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -222,6 +249,10 @@ public List<ScopeInformation> GetScopes(string requestUrl = null, string method
 
                 return null;
             }
+            catch (ArgumentNullException exception)
+            {
+                throw exception;
+            }
```
zengin, line 255: Q: What is the purpose of this change? It looks like the behavior is the same without the change as well.

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
```diff
@@ -288,9 +299,9 @@ public async Task<IActionResult> DeleteSampleQueryAsync(string id)
                 // Get the serialized JSON string of the list of sample queries
                 string newSampleQueriesJson = SamplesService.SerializeSampleQueriesList(sampleQueriesList);
 
-                // Save the JSON string to the source file
-                await _fileUtility.WriteToFile(newSampleQueriesJson, _queriesFilePathSource);
-                                
+                // Disabled functionality
+                // await _fileUtility.WriteToFile(updatedSampleQueriesJson, _queriesFilePathSource);
```
zengin, line 303: Remove it altogether?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Yes. Thanks

FileService/Common/FileServiceHelper.cs
irvinesunday: Nice catch

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
irvinesunday: Yes. That's the plan. I want to remove all disabled functionalities in a subsequent issue/branch/PR so as not to have too many unrelated file changes/deletions within this PR.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: We would want to capture the error thrown by this [exception](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/59872326cf184ea87e9b95893491c51049a550e9/GraphExplorerPermissionsService/Services/PermissionsStore.cs#L190)

FileService/Services/AzureBlobStorageUtility.cs
irvinesunday: Resolved

FileService/Common/FileServiceHelper.cs
irvinesunday: Thanks for the tip. Resolved

GraphExplorerPermissionsService/Services/PermissionsStore.cs
zengin: I see. `ArgumentNullException` inherits from `ArgumentException`.

I thought of them as two separate exceptions, so rethrowing would be the same as leaving it out. But the way it is now, it will be captured by the next catch statement, and you don't want that.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -61,13 +65,14 @@ private void SeedTables()
         /// Populates the template table with the request urls and the scopes table with the permission scopes.
         /// </summary>
         private void SeedPermissionsTables()
-        {
+        {
```
zengin, line 68: NIT: In order to avoid whitespace changes similar to this to show up in PRs, you can consider using this extension, which optionally clears trailing whitespaces on save: https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer

FileService/Services/AzureBlobStorageUtility.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        public async Task<string> ReadFromFile(string filePathSource)
+        {
+            FileServiceHelper.CheckArgumentNullOrEmpty(filePathSource, nameof(filePathSource));
+
+            if (filePathSource.IndexOf(FileServiceConstants.DirectorySeparator) < 1)
+            {
+                throw new ArgumentException("Improperly formatted file path source.", nameof(filePathSource));
+            }
```
ddyett, line 37: duplicative of the checks RetrieveFilePathSourceValues will provide

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -20,11 +21,13 @@ public class PermissionsStore : IPermissionsStore
...
-        private readonly IDictionary<string, ScopeInformation> _delegatedScopesInfoTable;
-        private readonly IDictionary<string, ScopeInformation> _applicationScopesInfoTable;
+        private IDictionary<string, ScopeInformation> _delegatedScopesInfoTable;
+        private IDictionary<string, ScopeInformation> _applicationScopesInfoTable;
         private readonly IFileUtility _fileUtility;
-        private readonly List<string> _permissionsFilePaths;
+        private readonly string _permissionsContainerName;
+        private readonly List<string> _permissionsBlobNames;
         private readonly string _scopesInformation;
+        private static string _localeCode = "en-US"; // default flag to be used to check against incoming Accept-Language header values
```
ddyett, line 30: >static [](start = 16, length = 6)

is this safe to be static?  since calls can be coming in from multiple callers couldn't setting this cause problems for other requests.

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
```diff
@@ -83,7 +94,7 @@ public async Task<IActionResult> GetSampleQueryByIdAsync(string id)
             try
             {
                 // Get the list of sample queries
-                SampleQueriesList sampleQueriesList = await GetSampleQueriesListAsync();
+                SampleQueriesList sampleQueriesList = await FetchSampleQueriesListAsync("En-Us");
```
ddyett, line 97: >En-Us [](start = 89, length = 5)

why not support getting the locale on these calls as well?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: That makes sense. Nice catch.

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
irvinesunday: I missed this one. Thanks.

FileService/Services/AzureBlobStorageUtility.cs
irvinesunday: The `RetrieveFilePathSourceValues` method is actually not providing the `filePathSource.IndexOf()` check above; It's only offering the argument nullability check. I added that in since it's a public facing method.

image

PR #235, opened 2020-04-30
Adds caching of samples and permissions files

### LINKED ISSUE #234: Add caching of sample queries and permission files
Since samples and permissions will now be fetched from the Azure blob storage, there is need to add a caching mechanism to help with performance.

irvinesunday:
Closes #234 

Proposes:
- To improve performance, caching of samples and permissions files has been added. The cache will will be refreshed every 24 hours with new data.
- Refactoring the `PermissionsStore.cs` class in line with the intro. of the caching feature. 
- Updating the `Permissions.ServiceTest` project tests to add mocking support for the `IMemoryCache` and to remove redundant test.
- A few minor code refactorings.

### DIFF
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -32,6 +32,8 @@ namespace GraphExplorerPermissionsService
         private readonly string _scopesInformation;
         private readonly int _defaultRefreshTimeInHours; // life span of the in-memory cache
         private const string DefaultLocale = "en-US"; // default locale language
+        private readonly object _permissionsLock = new object();
+        private static bool _permissionsRefreshed = false;
 
         public PermissionsStore(IFileUtility fileUtility, IConfiguration configuration, IMemoryCache permissionsCache)
         {
@@ -88,6 +90,8 @@ namespace GraphExplorerPermissionsService
                             _scopesListTable.Add(count, property.Value);
                         }
                     }
+
+                    _permissionsRefreshed = true;
                 }
             }
         }
@@ -136,14 +140,15 @@ namespace GraphExplorerPermissionsService
         /// refresh time duration.</returns>
         private bool RefreshPermissionsTables()
         {
-            bool refreshed = false;
+            bool refresh = false;
             bool cacheState = _permissionsCache.GetOrCreate("PermissionsTablesState", entry =>
             {
                 entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromHours(_defaultRefreshTimeInHours);
-                return refreshed = true;
+                _permissionsRefreshed = false;
+                return refresh = true;
             });
 
-            return refreshed;
+            return refresh;
         }
 
         /// <summary>
@@ -163,7 +168,15 @@ namespace GraphExplorerPermissionsService
                 {
                     /* Permissions tables are not localized, so no need to keep different localized cached copies.
                        Refresh tables only after the specified time duration has elapsed or no cached copy exists. */
-                    await SeedPermissionsTables();
+                    lock (_permissionsLock)
+                    {
+                        // Ensure permissions tables are seeded by only one executing thread,
+                        // once per refresh cycle.
+                        if (!_permissionsRefreshed)
+                        {
+                            SeedPermissionsTables();
+                        }
+                    }
                 }
 
                 /* Ensure that the requested localized copy of permissions descriptions
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -30,11 +30,12 @@ namespace GraphExplorerPermissionsService
         private readonly string _permissionsContainerName;
         private readonly List<string> _permissionsBlobNames;
         private readonly string _scopesInformation;
-        private const int RefreshTimeInDays = 1; // life span of the in-memory cache
+        private readonly int _defaultRefreshTimeInHours; // life span of the in-memory cache
         private const string LocaleCode = "en-US"; // default locale language
 
         public PermissionsStore(IFileUtility fileUtility, IConfiguration configuration, IMemoryCache permissionsCache)
         {
+            _defaultRefreshTimeInHours = FileServiceHelper.GetFileCacheRefreshTime(configuration["FileCacheRefreshTimeInHours"]);
             _permissionsCache = permissionsCache;
             _fileUtility = fileUtility;
             _permissionsContainerName = configuration["AzureBlobStorage:Containers:Permissions"];
@@ -136,7 +137,7 @@ namespace GraphExplorerPermissionsService
             bool refreshPermissionsTables = false;
             bool cacheState = _permissionsCache.GetOrCreate("PermissionsTablesState", entry =>
             {
-                entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromDays(RefreshTimeInDays);
+                entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromHours(_defaultRefreshTimeInHours);
                 return refreshPermissionsTables = true;
             });
 
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -6,6 +6,7 @@ using FileService.Common;
 using FileService.Interfaces;
 using GraphExplorerPermissionsService.Interfaces;
 using GraphExplorerPermissionsService.Models;
+using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Configuration;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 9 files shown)


GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -101,61 +89,84 @@ private void SeedPermissionsTables()
...
-                _applicationScopesInfoTable = new Dictionary<string, ScopeInformation>();
-            }
-
-            string relativeScopesInfoPath = FileServiceHelper.GetLocalizedFilePathSource(_permissionsContainerName, _scopesInformation, localeCode);
-            string scopesInfoJson = _fileUtility.ReadFromFile(relativeScopesInfoPath).GetAwaiter().GetResult();
+            string scopesInfoJson = _permissionsCache.GetOrCreate($"ScopesInfoJson_{localeCode}", cacheEntry =>
+             {
+                 string relativeScopesInfoPath = FileServiceHelper.GetLocalizedFilePathSource(_permissionsContainerName, _scopesInformation, localeCode);
+                 cacheEntry.AbsoluteExpirationRelativeToNow = TimeSpan.FromDays(RefreshTimeInDays);
+                 return  _fileUtility.ReadFromFile(relativeScopesInfoPath).GetAwaiter().GetResult();
```
darrelmiller, line 100: Can you make the passed func async so that you can await this?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -101,61 +89,84 @@ private void SeedPermissionsTables()
...
-            if (!string.IsNullOrEmpty(localeCode))
-            {
-                // Clear tables to store new localized descriptions
-                _delegatedScopesInfoTable = new Dictionary<string, ScopeInformation>();
-                _applicationScopesInfoTable = new Dictionary<string, ScopeInformation>();
-            }
-
-            string relativeScopesInfoPath = FileServiceHelper.GetLocalizedFilePathSource(_permissionsContainerName, _scopesInformation, localeCode);
-            string scopesInfoJson = _fileUtility.ReadFromFile(relativeScopesInfoPath).GetAwaiter().GetResult();
+            string scopesInfoJson = _permissionsCache.GetOrCreate($"ScopesInfoJson_{localeCode}", cacheEntry =>
```
darrelmiller, line 96: Why would we use an in-memory cache with an expiry over a static variable?  The permissions are only going to get updated when we redeploy a new file. At redeploy the service will be restarted and re-initialized.  The In-memory cache seems like overkill.

irvinesunday: No, we are now fetching the files over the wire from Azure blob storage. When we update the files there would be no need to redeploy and/or restart the app.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -101,61 +89,84 @@ private void SeedPermissionsTables()
...
+                    }
+
+                    foreach (ScopeInformation applicationScopeInfo in scopesInformationList.ApplicationScopesList)
+                    {
+                        _applicationScopesInfoTable.Add(applicationScopeInfo.ScopeName, applicationScopeInfo);
+                    }
+
+                    cacheEntry.AbsoluteExpirationRelativeToNow = TimeSpan.FromDays(RefreshTimeInDays);
+                    return scopesInformationList;
+                });
```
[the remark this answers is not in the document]
irvinesunday: This really is redundant. Resolved

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Yes. Should be async. 
Resolved

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -6,78 +6,67 @@
...
-                string jsonString = _fileUtility.ReadFromFile(relativePermissionPath).GetAwaiter().GetResult();
+                string jsonString = await _fileUtility.ReadFromFile(relativePermissionPath);
 
                 if (!string.IsNullOrEmpty(jsonString))
                 {
                     JObject permissionsObject = JObject.Parse(jsonString);
 
+                    if (permissionsObject.Count < 1)
+                    {
+                        throw new InvalidOperationException($"The permissions data sources cannot be empty.");
```
zengin, line 67: >The permissions data sources cannot be empty. [](start = 62, length = 45)

If it is easy to describe how to mitigate this, adding that could be valuable, i.e. by which action one can fill the data sources.

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
```diff
@@ -24,25 +25,21 @@ public GraphExplorerPermissionsController(IPermissionsStore permissionsStore)
...
 
-                if (result == null)
-                {
-                    return NotFound();
-                }
+                List<ScopeInformation> result = null;
+                result = await _permissionsStore.GetScopesAsync(scopeType, localeCode, requestUrl, method);
 
-                return Ok(result);
+                return result == null ? NotFound() : (IActionResult)Ok(result);
```
zengin, line 42: >(IActionResult) [](start = 53, length = 15)

Doesn't it work without the cast?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -128,34 +118,55 @@ private void SeedScopesInfoTables(string localeCode = null)
...
         /// <summary>
-        /// Retrieves the permission scopes
+        /// Determines whether the permissions tables need to be refreshed with new data based on the elapsed time
+        /// duration since the previous refresh.
+        /// </summary>
+        /// <returns>true or false based on whether the elapsed time duration is greater or less than the specified
+        /// refresh time duration.</returns>
+        private bool RefreshPermissionsTables()
+        {
+            bool refreshPermissionsTables = false;
```
zengin, line 136: >refreshPermissionsTables [](start = 17, length = 24)

NIT: When the context is clear, you can keep these names short. I would propose "refreshed".

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -6,78 +6,67 @@
...
+        private IDictionary<int, object> _scopesListTable;
         private IDictionary<string, ScopeInformation> _delegatedScopesInfoTable;
         private IDictionary<string, ScopeInformation> _applicationScopesInfoTable;
+        private readonly IMemoryCache _permissionsCache;
         private readonly IFileUtility _fileUtility;
         private readonly string _permissionsContainerName;
         private readonly List<string> _permissionsBlobNames;
         private readonly string _scopesInformation;
-        private string _localeCode = "en-US"; // default flag to be used to check against incoming Accept-Language header values
+        private const int RefreshTimeInDays = 1; // life span of the in-memory cache
```
ddyett, line 33: >RefreshTimeInDays [](start = 26, length = 17)

consider of this should be hours and whether this setting should be provided via config.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -128,34 +118,55 @@ private void SeedScopesInfoTables(string localeCode = null)
...
         }
 
         /// <summary>
-        /// Retrieves the permission scopes
+        /// Determines whether the permissions tables need to be refreshed with new data based on the elapsed time
+        /// duration since the previous refresh.
+        /// </summary>
+        /// <returns>true or false based on whether the elapsed time duration is greater or less than the specified
+        /// refresh time duration.</returns>
+        private bool RefreshPermissionsTables()
```
zengin, line 134: >RefreshPermissionsTables [](start = 21, length = 24)

Is it possible that two threads execute this function at the same time? If yes,
1. How often can that occur?
2. Is the operation idempotent, i.e. running it twice yields the same result as running it only once?
3. If the operation is costly, can one thread wait on another to complete and use its result?

The function name GetOrCreate suggests that these are all covered at that layer if the operation is handled in order. Please confirm if it is so.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -6,78 +6,67 @@
...
         private IDictionary<string, ScopeInformation> _delegatedScopesInfoTable;
         private IDictionary<string, ScopeInformation> _applicationScopesInfoTable;
+        private readonly IMemoryCache _permissionsCache;
         private readonly IFileUtility _fileUtility;
         private readonly string _permissionsContainerName;
         private readonly List<string> _permissionsBlobNames;
         private readonly string _scopesInformation;
-        private string _localeCode = "en-US"; // default flag to be used to check against incoming Accept-Language header values
+        private const int RefreshTimeInDays = 1; // life span of the in-memory cache
+        private const string LocaleCode = "en-US"; // default locale language
```
ddyett, line 34: >LocaleCode [](start = 29, length = 10)

NIT: better named DefaultLocaleCode

zengin: NIT: you can also drop Code postfix, locale is a widely known concept, so I don't think dropping it could cause any ambiguity.

---
In reply to: [418198950](https://github.com/microsoftgraph/microsoft-graph-explorer-api/pull/235#discussion_r418198950) [](ancestors = 418198950)

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -128,34 +118,55 @@ private void SeedScopesInfoTables(string localeCode = null)
...
+            try
             {
-                throw new InvalidOperationException($"The permissions and scopes data sources are empty; " +
-                    $"check the source file or check whether the file path is properly set. File path: {_permissionsBlobNames}");
-            }
+                if (RefreshPermissionsTables())
+                {
+                    /* Permissions tables are not localized, so no need to keep different localized cached copies.
+                       Refresh tables only after the specified time duration has elapsed or no cached copy exists. */
+                    await SeedPermissionsTables();
```
ddyett, line 163: >SeedPermissionsTables [](start = 26, length = 21)

how do you ensure that this is only done once, i.e if multiple requests come in it seems like they would all try to seed for some time.

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
irvinesunday: You would get the error: "Type of conditional expression cannot be determined because there is no implicit conversion between `Microsoft.AspNetCore.Mvc.NotFoundResult` and `Microsoft.AspNetCore.Mvc.OkObjectResult`"

zengin: Thanks for the information. I thought the compiler can use return type as a type hint, but I guess that is a limitation of the generic Task<T> type.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Good idea. Resolved

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Resolved with both suggestions.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Resolved

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Yes, it is possible that multiple threads might execute the function at the same time.
1. With the current setting that refreshes the cache every 24 hours, and with the possibility of having 8 different caches being created for the localized permissions, for example, I would say that multiple threads executing this function doesn't seem like an expensive operation. If we got the refresh rate down to every minute or perhaps every hour, however, then we would need to start thinking of adding locks. Answering question **3.**
2. Yes it is idempotent. Furthermore, in a race condition, the cache will most likely be set once by one of the threads. The other threads will Get the value of the newly created cache when attempting to Create the cache. And subsequent attempts will of course fetch this cached value.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Good point. 
I've made the method synchronous and added a lock to make sure only one thread seeds the Permissions table. Also added a flag to check whether it has been seeded or not in a refresh cycle.

GraphWebApi/appsettings.json
```diff
@@ -30,7 +30,8 @@
           "permissions-beta.json"
         ],
         "Descriptions": "permissions-descriptions.json"
-      }     
-    }    
-  }
+      }
+    }
+  },
+  "FileCacheRefreshTimeInHours": "12"
```
ddyett, line 36: >12 [](start = 34, length = 2)

should be 24 based on the default in FileServiceHelper, but I assume this is a placeholder.

irvinesunday: It should be 24

GraphExplorerPermissionsService/Services/PermissionsStore.cs
ddyett: not seeing this.

irvinesunday: Are you saying you can't see the changes? Here's the commit: 
https://github.com/microsoftgraph/microsoft-graph-explorer-api/pull/235/commits/0ea2b09c3a18d8006618dbeeacb51b6b3be2035c

PR #245, opened 2020-05-06
Formats paths function placeholder values in Graph OpenAPI spec.

### LINKED ISSUE #244: Format path function parameter values appropriately

irvinesunday:
Closes #244 #246 

Proposes:
- Adding single quotation marks to the placeholder values of path functions.
e.g. given the path
`/reports/microsoft.graph.getTeamsUserActivityCounts(period={period})`
the expected output should be...
`/reports/microsoft.graph.getTeamsUserActivityCounts(period='{period}')`

This fixes https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/151


OpenAPIService/OpenApiService.cs
```diff
@@ -451,5 +452,31 @@ private static void RemoveContent(OpenApiDocument target)
...
+        private static string FormatPathFunctions(string pathKey)
+        {
+            if (string.IsNullOrEmpty(pathKey))
+            {
+                return null;
+            }
+
+            /* Example:
+             * Actual ---->  /workbooks({id})/microsoft.graph.delta(token={token})'
+             * Expected -->  /workbooks({id})/microsoft.graph.delta(token='{token}')'
```
darrelmiller, line 471: I don't believe all token values are quoted.  I think it depends on the dataType. https://docs.microsoft.com/en-us/graph/api/reportroot-getoffice365activeuserdetail?view=graph-rest-1.0#http-request  It might be worth verifying this in the OData specification.

irvinesunday: The data types are defined in the parameter properties, `Format` and `Type`. I've modified the function to check for these.

PR #255, opened 2020-05-15
Fixes permissions descriptions cache storage issues

### LINKED ISSUE #254: Retrieving localized cached permissions as currently implemented is not working as expected.

irvinesunday:
Closes #254 

**Proposes**:
Retrieving cached localized permissions as currently implemented fetches the in-memory scopes information of the 1st cached instance. This fix adds a dictionary which holds the `Application` and `Delegated` scopes info. and caches it against the respective locale that is requested.

### DIFF
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -25,6 +25,7 @@ namespace GraphExplorerPermissionsService
         private IDictionary<int, object> _scopesListTable;
         private IDictionary<string, ScopeInformation> _delegatedScopesInfoTable;
         private IDictionary<string, ScopeInformation> _applicationScopesInfoTable;
+        private IDictionary<string, IDictionary<string, ScopeInformation>> _scopesInformationDictionary;
         private readonly IMemoryCache _permissionsCache;
         private readonly IFileUtility _fileUtility;
         private readonly string _permissionsContainerName;
@@ -34,6 +35,8 @@ namespace GraphExplorerPermissionsService
         private const string DefaultLocale = "en-US"; // default locale language
         private readonly object _permissionsLock = new object();
         private static bool _permissionsRefreshed = false;
+        private const string Delegated = "Delegated";
+        private const string Application = "Application";
 
         public PermissionsStore(IFileUtility fileUtility, IConfiguration configuration, IMemoryCache permissionsCache)
         {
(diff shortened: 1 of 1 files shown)


GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -101,7 +104,7 @@ private void SeedPermissionsTables()
         /// </summary>
         private async Task SeedScopesInfoTablesAsync(string locale = DefaultLocale)
         {
-            ScopesInformationList scopesInformationList = await _permissionsCache.GetOrCreateAsync($"ScopesInfoList_{locale}", async cacheEntry =>
+            _scopesInformationDictionary = await _permissionsCache.GetOrCreateAsync($"ScopesInfoList_{locale}", async cacheEntry =>
             {
                 _delegatedScopesInfoTable = new Dictionary<string, ScopeInformation>();
```
ddyett, line 109: >_delegatedScopesInfoTable [](start = 16, length = 25)

can be local now

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -101,7 +104,7 @@ private void SeedPermissionsTables()
...
                 _applicationScopesInfoTable = new Dictionary<string, ScopeInformation>();
```
ddyett, line 110: >_applicationScopesInfoTable [](start = 16, length = 27)

should be local now

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -128,7 +131,13 @@ private async Task SeedScopesInfoTablesAsync(string locale = DefaultLocale)
 
                 cacheEntry.AbsoluteExpirationRelativeToNow = TimeSpan.FromHours(_defaultRefreshTimeInHours);
 
-                return scopesInformationList;
+                _scopesInformationDictionary = new Dictionary<string, IDictionary<string, ScopeInformation>>
```
ddyett, line 134: >_scopesInformationDictionary [](start = 16, length = 28)

you set here then return it to be set onto the same variable. you can just return the new Dictionary

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -25,6 +25,7 @@ public class PermissionsStore : IPermissionsStore
         private IDictionary<int, object> _scopesListTable;
         private IDictionary<string, ScopeInformation> _delegatedScopesInfoTable;
         private IDictionary<string, ScopeInformation> _applicationScopesInfoTable;
```
zengin, line 27: The information these two carry are already in the field `_scopesInformationDictionary`. You can turn these into local variables within `SeedScopesInfoTablesAsync`.

irvinesunday:
@zengin 
Before this fix, the cached `scopesInformationList` value was not being used when retrieving the values of the delegated scopes and/or the application scopes, rather, these values were being fetched from the private fields `_delegatedScopesInfoTable` and `_applicationScopesInfoTable`, which were being set within the body of the locale permissions cache creation. This meant we were always referencing the delegated/application scope values of the last cache that was set when fetching the permissions descriptions for any locale option.

PR #263, opened 2020-06-03
Returns all and fixes rendering permissions

### LINKED ISSUE #262: Fix inconsistencies in fetching permissions in DevX API

irvinesunday:
Closes #262 #261

Proposes:
- Returning all the permissions definitions and descriptions defined in the `permissions-descriptions.json` file instead of fetching the permissions definitions from the request urls json file.
- Fixes the issue of inconsistent permissions rendering. Localized permissions descriptions are seeded by only one executing thread, and a `lock` is used to avoid collision. If a thread finally gets a hold of the released lock and attempts to seed an already seeded localized permissions descriptions list, it instead gets assigned the cached copy of the permissions descriptions that were added prior.

### DIFF
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -254,8 +254,7 @@ namespace GraphExplorerPermissionsService
                     JArray resultValue = new JArray();
                     resultValue = (JArray)_scopesListTable[int.Parse(resultMatch.Key)];
 
-                    string[] scopes = null;
-                    scopes = resultValue.FirstOrDefault(x => x.Value<string>("HttpVerb") == method)?
+                    var scopes = resultValue.FirstOrDefault(x => x.Value<string>("HttpVerb") == method)?
                         .SelectToken(scopeType)?
                         .Select(s => (string)s)
                         .ToArray();
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -115,7 +115,8 @@ namespace GraphExplorerPermissionsService
                      * instance of the localized permissions descriptions
                      * during the lock.
                      */
-                    if (_permissionsCache.Get($"ScopesInfoList_{locale}") == null)
+                    var seededScopesInfoDictionary = _permissionsCache.Get<IDictionary<string, IDictionary<string, ScopeInformation>>>($"ScopesInfoList_{locale}");
+                    if (seededScopesInfoDictionary == null)
                     {
                         var _delegatedScopesInfoTable = new Dictionary<string, ScopeInformation>();
                         var _applicationScopesInfoTable = new Dictionary<string, ScopeInformation>();
@@ -148,8 +149,9 @@ namespace GraphExplorerPermissionsService
                             { Application, _applicationScopesInfoTable }
                         };
                     }
-                    // Fetch the localized cached permissions descriptions already seeded
-                    return _permissionsCache.Get<IDictionary<string, IDictionary<string, ScopeInformation>>>($"ScopesInfoList_{locale}");
+                    /* Fetch the localized cached permissions descriptions
+                       already seeded by previous thread. */
+                    return seededScopesInfoDictionary;
                 }
             });
             return scopesInformationDictionary;
@@ -201,8 +203,8 @@ namespace GraphExplorerPermissionsService
                        Refresh tables only after the specified time duration has elapsed or no cached copy exists. */
                     lock (_permissionsLock)
                     {
-                        // Ensure permissions tables are seeded by only one executing thread,
-                        // once per refresh cycle.
+                        /* Ensure permissions tables are seeded by only one executing thread,
+                           once per refresh cycle. */
                         if (!_permissionsRefreshed)
                         {
                             SeedPermissionsTables();
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -100,6 +100,8 @@ namespace GraphExplorerPermissionsService
         /// <summary>
         /// Gets or creates the localized permissions descriptions from the cache.
         /// </summary>
+        /// <param name="locale">The locale of the permissions decriptions file.</param>
+        /// <returns>The localized instance of permissions descriptions.</returns>
         private async Task<IDictionary<string, IDictionary<string, ScopeInformation>>> GetOrCreatePermissionsDescriptionsAsync(string locale = DefaultLocale)
         {
             var scopesInformationDictionary = await _permissionsCache.GetOrCreateAsync($"ScopesInfoList_{locale}", async cacheEntry =>
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -218,16 +218,22 @@ namespace GraphExplorerPermissionsService
 
                     if (scopeType.Contains(Delegated))
                     {
-                        foreach(var scopesInfo in scopesInformationDictionary[Delegated])
+                        if (scopesInformationDictionary.ContainsKey(Delegated))
                         {
-                            scopesListInfo.Add(scopesInfo.Value);
+                            foreach (var scopesInfo in scopesInformationDictionary[Delegated])
+                            {
+                                scopesListInfo.Add(scopesInfo.Value);
+                            }
                         }
                     }
                     else // Application scopes
                     {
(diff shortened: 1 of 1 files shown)


GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -98,42 +99,57 @@ private void SeedPermissionsTables()
...
-                    { Application, _applicationScopesInfoTable }
-                };
+                        return new Dictionary<string, IDictionary<string, ScopeInformation>>
+                        {
+                            { Delegated, _delegatedScopesInfoTable },
+                            { Application, _applicationScopesInfoTable }
+                        };
+                    }
+                    // Fetch the localized cached permissions descriptions already seeded
+                    return _permissionsCache.Get<IDictionary<string, IDictionary<string, ScopeInformation>>>($"ScopesInfoList_{locale}");
```
ddyett, line 151: >_permissionsCache [](start = 27, length = 17)

function doesn't indicate it returns something and it doesn't look like these returns are ever used.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -98,42 +99,57 @@ private void SeedPermissionsTables()
         }
 
         /// <summary>
-        /// Populates the delegated and application scopes information table caches.
+        /// Retrieves or adds the localized permissions descriptions from the cache.
         /// </summary>
-        private async Task SeedScopesInfoTablesAsync(string locale = DefaultLocale)
+        private async Task SeedPermissionsDescriptionsAsync(string locale = DefaultLocale)
```
ddyett, line 104: >SeedPermissionsDescriptionsAsync [](start = 27, length = 32)

function really is more of a RetrievePermissionsDescriptions since most times it will grab a cached copy.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -171,7 +187,13 @@ private bool RefreshPermissionsTables()
...
             try
             {
-                if (RefreshPermissionsTables())
+                /* Add multiple checks to ensure thread that
+                 * populated scopes information successfully
+                 * completed seeding.
+                */
+                    if (RefreshPermissionsTables() ||
+                    _scopesListTable == null ||
+                    !_scopesListTable.Any())
```
ddyett, line 196: do you need to seedpermissionstables if requesturl is empty since you no longer need _scopesList to fulfill that?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
zengin, line 151: >Get [](start = 45, length = 3)

It looks like Get() gets called twice when there is a cache hit. Could you avoid that by assigning to a local variable?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
zengin: I propose something along the lines of "GetOrCreate" to stay consistent with the naming in library functions.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -186,36 +208,28 @@ private bool RefreshPermissionsTables()
...
-                            permissionsList.AddRange(permissions);
+                            scopesListInfo.Add(scopesInfo.Value);
                         }
-
                     }
-                    if (permissionsList.Count > 0)
+                    else // Application scopes
                     {
-                        scopes = permissionsList.Distinct().ToArray();
+                        foreach (var scopesInfo in _scopesInformationDictionary[Application])
```
zengin, line 226: Q: Is it guaranteed to have an application entry when there is no delegated entry? Otherwise you may get a `KeyNotFoundException`.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -224,6 +238,7 @@ private bool RefreshPermissionsTables()
                         throw new ArgumentNullException(nameof(method), "The HTTP method value cannot be null or empty.");
                     }
 
+                    string[] scopes = null;
```
zengin, line 241: STYLE: Please move declaration closer to the first usage of this local variable.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Since this table will only be seeded once per each refresh cycle, by only one executing thread, there's no major cost for the 1st request that hits the endpoint to do so. After all, we would want both tables cache to be refreshed at the same interval.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
andrueastman: +1

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: We are explicitly creating those two keys here:

image

They could be empty collections, but we should not expect a `KeyNotFoundException`

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Actually, it doesn't get hit whenever there is a cache `Get()`. Whenever there is a cache `Get()` execution does not continue inside the function:

image

If a cache value is found, the `scopesInformationDictionary`  variable gets assigned the value and execution leaves the function.

image

The above line only gets executed in the event that more than one executing thread is trying to create the same cache and only one gets an exclusive `lock` for this. The rest wait in turn and will attempt to create the same cache after the first one completes, but instead, they will each `Get()` the cache value that was created by the first `lock` holder.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Got it!

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: Resolved with `GetOrCreatePermissionsDescriptionsAsync`

GraphExplorerPermissionsService/Services/PermissionsStore.cs
zengin: Given that this is in an always called private function, it is a safe assumption at the moment. This assumption can be broken by someone else modifying the code. Two possible ways:
1. Other function gets modified assuming standalone behavior
2. Private member __scopesInformationDictionary gets assigned in other parts of the class to support new features.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
zengin: I was referring to the Get() call in the `if` block inside the lock. Don't we execute once for the if condition first and then for the return statement?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
zengin: Declaration and assignment can also be merged ;)

irvinesunday: Oh yes!

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: No, not at all. There is a return statement within the `if` block that ensures execution leaves the method if we get inside this block and only hits the outer `return` statement otherwise.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: I hear you. Better err on the side of caution. I can add guard clauses to validate the existence of the two keys :)

irvinesunday: **PS**: Private member `_scopesInformationDictionary` has been replaced with a local variable `scopesInformationDictionary` since it always gets assigned a potentially new local value with each request with the call to `GetOrCreatePermissionsDescriptionsAsync()`.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
zengin: if `if (Get() == null)` evaluates to false, you call `Get()` once to compare against null, and once again for `return`.

irvinesunday: Aaah.. yes! I get you now.

irvinesunday: Resolved

PR #268, opened 2020-06-10
Use transformed metadata in snippet generation

zengin:
SDK uses transformed metadata and snippet generation uses raw metadata. This causes discrepancies in parameter order.

Changed the url to clean (transformed) metadata and added fallback to original metadata in case any parsing errors occur.

@andrueastman, FYI once this gets checked in, you will observe a reordering of parameters for `accept`, `decline` and `tentativelyAccept` in the next snippets update. This is expected due to a rule in XSLT transform.

cc: @darrelmiller, @pixieofhugs

[AB#4973](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4973)

### DIFF
--- a/CodeSnippetsReflection/SnippetsGenerator.cs
+++ b/CodeSnippetsReflection/SnippetsGenerator.cs
@@ -42,23 +42,8 @@ namespace CodeSnippetsReflection
             ServiceRootBeta = new Uri("https://graph.microsoft.com/beta");
 
             // use transformed metadata, fallback to original if there are any parsing errors
-            try
-            {
-                IedmModelV1 = new Lazy<IEdmModel>(() => CsdlReader.Parse(XmlReader.Create("https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsv1.0.xml")));
-            }
-            catch
-            {
-                IedmModelV1 = new Lazy<IEdmModel>(() => CsdlReader.Parse(XmlReader.Create(ServiceRootV1 + "/$metadata")));
-            }
-
-            try
-            {
-                IedmModelBeta = new Lazy<IEdmModel>(() => CsdlReader.Parse(XmlReader.Create("https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_beta_metadata/cleanMetadataWithDescriptionsbeta.xml")));
-            }
-            catch
-            {
-                IedmModelBeta = new Lazy<IEdmModel>(() => CsdlReader.Parse(XmlReader.Create(ServiceRootBeta + "/$metadata")));
-            }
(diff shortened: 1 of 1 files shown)


CodeSnippetsReflection/SnippetsGenerator.cs
```diff
@@ -41,8 +41,24 @@ private void LoadGraphMetadata()
...
-            IedmModelV1 = new Lazy<IEdmModel>(() => CsdlReader.Parse(XmlReader.Create(ServiceRootV1 + "/$metadata")));
-            IedmModelBeta = new Lazy<IEdmModel>(() => CsdlReader.Parse(XmlReader.Create(ServiceRootBeta + "/$metadata")));
+            // use transformed metadata, fallback to original if there are any parsing errors
+            try
+            {
+                IedmModelV1 = new Lazy<IEdmModel>(() => CsdlReader.Parse(XmlReader.Create("https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsv1.0.xml")));
+            }
+            catch
+            {
+                IedmModelV1 = new Lazy<IEdmModel>(() => CsdlReader.Parse(XmlReader.Create(ServiceRootV1 + "/$metadata")));
```
ddyett, line 51: falling back seems like it can produce inconsistencies from week to week if something goes wrong.  It might be better to just let things fail.

zengin: For week to week doc generation, I agree. Are we ok with it failing in graph explorer if the clean metadata in github breaks?

ddyett: not ok with it but I don't think this is the right way to deal with it.  We need to be able to detect it and revert to last known good to restore GE functionality.  It also means our checks for updating the clean metadata have failed.

zengin: @MIchaelMainer, is it guaranteed that there is no parsing errors in clean metadata, or do we run the checks after writing the file into the repo?

MIchaelMainer: No guarantee. Before we consider the metadata as "clean", we have successfully:
* added doc comments.
* generated code files for .NET as validation.
* the .NET code files build success validation.

Once this level of "clean" is reached, then we check it in to the repo.

We don't control the contents of the input metadata and the documentation so it is possible that we will find new dirtiness that slipped pass our validation.

I suggest that the fallback be to the last successful clean metadata.

ddyett: i meant that a failure here is feedback to metadata validation that we should probably add some testing but I didn't necessarily mean the validation covers all things.  The fallback to a last known good should be the way to go if we can detect it.  Perhaps Raptor can also be a gating function for devX deployment though my senses are tell me that we are likely reading graph metadata directly from github instead of blob so raptor wouldn't have a chance to help prevent a bad deployment.  That's a follow up for me and @irvinesunday to talk about outside of this PR.

zengin: Could one of you then approve this please? ;)

PR #269, opened 2020-06-11
Added stream support for snippets

zengin:
[Snippets Diff](https://github.com/microsoftgraph/microsoft-graph-docs/compare/zengin/snippets-before-stream...zengin/snippets-after-stream)

In SDK, Stream PUT requests have two versions: `PutAsync()` and `PutAsync<T>()`. This PR addresses converting strings from HTTP sample to a MemoryStream and passing that as a parameter to PUT requests.

[AB#4976](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/4976)

### DIFF
--- a/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
+++ b/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
@@ -111,7 +111,7 @@ namespace CodeSnippetsReflection.LanguageGenerators
 
                     snippetBuilder.Append(GenerateRequestSection(snippetModel, $"{actions}\n\t.UpdateAsync({snippetModel.ResponseVariableName});"));
                 }
-                else if(snippetModel.Method == HttpMethod.Delete)
+                else if (snippetModel.Method == HttpMethod.Delete)
                 {
                     snippetBuilder.Append(GenerateRequestSection(snippetModel, $"{actions}\n\t.DeleteAsync();"));
                 }
@@ -120,6 +120,7 @@ namespace CodeSnippetsReflection.LanguageGenerators
                     if (string.IsNullOrEmpty(snippetModel.RequestBody))
                         throw new Exception($"No request Body present for PUT of entity {snippetModel.ResponseVariableName}");
 
+                    var genericType = string.Empty;
                     if (snippetModel.ContentType.Equals("application/json", StringComparison.OrdinalIgnoreCase))
                     {
                         snippetBuilder.Append($"var {snippetModel.ResponseVariableName} = ");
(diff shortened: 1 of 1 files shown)


CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -120,17 +120,25 @@ public string GenerateCodeSnippet(SnippetModel snippetModel, LanguageExpressions
...
                         snippetBuilder.Append($"var {snippetModel.ResponseVariableName} = ");
                         snippetBuilder.Append(CSharpGenerateObjectFromJson(segment, snippetModel.RequestBody, new List<string> { snippetModel.ResponseVariableName }));
                     }
                     else
                     {
-                        snippetBuilder.Append($"var {snippetModel.ResponseVariableName} = \"{snippetModel.RequestBody.Trim()}\"\n\n");
+                        snippetBuilder.Append($"using var {snippetModel.ResponseVariableName} = new System.IO.MemoryStream(Encoding.UTF8.GetBytes(\"{snippetModel.RequestBody.Trim()}\"));\n\n");
+
+                        // resolve type for PutAsync<T>
+                        var genericEdmType = snippetModel.Segments[snippetModel.Segments.Count - 2].EdmType;
```
ddyett, line 134: is it sage to assume count is always 2 or more or is this code just ok with exceptions getting thrown if things are off.

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -120,17 +120,25 @@ public string GenerateCodeSnippet(SnippetModel snippetModel, LanguageExpressions
...
+                    var genericType = string.Empty;
                     if (snippetModel.ContentType.Equals("application/json", StringComparison.OrdinalIgnoreCase))
                     {
                         snippetBuilder.Append($"var {snippetModel.ResponseVariableName} = ");
                         snippetBuilder.Append(CSharpGenerateObjectFromJson(segment, snippetModel.RequestBody, new List<string> { snippetModel.ResponseVariableName }));
                     }
                     else
                     {
-                        snippetBuilder.Append($"var {snippetModel.ResponseVariableName} = \"{snippetModel.RequestBody.Trim()}\"\n\n");
+                        snippetBuilder.Append($"using var {snippetModel.ResponseVariableName} = new System.IO.MemoryStream(Encoding.UTF8.GetBytes(\"{snippetModel.RequestBody.Trim()}\"));\n\n");
```
ddyett, line 131: should there be a check on the returned contenttype?

zengin: not a concern at the moment. If (or when) we get to a place where we actually run these against the graph endpoints, then we should consider that.

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
zengin: It is ok to throw exceptions in the way this code is structured. However, we should eventually handle reporting of these because all snippet generation failures are silent at the moment.

PR #276, opened 2020-07-04
Add additionalProperties to schemas for objects for PowerShell Style

darrelmiller:
AutoREST assumes additionalProperties is false and therefore doesn't create the additionalProperties dictionary for properties in the JSON response that are not defined in the Schema.  This is a problem from types that are considered OData "openTypes" and also an issue for accessing instance annotations.

This PR creates an empty schema for objects that always validates to true.  This is equivalent to setting AdditionalProperties to true.  Settting additionalProperties to true doesn't help because that is the default value and is not rendered during serialization.


darrelmiller:
AB5154

OpenAPIService/PowershellFormatter.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Microsoft.OpenApi.Models;
```
irvinesunday, line 1: Please add license info.

PR #290, opened 2020-08-07
Fix for incorrect sample queries language displayed in GE

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/659

**Cause**: This happens when multiple requests are attempting to seed the sample queries simultaneously in the DevX API. This is an intermittent problem which is temporarily remedied by initiating a restart of the app service.

**Proposed fix**:
- Refactoring out the samples queries fetching process to a separate class, `SamplesStore.cs`; and injecting this class through DI.
- Using thread locking to ensure one locale cache is seeded by one executing thread at any given time.
- Adding a test utility library, `MockTestUtility`, for helping out with mocking of tests among test projects, and refactoring out common code into this library.
- Adding tests to validate the `SamplesStore.cs` class functionality.
- Substituting deprecated [functionality ](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/89fa6533cf7dd89d7309381a492b82a7d9de20b3/GraphWebApi/Controllers/GraphExplorerSamplesController.cs#L316-L319)serving a deprecated method with a `NotImplementedException()`. This was taking a dependency on some of the fields that were deleted (due to the sample queries fetch functionality being refactored out to a separate class), hence throwing errors.


GraphExplorerSamplesService/Services/SamplesStore.cs
```diff
@@ -0,0 +1,89 @@
...
+            {
+                // Localized copy of samples is to be seeded by only one executing thread.
+                lock (_samplesLock)
+                {
+                    /* Check whether a previous thread already seeded an
+                     * instance of the localized samples during the lock.
+                     */
+                    var lockedLocale = locale;
+                    var seededSampleQueriesList = _samplesCache.Get<SampleQueriesList>(lockedLocale);
+                    if (seededSampleQueriesList == null)
```
zengin, line 58: nit:
invert if with an early return to reduce indentation:
```
if (seededSampleQueriesList != null)
{
  return seededSampleQueriesList;
}

// lines 60-80 goes here with less indentation
```

GraphExplorerSamplesService/Services/SamplesStore.cs
```diff
@@ -0,0 +1,89 @@
...
+
+                        bool orderSamples = false;
+
+                        if (lockedLocale.Equals("en-us", StringComparison.OrdinalIgnoreCase))
+                        {
+                            /* Current business process only supports ordering of the English
+                               translation of the sample queries.
+                            */
+                            orderSamples = true;
+                        }
```
zengin, line 77:
```suggestion
                        /* Current business process only supports ordering of the English
                           translation of the sample queries.
                        */
                        bool orderSamples = lockedLocale.Equals("en-us", StringComparison.OrdinalIgnoreCase);
```

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
```diff
@@ -323,76 +308,14 @@ public async Task<IActionResult> DeleteSampleQueryAsync(string id)
...
-                // Save the JSON string to the source file
-                await _fileUtility.WriteToFile(policiesJson, _policiesFilePathSource);
-
-                // Return the list of policies
-                return policies;
-            }
-
-            // Return the list of policies
-            return SamplesPolicyService.DeserializeSampleQueriesPolicies(jsonFileContents);
+            throw new NotImplementedException();
```
zengin, line 318: I don't see anything related to this in the description. Is this an intentional change?

irvinesunday: This is a deprecated functionality serving a deprecated method. It was taking a dependency on some of the fields that were deleted (due to the sample queries fetch functionality being refactored out to a separate class), hence throwing errors. It didn't make sense to try and resolve these.

GraphExplorerSamplesService/Services/SamplesStore.cs
irvinesunday: Sorry, your commit suggestion was rendered stale from the above commit. The suggested changes have been added though.

GraphExplorerSamplesService/Services/SamplesStore.cs
irvinesunday: Resolved

irvinesunday:
> > **Cause**: This happens when multiple requests are attempting to seed the sample queries simultaneously in the DevX API. This is an intermittent problem which is temporarily remedied by initiating a restart of the app service.
> 
> Are multiple requests coming from the same client, or are we mixing different sources of requests?
> 
> Do you mind providing a call pattern from client with corresponding actions on the service side and how singleton pattern is interrupting the same call pattern with the fix?

This is caused by multiple requests coming in from different clients with different `Accept-Language` headers. When one of the request is seeding the cache for its specific locale, the subsequent request coming in alters the locale variable state, hence the mix up in the cached name and the cached sample query data. Using a Singleton will enforce only a single access to the class for every request. This will help prevent inconsistent state by allowing us to perform thread locking.

PR #294, opened 2020-08-12
Fix: Use OpenApiStyle as key to fetch OpenApi docs

### LINKED ISSUE #295: Inconsistent state of OpenAPI spec. when requested with different OpenApiStyles
When a request for a specific OpenAPI spec. for a url/operation from, e.g. PowerShell calls the DevX API, this [expression ](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/feb0f3ca421f1182684a78d64b84a6065cc8d0db/GraphWebApi/Controllers/OpenApiController.cs#L56) gets executed, calling the method that executes these two [statements ](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/feb0f3ca421f1182684a78d64b84a6065cc8d0db/OpenAPIService/OpenApiService.cs#L294-L295) (assuming the forceRefresh argument is false). The last statement saves the generated OpenAPI doc. to this [Concurrent Dictionary](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/feb0f3ca421f1182684a78d64b84a6065cc8d0db/OpenAPIService/OpenApiService.cs#L37). Remember, this doc. has been generated using some custom PowerShell specific OpenAPI settings. Now, assuming a second request comes in, from Graph Explorer for example, to fetch an OpenAPI spec. for a url for Autocomplete. The [first expression](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/feb0f3ca421f1182684a78d64b84a6065cc8d0db/GraphWebApi/Controllers/OpenApiController.cs#L56) gets executed again, calling the previous method that then executes this [block](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/feb0f3ca421f1182684a78d64b84a6065cc8d0db/OpenAPIService/OpenApiService.cs#L289-L292) (assuming the forceRefresh arg. is false). What this will do is to serve back the previously generated doc. cache (which has some PowerShell specific OpenAPI settings and config.) to the GE Autocomplete request. Meaning, we haven't applied the settings for 1) Showing Links and 2) Showing Root Path. The GE Autocomplete feature will therefore have inconsistent experiences, same as the other [OpenApiStyles](https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/feb0f3ca421f1182684a78d64b84a6065cc8d0db/OpenAPIService/OpenApiService.cs#L27-L33) --> PowerShell, PowerPlatform, Plain.  This can be mitigated by setting the argument `forceRefresh = true`, to force generation of new OpenAPI spec. for each `OpenApiStyle`. However, other than this being an expensive operation to perform for each request, it will also overwrite previously generated specs.

irvinesunday:
Fixes #295 

**This PR propose**s:
- Updating the Concurrency Dictionary to be of type `<string, OpenApiDocument>`, such that we store as key, the respective OpenApiStyle options; where the key is `{OpenApiStyle}`_`{graphUri}`
- Adding checks to validate against null arguments.

### DIFF
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -34,7 +34,7 @@ namespace OpenAPIService
 
     public class OpenApiService
     {
-        private static ConcurrentDictionary<OpenApiStyle, OpenApiDocument> _OpenApiDocuments = new ConcurrentDictionary<OpenApiStyle, OpenApiDocument>();
+        private static ConcurrentDictionary<string, OpenApiDocument> _OpenApiDocuments = new ConcurrentDictionary<string, OpenApiDocument>();
         private static UriTemplateTable _uriTemplateTable = new UriTemplateTable();
         private static IDictionary<int, OpenApiOperation[]> _openApiOperationsTable = new Dictionary<int, OpenApiOperation[]>();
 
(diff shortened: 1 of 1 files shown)


irvinesunday:
Closing this PR as there exists another PR offering a better solution.

PR #304, opened 2020-08-24
Setup Application Insights for server-side telemetry

MaggieKimani1:
This PR captures configuring app insights and Serilog to monitor DevX API in order to collect server-side telemetry data including:
- failed requests
- server requests
- server response time
- Exceptions thrown when requests are sent to the API

### DIFF
--- a/GraphWebApi/appsettings.json
+++ b/GraphWebApi/appsettings.json
@@ -10,6 +10,10 @@
       "Default": "Warning"
     }
   },
+  "ApplicationInsights": {
+    "InstrumentationKey": "9a1442b6-0ca1-4827-90b5-812decbeaeba",
+    "AppInsightsApiKey": "ENTER_KEY_HERE"
+  },
   "AllowedHosts": "*",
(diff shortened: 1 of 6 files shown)


GraphWebApi/appsettings.json
```diff
@@ -10,6 +10,10 @@
       "Default": "Warning"
     }
   },
+  "ApplicationInsights": {
+    "InstrumentationKey": "9a1442b6-0ca1-4827-90b5-812decbeaeba",
```
zengin, line 14: I think this should be a secret.

MaggieKimani1: Let me set it as a secret

GraphWebApi/appsettings.json
```diff
@@ -10,6 +10,10 @@
       "Default": "Warning"
     }
   },
```
darrelmiller, line 12: @MaggieKimani1  This is where the Key needs to go.

PR #322, opened 2020-09-28
Adds support for default Namespace in C#

andrueastman:
This PR add support for multiple namespaces in .NET snippets

A relevant test has been added to validate how .net snippets would look like.

Ideally the snippet [here](https://docs.microsoft.com/en-us/graph/api/termstore-set-update?view=graph-rest-beta&tabs=csharp#request) would change to look like this,

```c#

var set = new Microsoft.Graph.TermStore.Set
{
	Description = "mySet"
};

await graphClient.TermStore.Sets["{setId}"]
	.Request()
	.UpdateAsync(set);
```

[Snippets Diff ](https://github.com/microsoftgraph/microsoft-graph-docs/commit/cc12eadd210b7b6bf6989a736c9950be3669ee63)

### DIFF
--- a/CodeSnippetsReflection/TypeProperties/CSharpTypeProperties.cs
+++ b/CodeSnippetsReflection/TypeProperties/CSharpTypeProperties.cs
@@ -1,6 +1,7 @@
 ﻿using CodeSnippetsReflection.LanguageGenerators;
 using Microsoft.OData.Edm;
 using System.Linq;
+using System.Reflection.Metadata;
 
 namespace CodeSnippetsReflection.TypeProperties
 {
@@ -32,17 +33,17 @@ namespace CodeSnippetsReflection.TypeProperties
                     return className;
 
                 // Otherwise modify the classname by concatenation the Namespace prefix.
-                // 1. Fist remove the prefix which is the default namespace
-                // 2. Join any parts by uppercase first letter and dots.
-                var suffix = namespaceString.Replace(DefaultNamespace + ".", string.Empty);
-                var segments = suffix.Split(".");
+                // Join any parts by uppercase first letter and dots.
+                var segments = namespaceString.Split(".");
+                var fullClassName = string.Empty;
                 foreach (var segment in segments)
                 {
                     // prepend the uppercase of each segment and form the classname
                     var uppercaseSegment = CommonGenerator.UppercaseFirstLetter(segment);
-                    className = uppercaseSegment + "." + className;
+                    fullClassName = fullClassName + "." + uppercaseSegment;
                 }
-                return className;
+                // remove starting dot
+                return fullClassName.Substring(1) + "." + className;
             }
         }
 
--- a/CodeSnippetsReflection/TypeProperties/CSharpTypeProperties.cs
+++ b/CodeSnippetsReflection/TypeProperties/CSharpTypeProperties.cs
@@ -1,7 +1,7 @@
 ﻿using CodeSnippetsReflection.LanguageGenerators;
 using Microsoft.OData.Edm;
 using System.Linq;
-using System.Reflection.Metadata;
+using System.Text;
 
 namespace CodeSnippetsReflection.TypeProperties
 {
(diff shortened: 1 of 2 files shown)


ddyett:
can you just link this one to an issue or create one for tracking purposes?

ddyett:
var set = new Microsoft.Graph.TermStore.Set
{
	Description = "mySet"
};

there's a part of me that says only add namespaces for non default namespace.  what are your thoughts?

CodeSnippetsReflection/TypeProperties/CSharpTypeProperties.cs
```diff
@@ -20,7 +26,24 @@ public class CSharpTypeProperties
...
                     className += "Object"; // disambiguate class names that end with Request
                 }
-                return className;
+
+                string namespaceString = (string)EdmType.GetType().GetProperty("Namespace")?.GetValue(EdmType, null) ?? DefaultNamespace;
+                // the classname is okay if we are in the default(or Edm) namespace.
+                if (namespaceString.Equals(DefaultNamespace) || namespaceString == "Edm") 
+                    return className;
+
+                // Otherwise modify the classname by concatenation the Namespace prefix.
```
ddyett, line 35:
```suggestion
                // Otherwise modify the classname by concatenating with the Namespace prefix.
```

CodeSnippetsReflection/TypeProperties/CSharpTypeProperties.cs
```diff
@@ -20,7 +26,24 @@ public class CSharpTypeProperties
...
+                // Join any parts by uppercase first letter and dots.
+                var segments = namespaceString.Split(".");
+                var fullClassName = string.Empty;
+                foreach (var segment in segments)
+                {
+                    // prepend the uppercase of each segment and form the classname
+                    var uppercaseSegment = CommonGenerator.UppercaseFirstLetter(segment);
+                    fullClassName = fullClassName + "." + uppercaseSegment;
```
ddyett, line 43: consider changing to stringbuilder for better garbage collection.

CodeSnippetsReflection/TypeProperties/CSharpTypeProperties.cs
```diff
@@ -20,7 +26,24 @@ public class CSharpTypeProperties
                 {
                     className += "Object"; // disambiguate class names that end with Request
                 }
-                return className;
+
+                string namespaceString = (string)EdmType.GetType().GetProperty("Namespace")?.GetValue(EdmType, null) ?? DefaultNamespace;
+                // the classname is okay if we are in the default(or Edm) namespace.
+                if (namespaceString.Equals(DefaultNamespace) || namespaceString == "Edm")
```
ddyett, line 32: shouldn't this check have prevented going to the concatenation code?

zengin:
@andrueastman please add the full diff of the snippet generation using the new code (similar to here https://github.com/microsoftgraph/microsoft-graph-explorer-api/pull/269). The test you added makes sure that the happy path works, but having a full diff would give us confidence in terms of not having any regressions.

zengin (changes requested):
Can we have `TermStore.Set` instead of `Microsoft.Graph.TermStore.Set` since it is assumed that there is a `using Microsoft.Graph` in the contexts which these snippets appear?

CodeSnippetsReflection/TypeProperties/CSharpTypeProperties.cs
andrueastman: The code returns the default classname if its in the default namesapace.
Otherwise it builds the full type name including the name space.

andrueastman:
@zengin 
If is use `Termstore.Set` compilation fails. I will reach out to you to get more clarification on this front.

zengin:
@andrueastman I thought about partial namespace references in C++ when I wrote that comment. It seems like C# doesn't support it without explicit aliasing.

I will sign-off if you have no regression on the full generation.

andrueastman:
[Snippets Diff ](https://github.com/microsoftgraph/microsoft-graph-docs/commit/cc12eadd210b7b6bf6989a736c9950be3669ee63)

CodeSnippetsReflection/TypeProperties/CSharpTypeProperties.cs
ddyett: my implication here is that since the defaultnamespace is microsoft.graph and termstore is in microsoft.graph shouldn't I just get termstore instead of microsoft.graph.termstore?

ddyett: my confusion is i didn't realize there was a new namespace for termstore.

ddyett (approved):

zengin (approved):
Thanks for the diff.

PR #326, opened 2020-09-29
Raptor - Expose snippet generation logic as an executable

zengin:
The purpose of this change is to have direct access to the snippet generation logic from ApiDoctor so that we don't have to make an HTTP call per snippet.

Snippet Generation is on .NET core 2.2 and ApiDoctor is on .NET Framework 4.5. Easiest way to expose the logic was through creating a thin layer of executable which wraps the snippet generation logic. Snippet generation logic continues to live in the Graph Explorer API repo to continue serving HTTP endpoint for Graph Explorer.

The contract is defined as follows:
- Caller is expected to dump all the HTTP snippets into a directory. File names are expected to end with `-httpSnippet`. Full path to this directory is the first command line argument.
- Second command line argument is the list of comma separated languages.

Example:
- Caller creates a directory `C:\snippets`
- Caller adds HTTP snippets:
  - `C:\snippets\snippet-name-httpSnippet`
  - `C:\snippets\snippet-name2-httpSnippet`
- Executable is called on ruby, c# and javascript like:
  - `CodeSnippetsReflection.App.exe C:\snippets ruby,c#,javascript`
- Console output will be:
  ```
  Skipping these languages as they are not currently supported: ruby
  Supported languages: c# javascript objective-c java
  Running snippet generation for these languages: c# javascript
  ...
  Generation related output
  ...
  ```
- At the end of execution `C:\snippets` will have the following files where the latter four will hold the snippets in corresponding languages:
  - `snippet-name-httpSnippet`
  - `snippet-name2-httpSnippet`
  - `snippet-name---c#`
  - `snippet-name---javascript`
  - `snippet-name2---c#`
  - `snippet-name2---javascript`

### Links
Addressing the prerequisite \# 2  from https://github.com/microsoftgraph/msgraph-sdk-design/blob/master/raptor/pipeline.md#prerequisites-for-defining-the-pipeline-above

AB#6091

cc: @darrelmiller @pixieofhugs

### DIFF
--- a/CodeSnippetsReflection.App/Program.cs
+++ b/CodeSnippetsReflection.App/Program.cs
@@ -1,0 +1,107 @@
+﻿using System;
+using System.IO;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using System.Linq;
+
+namespace CodeSnippetsReflection.App
+{
+    /// <summary>
+    /// This is a thin layer that exposes snippet generation logic as an executable.
+    /// It takes two arguments:
+    /// Argument 0: Full path to a directory which holds HTTP snippets. HTTP snippets are expected to appear
+    ///             one per file where the file name ends with -httpSnippet.
+    /// Argument 1: Languages, comma separated.
+    ///             As of this writing, values are c#, javascript, objective-c, java
+    /// 
+    /// Output is generated in the same folder as the HTTP snippets. -httpSnippet part of the file name is
+    /// replaced with ---language.
+    /// </summary>
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            if (args.Length != 2)
+            {
+                Console.Error.WriteLine("Http snippets directory and languages should be specified");
+                Console.WriteLine(@"Example usage:
+  .\CodeSnippetReflection.App.exe C:\snippets c#,javascript");
+                return;
+            }
+
+            var httpSnippetsDir = args[0];
+            if (!Directory.Exists(httpSnippetsDir))
+            {
+                Console.Error.WriteLine($@"Directory {httpSnippetsDir} does not exist!");
+                return;
+            }
+
+            var languages = args[1]
+                .Split(",")
+                .Where(l => l != string.Empty) // eliminate trailing, leading or consecutive commas
+                .Distinct();
+
+            // splits language list into supported and unsupported languages
+            // where key "true" holds supported and key "false" holds unsupported languages
+            var languageGroups = languages
+                .GroupBy(l => SnippetsGenerator.SupportedLanguages.Contains(l.ToLowerInvariant()))
+                .ToDictionary(g => g.Key, g => g.ToList());
+
+            var supportedLanguages = languageGroups.ContainsKey(true) ? languageGroups[true] : null;
+            var unsupportedLanguages = languageGroups.ContainsKey(false) ? languageGroups[false] : null;
+
+            if (supportedLanguages == null)
+            {
+                Console.Error.WriteLine($"None of the given languages are supported. Supported languages: {string.Join(" ", SnippetsGenerator.SupportedLanguages)}");
+                return;
+            }
+
+            if (unsupportedLanguages != null)
+            {
+                Console.WriteLine($"Skipping these languages as they are not currently supported: {string.Join(" ", unsupportedLanguages)}");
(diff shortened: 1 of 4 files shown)


ddyett (dismissed):

CodeSnippetsReflection.App/Program.cs
```diff
@@ -0,0 +1,107 @@
...
+            var httpSnippetsDir = args[0];
+            if (!Directory.Exists(httpSnippetsDir))
+            {
+                Console.Error.WriteLine($@"Directory {httpSnippetsDir} does not exist!");
+                return;
+            }
+
+            var languages = args[1]
+                .Split(",")
+                .Where(l => l != string.Empty) // eliminate trailing, leading or consecutive commas
```
nikithauc, line 42: Suggestion: Check if string is not empty and not a white space.

nikithauc (dismissed):

andrueastman (dismissed):

CodeSnippetsReflection.App/Program.cs
```diff
@@ -0,0 +1,107 @@
...
+    ///             one per file where the file name ends with -httpSnippet.
+    /// Argument 1: Languages, comma separated.
+    ///             As of this writing, values are c#, javascript, objective-c, java
+    /// 
+    /// Output is generated in the same folder as the HTTP snippets. -httpSnippet part of the file name is
+    /// replaced with ---language.
+    /// </summary>
+    class Program
+    {
+        static void Main(string[] args)
```
baywet, line 23: suggestion: use dotnet command line configuration provider to be a bit more versatile https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-3.1#command-line

CodeSnippetsReflection.App/Program.cs
```diff
@@ -0,0 +1,107 @@
...
+            using var streamContent = new StreamContent(new MemoryStream(Encoding.UTF8.GetBytes(File.ReadAllText(file))));
+            streamContent.Headers.Add("Content-Type", "application/http;msgtype=request");
+
+            string snippet;
+            try
+            {
+                // This is a very fast operation, it is fine to make is synchronuous.
+                // With the parallel foreach in the main method, processing all snippets for C# in both Beta and V1 takes about 7 seconds.
+                // As of this writing, the code was processing 2650 snippets
+                using var message = streamContent.ReadAsHttpRequestMessageAsync().Result;
```
baywet, line 95: suggestion: make this method async (use async await, not the result) and replace the Parralel foreach by task.whenall to let the TPL take care of the scheduling for us

baywet (changes requested):

CodeSnippetsReflection.App/Program.cs
zengin: I have initially implemented it like that. It turned out to be too much overhead for the task creation and scheduling costs to read stuff that is already in a memory stream.

I should have shared the numbers as well.
- 1 minute 7 seconds with async await.
- 7 seconds with parallel foreach.

If the tasks are too small, task creation costs can outweigh the benefits achieved from parallelism. Granularity control is a whole research topic and if you are interested, see this paper from my earlier research group for the discussion of when to go sequential vs parallel: http://deepsea.inria.fr/pasl/oracle_scheduling.pdf

CodeSnippetsReflection.App/Program.cs
zengin: I initially thought that the command line parameter can't have spaces in it because then it will be a separate command line argument. But one may pass in arguments in quotes where there are spaces in them. I will add this check.

CodeSnippetsReflection.App/Program.cs
baywet: thanks for sharing this information! I ran so many times into code that was written with Parralel foreach "out of habbit" in my past consulting life that I wasn't sure you had evaluated the TPL option.

CodeSnippetsReflection.App/Program.cs
zengin: Thanks, added `Trim()` to handle that.

CodeSnippetsReflection.App/Program.cs
zengin: Thanks, added.

nikithauc (approved):

baywet (approved):

PR #329, opened 2020-10-08
Turns off telemetry

irvinesunday:
This PR turns off telemetry.


GraphWebApi/Startup.cs
```diff
@@ -63,14 +63,13 @@ public void ConfigureServices(IServiceCollection services)
             services.AddApplicationInsightsTelemetry(options =>
             {
                 options.InstrumentationKey = Configuration["ApplicationInsights:InstrumentationKey"];
```
ddyett, line 65: pull the setting of this just in case any other option still sends something.  this should make sure we capture nothing

ddyett (approved):

PR #366, opened 2020-11-04
- fixes a bug where non-flagabble enum properties would not be generated with proper type in java

### LINKED ISSUE #352: java: affecting values with enums should be using enumsets for flaggable properties
Whenever a property can contain a collection of enum values (equivalent of flaggable from dotnet in java), the snippet should be using an enumset to affect the value. https://docs.microsoft.com/en-us/graph/api/chatmessage-update?view=graph-rest-1.0&tabs=java

baywet:
fixes #352

https://github.com/microsoftgraph/microsoft-graph-docs/compare/snippet-generation/31225...snippet-generation/31231?expand=1

### DIFF
--- a/CodeSnippetsReflection/LanguageGenerators/JavaGenerator.cs
+++ b/CodeSnippetsReflection/LanguageGenerators/JavaGenerator.cs
@@ -338,10 +338,10 @@ namespace CodeSnippetsReflection.LanguageGenerators
                         return $"Base64.getDecoder().decode({AddQuotesIfMising(stringParameter)})";
 
                     case "Double":
-                        return $"{stringParameter}d";
+                        return $"{DefaultNumericValues(stringParameter)}d";
 
                     case "Int64":
-                        return $"{stringParameter}L";
+                        return $"{DefaultNumericValues(stringParameter)}L";
 
                     case "TimeOfDay":
                         return DateTime.TryParse(stringParameter, out var timeOfDay)
@@ -357,6 +357,7 @@ namespace CodeSnippetsReflection.LanguageGenerators
 
                     case "Int16":
                     case "Int32":
+                        return DefaultNumericValues(stringParameter);
                     default:
                         return stringParameter;
                 }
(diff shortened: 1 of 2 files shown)


CodeSnippetsReflection/LanguageGenerators/JavaGenerator.cs
```diff
@@ -257,14 +257,23 @@ private string JavaGenerateObjectFromJson(ODataPathSegment pathSegment, string j
...
+                            foreach (var element in array)
+                            {
+                                var enumString = GenerateEnumString(element.Value<string>(), pathSegment, path);
+                                stringBuilder.Append($"{currentListName}.add({enumString});\r\n");
+                            }
                         }
                         else
                         {
-                            stringBuilder.Append($"LinkedList<String> {path.Last()}List = new LinkedList<String>();\r\n");
+                            stringBuilder.Append($"LinkedList<String> {currentListName} = new LinkedList<String>();\r\n");
```
MIchaelMainer, line 272: I suggest that we either:
1) Make this an `elseif` that checks that this is an array of strings with a subsequent `else` that reports the identification that this as a new scenario that we need to support. For example, if we get an array of any OData primitive types that is not a string, we should identify the scenario and where we need to add the change. 
2) Add conditions for all OData primitive types.

I prefer the first for now.

Can we check that this is an array of strings?

MIchaelMainer (changes requested):

CodeSnippetsReflection/LanguageGenerators/JavaGenerator.cs
baywet: great minds connect! (did I properly translate the expression?) I was hunting for those kind of issues (wrong mapping of collections) and I was thinking about tackling the issue in a separate PR. But I included the fix here as well.

zengin (dismissed):

zengin (approved):

nikithauc (approved):

MIchaelMainer (approved):

PR #372, opened 2020-11-06
[Snippet Generation] Fix type inference when the value is null

### LINKED ISSUE #297: [Snippet Generation] Type inference needs a type hint in at least one of lvalue or rvalue
Following snippet doesn't compile, because `var physicalDeviceId = null` is not enough to resolve the type of `physicalDeviceId`.

```csharp
var displayName = "Test Printer";

var manufacturer = "Test Printer Manufacturer";

var model = "Test Printer Model";

var physicalDeviceId = null;

var hasPhysicalDevice = false;

var certificateSigningRequest = new PrintCertificateSigningRequestObject
{
    Content = "{content}",
    TransportKey = "{sampleTransportKey}"
};

var connectorId = null;

await graphClient.Print.Printers
    .Create(displayName,manufacturer,model,certificateSigningRequest,physicalDeviceId,hasPhysicalDevice,connectorId)
    .Request()
    .PostAsync();
```

https://docs.microsoft.com/en-us/graph/api/printer-create?view=graph-rest-beta&tabs=csharp#request

zengin:
fixes #297

In the samples, we see that some of the values for action parameters are `null`. And the default generation process relies on type inference for variables. However the statement
```csharp
var parameter = null;
```
is not good enough for type inference.

This PR adds a special case for these situations where `var` is replaced by the type from the action parameter, e.g.: 
```csharp
Int32? parameter = null;
```

When the type is already a nullable type (i.e. any custom Microsoft.Graph type or System types such as String), we skip adding question mark after the type name:
```csharp
String parameter = null;
```

Also evaluated the option of inlining null as an action parameter, but that required passing the context around, so the current fix ended up being more local.

[Snippet Generation Diff](https://github.com/microsoftgraph/microsoft-graph-docs/compare/snippet-generation/31285...snippet-generation/31331)

### DIFF
--- a/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
+++ b/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
@@ -118,11 +118,6 @@ namespace CodeSnippetsReflection.LanguageGenerators
                                     // follow the values in the comments with the example above
 
                                     var parameter = CommonGenerator.LowerCaseFirstLetter(key);  // "index"
-                                    var parameterType = os.Operations
-                                        .Single(o => o.Name == os.Identifier)                   // selects "add" action
-                                        .Parameters                                             // [bindparameter, index, values]
-                                        .Single(p => p.Name == parameter)                       // [index : Edm.Int32]
-                                        .Type;                                                  // Edm.Int32
 
                                     var path = new List<string> { CommonGenerator.LowerCaseFirstLetter(key) }; // { "index" }
                                     var value = CSharpGenerateObjectFromJson(segment, jsonString, path);       // null;
@@ -133,6 +128,12 @@ namespace CodeSnippetsReflection.LanguageGenerators
                                     // In these cases, we look for the type of "index" in action parameters.
                                     if (value.Trim() == "null;")
                                     {
+                                        var parameterType = os.Operations
+                                            .Single(o => o.Name == os.Identifier)                   // selects "add" action
+                                            .Parameters                                             // [bindparameter, index, values]
+                                            .Single(p => p.Name == parameter)                       // [index : Edm.Int32]
+                                            .Type;                                                  // Edm.Int32
+
                                         if (parameterType.IsNullable)
                                         {
                                             typeHintOnTheLeftHandSide = new CSharpTypeProperties(parameterType.Definition, false).ClassName; // Int32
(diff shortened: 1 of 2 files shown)


CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -79,15 +87,68 @@ public string GenerateCodeSnippet(SnippetModel snippetModel, LanguageExpressions
...
                             //deserialize the object since the json top level contains the list of parameter objects
                             if (JsonConvert.DeserializeObject(snippetModel.RequestBody) is JObject testObj)
                             {
                                 foreach (var (key, jToken) in testObj)
                                 {
                                     var jsonString = JsonConvert.SerializeObject(jToken);
-                                    snippetBuilder.Append($"var {CommonGenerator.LowerCaseFirstLetter(key)} = ");
-                                    snippetBuilder.Append(CSharpGenerateObjectFromJson(segment, jsonString, new List<string> { CommonGenerator.LowerCaseFirstLetter(key) }));
+
+                                    // example URL:
```
baywet, line 98: is it worth keeping the code as comment since we have a unit test for it? maybe point to the unit test instead?

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -79,15 +87,68 @@ public string GenerateCodeSnippet(SnippetModel snippetModel, LanguageExpressions
...
+                                    var path = new List<string> { CommonGenerator.LowerCaseFirstLetter(key) }; // { "index" }
+                                    var value = CSharpGenerateObjectFromJson(segment, jsonString, path);       // null;
+                                    var typeHintOnTheLeftHandSide = "var";
+
+                                    // If the value is null we can't resolve the type by using var on the left hand side.
+                                    // For example, we can't say "var index = null;". That won't be a compilable snippet.
+                                    // In these cases, we look for the type of "index" in action parameters.
+                                    if (value.Trim() == "null;")
+                                    {
+                                        var parameterType = os.Operations
```
baywet, line 131: null propagation in case the operation and/or the parameter is not found?

baywet (approved):

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
zengin: makes sense.

baywet (approved):

PR #379, opened 2020-11-23
Loads reference model CSDL into main CSDL

irvinesunday:
**This PR proposes:**

- Updating the code that translates a CSDL file into an EDM model by extending this functionality to be able to read referenced XML files into the main XML file.
- Adding unit tests to validate this capability.

This is useful in referencing, for example, MS Graph permissions annotations described in a separate XML file within the main XML file. 
Referencing the annotations in the main file is done as below: 
image

**Sample files:**
- Main metadata file: https://raw.githubusercontent.com/irvinesunday/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsv1.0.xml
- permissions annotations file: https://raw.githubusercontent.com/irvinesunday/msgraph-metadata/master/scopes_metadata/apiPermissionsAndScopes-v1.0.xml

**Tests:** 

Unit tests have been added to test the secondary model XML referencing capability, however these currently **fail**.

This referencing currently as-is structured reads the annotations described from the secondary file but doesn't load these into the final EDM model. Further investigation is needed to establish whether the structure of the permissions annotations file could be the cause for this.


irvinesunday:
Closing this PR as there's no current business justification for this feature.

PR #384, opened 2020-12-03
[Snippet Generation] Add custom metadata support

### LINKED ISSUE #383: Snippet generation should consume Edm model from arbitrary metadata
Snippet generation relies on type information from metadata and models are currently hardcoded to clean metadata that we publish in `msgraph-metadata` repo.

This results in inconsistent type names if the type name changes in arbitrary metadata inputs.

I hit into one such instance where beta metadata had a type "ediscoveryCase" where the same type changed into "case" in staging beta metadata.

Staging Run failed with the following error, because snippet generation used the public clean metadata which had "ediscoveryCase" as the type, but we generated the SDK using staging metadata which had "case" as the type.
```
 1 using System;
 2 using Microsoft.Graph;
 3 using MsGraphSDKSnippetsCompiler;
 4 using Newtonsoft.Json.Linq;
 5 using System.Collections.Generic;
 6 using System.Text;
 7 using System.Threading.Tasks;
 8 
 9 // Disambiguate colliding namespaces
10 using DayOfWeek = Microsoft.Graph.DayOfWeek;
11 using TimeOfDay = Microsoft.Graph.TimeOfDay;
12 using KeyValuePair = Microsoft.Graph.KeyValuePair;
13 
14 public class GraphSDKTest
15 {
16     private IAuthenticationProvider authProvider = null;
17 
18     private async void Main()
19     {
20         authProvider = AuthenticationProvider.GetIAuthenticationProvider();
21 
22         #region msgraphsnippets
23         
24 
25         GraphServiceClient graphClient = new GraphServiceClient( authProvider );
26 
27         var ediscoveryCase = new EdiscoveryCase
28         {
29             DisplayName = "My Case 1"
30         };
31 
32         await graphClient.Compliance.Ediscovery.Cases
33     .Request()
34     .AddAsync(ediscoveryCase);
35 
36         
37         #endregion
38     }
39 }
CS0246: (Line:27, Column:33) The type or namespace name 'EdiscoveryCase' could not be found (are you missing a using directive or an assembly reference?)
```

Proposed solution involves 3 steps:
1. Add optional parameter to snippet generation console app for custom metadata path
1. Add similar optional parameter to apidoctor and plumb the value down to snippet generation tool
1. Update Raptor pipelines to provide custom metadata to the snippet generation tool via apidoctor

zengin:
Fixes #383

This PR includes step 1 of the proposed solution in the issue description above.

- Adds a new command line parameter for custom metadata inputs
- Consumes the parameter mentioned above if specified to initialize an Edm model.

Related
https://github.com/microsoftgraph/apidoctor/pull/4
https://github.com/microsoftgraph/msgraph-sdk-raptor/pull/117

### DIFF
--- a/CodeSnippetsReflection/SnippetsGenerator.cs
+++ b/CodeSnippetsReflection/SnippetsGenerator.cs
@@ -1,4 +1,5 @@
 using System;
+using System.IO;
 using System.Net.Http;
 using System.Xml;
 using Microsoft.OData.Edm;
@@ -23,6 +24,12 @@ namespace CodeSnippetsReflection
 
         private Lazy<IEdmModel> IedmModelV1 { get; set; }
         private Lazy<IEdmModel> IedmModelBeta { get; set; }
+
+        /// <summary>
+        /// initialized only if custom metadata path is specified in constructor
+        /// </summary>
+        private Lazy<IEdmModel> CustomEdmModel { get; set; }
+
         private Uri ServiceRootV1 { get; set; }
         private Uri ServiceRootBeta { get; set; }
         private JavascriptExpressions JavascriptExpressions { get; }
@@ -33,9 +40,10 @@ namespace CodeSnippetsReflection
         /// <summary>
         /// Class holding the Edm model and request processing for snippet generations
         /// </summary>
-        public SnippetsGenerator()
+        /// <param name="customMetadataPath">Full file path to the metadata</param>
+        public SnippetsGenerator(string customMetadataPath = null)
         {
-            LoadGraphMetadata();
+            LoadGraphMetadata(customMetadataPath);
             JavascriptExpressions = new JavascriptExpressions();
             CSharpExpressions = new CSharpExpressions();
             ObjectiveCExpressions = new ObjectiveCExpressions();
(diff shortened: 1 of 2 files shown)


CodeSnippetsReflection/SnippetsGenerator.cs
```diff
@@ -99,10 +126,10 @@ public string ProcessPayloadRequest(HttpRequestMessage httpRequestMessage, strin
             switch (requestUri.Segments[1])
             {
                 case "v1.0/":
-                    return (IedmModelV1.Value, ServiceRootV1);
+                    return ((CustomEdmModel ?? IedmModelV1).Value, ServiceRootV1);
```
andrueastman, line 129: Just to confirm. We would be using the same metadata for both v1 and beta? Or will the tool be run separately on both workloads?

zengin: The tool is used separately for different workloads, so there is a single metadata for each run.

This will result in generating v1 snippets using the beta model for a beta run and vice versa. But this is not an issue given that snippet generation result will be consumed only within the pipeline and only beta snippets will be used in the subsequent steps for a beta run. Same goes for v1 too.

I thought about specifying for each version for more clarity, but a combined run is not a use case at the moment.

Weekly generation, which is a combined run for public docs, will continue to rely on public clean metadata as it won’t specify the optional custom metadata parameter.

andrueastman: Thanks for that.

andrueastman (approved):

baywet (approved):

MIchaelMainer (approved):

PR #386, opened 2020-12-03
Forward Integration [master -> dev]

zengin:


baywet (approved):
I'm not sure this is needed, there are no changes besides the merge commits.

zengin:
@baywet this causes a circular dependency issue if there is a rule for PRs blocking merge if the branches are not in sync and the only way out was a forced merge bypassing that rule in at least one direction. It happened to me in this repo before. I am forwarding you the related thread.

That said, I am not sure why I didn't get that warning in the last PR #385

MIchaelMainer (approved):

PR #407, opened 2020-12-21
- adds a step to remove all snippets that are not referenced anymore

baywet:
- adds a script and a step to cleanup snippets that are not in use anymore

Successful run : https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=35531&view=results

Result: https://github.com/microsoftgraph/microsoft-graph-docs/compare/snippet-generation/35531?expand=1

### DIFF
--- a/scripts/cleanupUnusedSnippets.ps1
+++ b/scripts/cleanupUnusedSnippets.ps1
@@ -3,11 +3,11 @@ $docsFiles = Get-ChildItem -Recurse -Filter *.md -Exclude */snippets/*
 $resolvedSnippetsFiles = New-Object System.Collections.Generic.HashSet[string]
 foreach ($docFile in $docsFiles) {
 	$content = Get-Content -Path $docFile.FullName 
-	if($content -cmatch ".*INCLUDE.*") {
-		$pathMatches = [regex]::Matches($content, "\(([^\)]+)\)");
+	if($content -cmatch ".*INCLUDE.*") { # matches this in the referencing files [!INCLUDE [sample-code](../includes/snippets/csharp/get-accesspackageassignmentresourcerole-csharp-snippets.md)]
+		$pathMatches = [regex]::Matches($content, "\(([^\)]+)\)"); # gets the path portion of a MD link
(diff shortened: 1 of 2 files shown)


scripts/cleanupUnusedSnippets.ps1
```diff
@@ -0,0 +1,21 @@
+$snippetFiles = Get-ChildItem -Recurse -Filter snippets | Get-ChildItem -Recurse -Filter *.md | Select-Object -ExpandProperty FullName
```
zengin, line 1: I respect the time you have spent to write this logic with regular expressions, but I believe it can be easier to reason about and maintain if we have the steps as: clean all the snippet folders -> regenerate all the snippets.

What do you think?

scripts/cleanupUnusedSnippets.ps1
```diff
@@ -0,0 +1,21 @@
...
+$docsFiles = Get-ChildItem -Recurse -Filter *.md -Exclude */snippets/*
+$resolvedSnippetsFiles = New-Object System.Collections.Generic.HashSet[string]
+foreach ($docFile in $docsFiles) {
+	$content = Get-Content -Path $docFile.FullName 
+	if($content -cmatch ".*INCLUDE.*") {
+		$pathMatches = [regex]::Matches($content, "\(([^\)]+)\)");
```
MIchaelMainer, line 7: nit: can you add a note about what the regex matches? same below

scripts/cleanupUnusedSnippets.ps1
baywet: I thought about that and then I thought that some people might have created samples manually, in which case we'd delete them and not restore them. This is why I thought I needed to build an index of files in use and compare the files I found against that index

scripts/cleanupUnusedSnippets.ps1
baywet: done!

zengin (approved):

PR #411, opened 2021-01-07
feature/separate-cleanup

baywet:
This pull request separates the unused snippets cleanup I added recently from the actual snippet generation.
This is because in order to make sure we're not deleting snippets that are referenced, we need to parse a lot of markdown docs, and this takes almost as much time as the generation process. It also doesn't need to happen at the same time/candence than the generation process.

So I separated it to make tests generation cycles faster.

@MIchaelmainer I don't have permissions to add a new pipeline for that repo, can you add the new one (uused-snippets) under that folder for me please? https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build?definitionScope=%5CSnippet%20Generation

Edit: this also adds a conditional statement to avoid trying to commit and push things when there are no changes. This way the pipeline doesn't fail anymore for expected reasons

### DIFF
--- a/pipelines/unused-snippets.yml
+++ b/pipelines/unused-snippets.yml
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-# Pipeline for Snippet injection into docs
+# Pipeline for removing unused snippets from the docs repo
 
 trigger: none
 pr: none
--- a/pipelines/unused-snippets.yml
+++ b/pipelines/unused-snippets.yml
@@ -1,0 +1,44 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Pipeline for Snippet injection into docs
+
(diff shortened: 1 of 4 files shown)


pipelines/unused-snippets.yml
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# Pipeline for Snippet injection into docs
```
zengin, line 4:
```suggestion
# Pipeline for removing unused snippets from the docs repo
```

zengin (approved):

MIchaelMainer (approved):
Add pipeline

baywet:
started a run here https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=36462&view=results

baywet:
@MIchaelMainer I'm logging off for the day, if the pipeline completes successfully, would you mind merging the PR please? thanks!

PR #420, opened 2021-01-19
Update pipelines Tests

andrueastman:
Ensure tests are run in pipelines after net5.0 upgrade


irvinesunday (approved):

baywet:
Oops I forgot to check the CI for the unit tests since it wasn't falling. Sorry about that and thanks for fixing it!

> Note: you could upgrade to the new cross platform tasks and move to Ubuntu agent for better performances. See typewriter's pipelines for example.

PR #429, opened 2021-01-27
Use NewtonsoftJson as default converter in upgraded .NET 5.0 project

irvinesunday:
The upgraded .NET 5.0 project uses `System.Text.Json` as the default json converter, this PR adds `NewtonsoftJson` as the default converter.


andrueastman (approved):
For context, the enum fields(i.e method) were being deserialized into integer values rather than strings values.
**Before fix.** 
image
**After fix**
image

PR #439, opened 2021-02-03
Snippets support for System.Text.Json

andrueastman:
This PR contains snippet changes required for System.Text.Json in .Net

Example code generation can be found at https://github.com/microsoftgraph/microsoft-graph-docs/compare/main...snippet-generation/52453

Note: To be merged in after https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/382


CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -390,8 +390,8 @@ private string CSharpGenerateObjectFromJson(ODataPathSegment pathSegment, string
             if (className == "JToken") // handle special case of JToken
             {
                 var jToken = jsonObject is string _
-                    ? jsonBody                            // simple strings auto cast to JToken
-                    : $"JToken.Parse(\"{jsonBody}\")";    // complex objects need JToken.Parse()
+                    ? $"JsonDocument.Parse(@\"\"{jsonBody}\"\")"     // parse string differently
```
baywet, line 393: don't you need a dollar sign for the string interpolation to work here?

baywet (changes requested):

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
andrueastman: Sorry, do you mean inside the `JsonDocument.Parse` method?
The `JsonDocument.Parse` is not actual code but a string of the snippet to be generated. So what goes inside is just a string literal that will be interpolated since there is already a dollar sign before the `JsonDocument.Parse`

baywet: right, I read that wrong, thanks for clarifying.
shouldn't it be `$"JsonDocument.Parse(@\"\\\"{jsonBody}\\\"\")"` so you get escaping backslashes in the output code then?

andrueastman: In this case since we have the string literal starting with the `@` character, the right way to escape the quotation marks would be to use other quotation marks. Since the `jsonObject` is already confirmed to be json string object, it will have the expected quotation marks for it to be valid json.
You can view a generated sample of this in the diff at the link below.
https://github.com/microsoftgraph/microsoft-graph-docs/compare/main...snippet-generation/52453

baywet: right, ignore my confused mind. Adding @ in front of $ and removing all the \ would have made it easier to parse. but that's a tiny detail.

baywet (approved):

Issue #412, opened 2021-01-08
Permission API throws 500 when Accept-Language header is not specified


zengin:
Permissions API throws an exception if Accept-Language header is not specified even though there is a default locale defined in the code.

Expected result:
Permissions API should consume default locale when locale is not specified with Accept-Language. 
[AB#7004](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7004)

PR #455, opened 2021-02-10
Adds Changes endpoint

irvinesunday:
**Summary**
The Changes endpoint will be used primarily by Graph Explorer to surface changelog info. The changelog info describes Microsoft Graph API endpoint changes that have occurred in both V1.0 and Beta. Graph Explorer will call this Changelog endpoint to retrieve any changelog info. associated with a particular Microsoft Graph endpoint.

Changelog info can be retrieved by:
- Microsoft Graph url
- Workload name
- Dates
- Number of days

The response can also be paginated.

**This PR proposes**:
- Adding in a new `/Changes` endpoint.
- Updating the OpenAPI description with the new `/Changes` endpoint.
- Adding a `ChangesService` service library.
- Updating the `appsettings.json` with new configs for the `ChangesService` service library 
- Adding tests to validate the `ChangesService` service library.

Closes [AB#6825](https://dev.azure.com/microsoftgraph/Graph%20Developer%20Experiences/_workitems/edit/6825/)
Closes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/457
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/412

Requirements doc: https://microsofteur-my.sharepoint.com/:w:/g/personal/bengugi_microsoft_com/ETcytiFdE_tMo7WNqgWTzbIBfZCwjLf_lGL2ckFhCXbGUQ?e=PxywbK


ChangesService/Common/ChangesServiceConstants.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <summary>
+    /// Constants for the ChangesService library.
+    /// </summary>
+    public static class ChangesServiceConstants
+    {
+        // Changelog configuration paths constants
+        public const string ChangelogRelativeUrlConfigPath = "BlobStorage:Blobs:ChangeLog:RelativeUrl";
+        public const string ChangelogBaseUrlConfigPath = "BlobStorage:Blobs:ChangeLog:BaseUrl";
+        public const string ChangelogRefreshTimeConfigPath = "FileCacheRefreshTimeInHours:ChangeLog";
+        public const string ChangelogRefreshTimeDefault = "24";
```
ddyett, line 16: is this one used somewhere?

ChangesService/Models/ChangeLogList.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <summary>
+        /// The list of changelogs.
+        /// </summary>
+        [JsonProperty(PropertyName = "ChangeLog")]
+        public List<ChangeLog> ChangeLogs
+        {
+            get
+            {
+                return _changeLogs;
+            }
```
ddyett, line 27: consider get =>  _changeLogs

ChangesService/Models/MicrosoftGraphProxyConfigs.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace ChangesService.Models
+{
+    /// <summary>
+    /// Configs for connecting to the Microsoft Graph Proxy to
+    /// fetch workload names from Graph urls.
+    /// </summary>
+    public class MicrosoftGraphProxyConfigs
+    {
+        public string GraphVersion { get; set; } = "v1.0";
```
ddyett, line 13: consider making these init only {get; init; } =  "v1.0"

ChangesService/Models/ChangeLogList.cs
```diff
@@ -0,0 +1,102 @@
+﻿// ------------------------------------------------------------------------------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------------------------------------------------------------------------------
+
+using Newtonsoft.Json;
```
ddyett, line 5: how far off are we with work to move away from newtonsoft.

ChangesService/Models/ChangeLogList.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Linq;
+
+namespace ChangesService.Models
+{
+    /// <summary>
+    /// A <see cref="ChangeLog"/> list.
+    /// </summary>
+    public class ChangeLogList : ChangeLogPagination
+    {
+        private List<ChangeLog> _changeLogs;
```
ddyett, line 16: i don't see anything to add,remove, etc on this list so it looks like there's a set up to create lists and replace them

ChangesService/Services/ChangesService.cs
```diff
@@ -0,0 +1,251 @@
...
+
+namespace ChangesService.Services
+{
+    /// <summary>
+    /// Utility functions for transforming and filtering <see cref="ChangeLogList"/> and <see cref="ChangeLog"/> objects.
+    /// </summary>
+    public static class ChangesService
+    {
+        // Field to hold key-value pairs of url and workload names
+        private static readonly IDictionary<string, string> _urlWorkloadDict = new Dictionary<string, string>();
```
ddyett, line 25: since now on C# 9 consider

private static readonly Dictionary<string, string> _urlWorkloadDict = new();

ChangesService/Services/ChangesService.cs
```diff
@@ -0,0 +1,251 @@
...
+            if (graphProxyConfigs == null)
+            {
+                throw new ArgumentNullException(nameof(graphProxyConfigs), ChangesServiceConstants.ValueNullError);
+            }
+
+            // Temp. var to hold cascading filtered data
+            ChangeLogList tempChangeLogList = changeLogList;
+
+            // Search the changelog list by workload name
+            ChangeLogList changeLogListByWorkload = new ChangeLogList();
```
ddyett, line 80: why do you need this variable when tempChangeLogList can just be assigned the filter return directly

ChangesService/Services/ChangesService.cs
```diff
@@ -0,0 +1,251 @@
...
+
+            if (!string.IsNullOrEmpty(searchOptions.RequestUrl))
+            {
+                // Retrieve the workload name from the requestUrl
+                var workload = RetrieveWorkloadNameFromRequestUrl(searchOptions, graphProxyConfigs, httpClientUtility)
+                                .GetAwaiter().GetResult();
+
+                // Search by retrieved workload name
+                changeLogListByWorkload.ChangeLogs = FilterChangeLogListByWorkload(tempChangeLogList,
+                                                                      workload).ToList();
```
ddyett, line 90: the code here looks inefficient.  you filter, then convert to the list, then potentially filter again for date and convert to list again.  Why not operate on the IEnumerable then convert later on?

ChangesService/Services/ChangesService.cs
```diff
@@ -0,0 +1,251 @@
...
+                    /* The first page of several pages or
+                     * the first page of only one page
+                     */
+
+                    tempChangeLogList.Page = 1;
+                    tempChangeLogList.ChangeLogs = tempChangeLogList.ChangeLogs
+                                                    .Take(searchOptions.PageLimit.Value)
+                                                    .ToList();
+                }
+                else if (searchOptions.Page < tempChangeLogList.TotalPages)
```
ddyett, line 145: there seems to be some inefficiency in that  when I call for page 1 we do all that filtering work, then when I call for page 2 we do all that filtering work again.  Just a lot of list being created and destroyed.

ChangesService/Services/ChangesStore.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        /// <summary>
+        /// Fetches a <see cref="ChangeLog"/> list from a uri source or an in-memory cache.
+        /// </summary>
+        /// <param name="locale">The language code for the preferred localized file.</param>
+        /// <returns>A <see cref="ChangeLogList"/> containing the list of <see cref="ChangeLog"/>.</returns>
+        public async Task<ChangeLogList> FetchChangeLogListAsync(string locale)
+        {
+            if (string.IsNullOrEmpty(locale) || locale.Equals("en-gb", StringComparison.OrdinalIgnoreCase))
```
ddyett, line 50: why call out en-gb in particular since there are several other english variants?

ChangesService/Services/ChangesStore.cs
```diff
@@ -0,0 +1,101 @@
...
+                    {
+                        // Already seeded by another thread
+                        return Task.FromResult(seededChangeLogList);
+                    }
+
+                    // Set cache expiry
+                    cacheEntry.AbsoluteExpirationRelativeToNow = TimeSpan.FromHours(_defaultRefreshTimeInHours);
+
+                    // Construct the locale-specific relative uri
+                    string relativeUrl = string.Format(_changeLogRelativeUrl, locale);
```
ddyett, line 80: try to be consistent if using var or using type

FileService/Common/FileServiceConstants.cs
```diff
@@ -10,5 +12,19 @@ namespace FileService.Common
...
+        {
+            public enum Headers
+            {
+                Accept,
+                Authorization,
+                [Description("User-Agent")]
+                UserAgent
+            }
+
+            public static string ApplicationJsonMediaType = "application/json";
```
ddyett, line 26: make readonly.

FileService/Services/HttpClientUtility.cs
```diff
@@ -39,14 +39,9 @@ public async Task<string> ReadFromDocument(HttpRequestMessage requestMessage)
             requestMessage.Method = requestMessage.Method ?? HttpMethod.Get; // default is GET
 
             var httpResponseMessage = await _httpClient?.SendAsync(requestMessage);
-            var fileContents = await httpResponseMessage.Content.ReadAsStringAsync();
+            var fileContents = await httpResponseMessage?.Content.ReadAsStringAsync();
```
baywet, line 42:
```suggestion
            var fileContents = await httpResponseMessage?.Content?.ReadAsStringAsync();
```

FileService/Services/HttpClientUtility.cs
```diff
@@ -39,14 +39,9 @@ public async Task<string> ReadFromDocument(HttpRequestMessage requestMessage)
             requestMessage.Method = requestMessage.Method ?? HttpMethod.Get; // default is GET
 
             var httpResponseMessage = await _httpClient?.SendAsync(requestMessage);
```
baywet, line 41:
```suggestion
            using var httpResponseMessage = await _httpClient?.SendAsync(requestMessage);
```

baywet (changes requested):

ChangesService/Services/ChangesService.cs
irvinesunday: Directly altering the `tempChangeLogList` object will inadvertently alter the param object `changeLogList` since it is assigned its reference. An improved approach could be to declare the `ChangeLogList` variables and `new()` them up when necessary within the code blocks.

```
            ChangeLogList changeLogListByWorkload; // declare here

            if (!string.IsNullOrEmpty(searchOptions.RequestUrl))
            {
                // Retrieve the workload name from the requestUrl
                var workload = RetrieveWorkloadNameFromRequestUrl(searchOptions, graphProxyConfigs, httpClientUtility)
                                .GetAwaiter().GetResult();

                changeLogListByWorkload = new(); // instantiate here

                // Search by retrieved workload name
                changeLogListByWorkload.ChangeLogs = FilterChangeLogListByWorkload(tempChangeLogList,
                                                                      workload).ToList();

                tempChangeLogList = changeLogListByWorkload;
            }
```
So we don't end up creating lists we don't use.

ChangesService.Test/ChangeLogSearchOptionsModelShould.cs
```diff
@@ -0,0 +1,41 @@
...
+            var startDate = DateTime.Parse("2020-06-01");
+            var endDate = DateTime.Parse("2020-12-31");
+
+            // Act & Assert -> both startDate and endDate specified + daysRange
+            Assert.Throws<InvalidOperationException>(() =>
+                new ChangeLogSearchOptions(startDate: startDate, endDate: endDate, daysRange: 90));
+
+            // Act & Assert -> just startDate specified + daysRange
+            Assert.Throws<InvalidOperationException>(() =>
+                new ChangeLogSearchOptions(startDate: startDate, daysRange: 90));
```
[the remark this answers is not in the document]
irvinesunday: So technically speaking, when the `daysRange` is specified, the `endDate` will be the current day's date while the `startDate` will be calculated from `endDate` - `daysRange`. 
1. If you want data for the past x number of days up to current day, you just specify daysRange as x
2. If you want data for a given time duration, you specify the `startDate` and `endDate`. I'm guessing someone might have the `startDate` and just wants to get data for the following x number of days without needing to explicitly specifying the `endDate`. This is the scenario where your question falls into?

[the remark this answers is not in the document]
baywet: I think that cover the case where only daysRange is specified. I feel like we could support specifying daysRange + startDate and daysRange + endDate without throwing an exception back to the consumer.

ChangesService/Interfaces/IChangesStore.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading.Tasks;
+
+namespace ChangesService.Interfaces
+{
+    /// <summary>
+    /// A contract for fetching changelog data.
+    /// </summary>
+    public interface IChangesStore
+    {
+        Task<ChangeLogList> FetchChangeLogListAsync(string locale);
```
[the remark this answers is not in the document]
irvinesunday: Thanks for the tip. I can implement the CultureInfo class within the existing method as below for the specific use case of checking whether locale values are formed correctly and for picking out English culture variations.

```
        public async Task<ChangeLogList> FetchChangeLogListAsync(string locale)
        {
            var englishId = "en-us"; // default locale
            if (string.IsNullOrEmpty(locale))
            {
                locale = englishId;
            }

            // Check for English CultureInfo variants
            if (new CultureInfo(locale).TwoLetterISOLanguageName.Equals("en", StringComparison.OrdinalIgnoreCase))
            {
                // All English CultureInfo variants
                // should default to 'en-us'
                locale = englishId;
            }
        }
```

In your suggestion though, how different would the method with a default implementation accepting a `CultureInfo` be?

FileService/Common/FileServiceConstants.cs
irvinesunday: Resolved with `const`

ChangesService/Services/ChangesStore.cs
irvinesunday: Good call.

ChangesService/Common/ChangesServiceConstants.cs
irvinesunday: No. Redundant.

ChangesService/Models/ChangeLogList.cs
irvinesunday: Are you inferring that the choice of the class name is misleading?

ChangesService/Services/ChangesService.cs
irvinesunday: Resolved: `tempChangeLogList` replaced with an `IEnumerable<ChangeLog>`.

ChangesService/Services/ChangesService.cs
irvinesunday: Resolved: List is created at the end of all filtering

ChangesService/Services/ChangesStore.cs
irvinesunday: Resolved: All variants of supported locales are defaulted.

ChangesService/Services/ChangesService.cs
irvinesunday: Resolved with the `IEnumerable<ChangeLog> enumerableChangeLog` object

ChangesService/Models/ChangeLogList.cs
irvinesunday: What is the motivation for wanting to move the project away from `Newtonsoft.Json`?

ChangesService/Interfaces/IChangesStore.cs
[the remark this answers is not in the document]
baywet: Back in the aspnet classic days (before core) the accept-language request header used to set the current ui thread culture so the MVC stack could switch between resources.
I believe this is still the case with core if you add this middleware https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.localization.requestlocalizationmiddleware?view=aspnetcore-5.0

In which case, the method in the RequestHelper is obsolete, and instead of passing the culture as a string, the process could rely on the current ui thread culture.

The other advantage of using cultures over plain string is because of their capability to parse the culture names for us, and link cultures to their parents. I don't believe we should be manipulating strings anymore, just looking at the two/three letters language name and/or the parent when we don't have that specific culture.

Additionally, because we don't have specialized locales files, I believe our files should be stored as the main culture for all the locales (i.e. fr instead of fr-fr). This way we'd match [the common fallback behavior](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization?view=aspnetcore-5.0#culture-fallback-behavior-2) prescribed by dotnet for the last decades.
This means that instead of building [a switch case](https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/455/files#diff-bd8fedd6df77a56a7aa4adcc8e7495e2ed9f7ac5506f7dfe0b0c48b4fba0fb57R60) we should:

1. Read the thread culture
2. Check for the presence in the cache or on disk of the xx-XX variant (return it if we find it)
3. Check for the presence in the cache or on disk of the xx variant (return it if we find it)
4. Check for the presence in the cache or on disk of the en variant (default fallback)

Now that being said. I can understand that we don't anticipate to provide a variant for fr-CA as opposed to fr-FR. So we could imagine skip step 2. (although our [public docs](https://docs.microsoft.com/en-us/locale/?target=https://docs.microsoft.com/en-us/graph/throttling) do make the distinction between fr-CA, fr-CH, fr-FR, and fr-BE)

The reason I'm being so specific is because I've worked a couple of years on [a localization product for SharePoint](https://oceanik.com/) and I've had to dive in the inner details of how localization works in dotnet, the specific in SharePoint, Windows, browsers... One of the key aspects to a functional localization process is to make sure it's implemented consistently with the platform the application runs on. Otherwise you'll run into weird edge cases where for some reason the localization is not working as expected for this or that locale.

Lastly, instead of building the cache and language switch mechanism ourselves, we could rely on resx, they do pre-load the content in memory and they do the switch of resource based on the UI thread. Maybe something to explore in addition to what I was mentioning above?

ChangesService/Models/ChangeLogList.cs
irvinesunday: Here's the issue to track this: https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/486

ChangesService/Services/ChangesService.cs
```diff
@@ -0,0 +1,251 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace ChangesService.Services
+{
+    /// <summary>
+    /// Utility functions for transforming and filtering <see cref="ChangeLogList"/> and <see cref="ChangeLog"/> objects.
+    /// </summary>
+    public static class ChangesService
```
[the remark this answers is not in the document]
irvinesunday: The behaviour and responsibilities exposed by this class are essentially utility functions; I did not see any reason to create an interface that defines functionalities to be inherited (and implemented) by this class (and consequently injected through DI) because there would be no need for creating a different class that would implement the interface definitions differently. 
This class currently has methods that test its functionality.

ChangesService/Models/ChangeLogList.cs
irvinesunday: Changed to `ChangeLogRecords`

ChangesService/Models/ChangeLogList.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        public ChangeLogList()
+        {
+            ChangeLogs = new List<ChangeLog>();
+        }
+
+        /// <summary>
+        /// Updates the TotalPages property.
+        /// </summary>
+        private void UpdateTotalPages()
```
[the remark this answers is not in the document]
irvinesunday: @baywet I didn't quite get that, please explain further.

ChangesService.Test/ChangeLogPaginationModelShould.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace ChangesService.Test
+{
+    public class ChangeLogPaginationModelShould
+    {
+        [Fact]
+        public void ThrowArgumentExceptionIfPageValueIsZeroOrNegativeInteger()
+        {
+            // Arrange
+            var pagination = new ChangeLogPagination();
```
thewahome, line 16: This arrangement is repeated in the next test. It can go into a setup function

ChangesService/Services/ChangesService.cs
```diff
@@ -0,0 +1,267 @@
...
+        /// </summary>
+        /// <param name="changeLogRecords">The <see cref="ChangeLogRecords"/> with the target
+        /// <see cref="ChangeLog"/> entries.</param>
+        /// <param name="searchOptions">The <see cref="ChangeLogSearchOptions"/> containing options for filtering
+        /// and paginating the target <see cref="ChangeLog"/> entries.</param>
+        /// <param name="graphProxyConfigs">Configuration settings for connecting to the Microsoft Graph Proxy.</param>
+        /// <param name="httpClientUtility">Optional. An implementation instance of <see cref="IHttpClientUtility"/>.</param>
+        /// <returns><see cref="ChangeLogRecords"/> containing the filtered and/or paginated
+        /// <see cref="ChangeLog"/> entries.</returns>
+        public static ChangeLogRecords FilterChangeLogRecords(ChangeLogRecords changeLogRecords,
```
thewahome, line 56: Consider breaking down this function into smaller private functions that get called by this public function.

ChangesService/Services/ChangesService.cs
```diff
@@ -0,0 +1,267 @@
...
+
+                    filteredChangeLogRecords.Page = filteredChangeLogRecords.TotalPages;
+
+                    int lastItems = filteredChangeLogRecords.ChangeLogs.Count() % searchOptions.PageLimit.Value;
+                    enumerableChangeLog = enumerableChangeLog.TakeLast(lastItems);
+                }
+
+                // Update with the paginated result
+                filteredChangeLogRecords.ChangeLogs = enumerableChangeLog.ToList();
+            }
```
thewahome, line 171: This can be extracted to its own function called paginateFilteredResults so that the `FilterChangeLogRecords` is smaller and easier to understand

ChangesService/Services/ChangesService.cs
```diff
@@ -0,0 +1,267 @@
...
+            httpRequestMessage.Headers.Add(FileServiceConstants.HttpRequest.Headers.UserAgent.ToString(),
+                FileServiceConstants.HttpRequest.DevxApiUserAgent); // User Agent
+
+            // Fetch the request url workload info. content from Microsoft Graph
+            string workloadInfo = await httpClientUtility.ReadFromDocumentAsync(httpRequestMessage);
+
+            // Extract the workload name from the response content
+            JToken workloadInfoToken = JObject.Parse(workloadInfo);
+            var targetWorkloadId = (string)workloadInfoToken["TargetWorkloadId"];
+            var workloadName = targetWorkloadId.Split('.').Last();
```
thewahome, line 258: The comment may be rendered unnecessary if this block is turned into a self-documenting function

irvinesunday: I think the block is too small to warrant its own function. Furthermore, the function of the block is in line with the method name wherein it is contained.

ChangesService/Models/ChangeLogList.cs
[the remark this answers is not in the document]
baywet: instead of calling this update method (remove it) have a total pages and a total items property that lazy loads the value by calling count. The only additional complexity it so make sure you compare the previous page size and the current one to determine whether or not you need to recalculate that value before returning it.

Note for myself, this has moved to ChangeLogRecords

ChangesService/Services/ChangesService.cs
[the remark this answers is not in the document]
baywet: sure. Static instances usually lead to taking shortcuts and making thread safety as well as testability harder. But if you're happy with the current version, I'll leave it up to you.

ChangesService/Interfaces/IChangesStore.cs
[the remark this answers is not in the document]
baywet: thanks for implementing the changes, resolving

ChangesService/Models/ChangeLogList.cs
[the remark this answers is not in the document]
irvinesunday: Thanks for the clarification.
This is okay for the _TotalPages_ property. We can have the logic for getting the total pages in the getter. However for the _TotalItems_ property, its value needs to be set only once during the lifetime of an instance, and that's when the _ChangeLogs_ property is first initialized with a value. We use this property to get the total item count for _ChangeLogs_ before any pagination (which updates the _ChangeLogs_ count) is done. I think this can remain as is.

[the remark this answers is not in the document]
baywet: then use a readonly property instead?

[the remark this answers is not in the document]
irvinesunday: For _TotalPages_? This is updated to a readonly property. _TotalItems_ has a private setter. It needs the private setter to be able to capture the count of _ChangeLogs_ when this is first initialized.

FileService/Common/FileServiceHelper.cs
```diff
@@ -94,6 +68,29 @@ public static string GetLocalizedFilePathSource(string containerName, string def
             return $"{containerName}{FileServiceConstants.DirectorySeparator}{defaultBlobName}";
         }
 
+        /// <summary>
+        /// Gets the locale name of the supported <see cref="CultureInfo"/> variants.
+        /// </summary>
+        /// <param name="cultureInfo">The <see cref="CultureInfo"/> object to retrieve supported variant from.</param>
+        /// <returns>The locale name of the supported <see cref="CultureInfo"/> variant.</returns>
+        public static string GetSupportedLocaleVariant(CultureInfo cultureInfo)
```
baywet, line 76: remove in favor of the extension method

FileService/Extensions/LocalizationExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        /// <summary>
+        /// Gets the locale name of the supported <see cref="CultureInfo"/> variant.
+        /// </summary>
+        /// <param name="cultureInfoName">The name of the <see cref="CultureInfo"/> from which to retrieve the supported variant from.</param>
+        /// <returns>The locale name of the supported <see cref="CultureInfo"/> variant. The default locale is en-US.</returns>
+        public static string GetSupportedLocaleVariant(this string cultureInfoName)
+        {
+            var langName = new CultureInfo(cultureInfoName).TwoLetterISOLanguageName ?? "en"; // english is the default
```
baywet, line 44:
```suggestion
            return new CultureInfo(cultureInfoName).GetSupportedLocaleVariant();
```
and remove everything below

baywet (changes requested):
thanks for making the changes around localization. A few minor follow ups, mostly around code duplication

FileService/Common/FileServiceHelper.cs
irvinesunday: I missed this. Thank you

baywet (approved):

PR #461, opened 2021-02-10
Replace single-quotes in string literals with escape sequence for JS

### LINKED ISSUE #460: Query parameters with quotes result in invalid JavaScript snippet
It's common for `$filter` parameters to contain quotes around a value, like `$filter=mailNickName eq 'meganb'`. When generating snippets for queries like this, the JavaScript snippet is invalid, because it uses single-quotes for string literals and doesn't properly escape the single-quote in the value.

Repro:

```
POST https://localhost:44399/snippetgenerator?lang=javascript

GET /v1.0/users?$filter=mailNickName%20eq%20'meganb' HTTP/1.1
Host: graph.microsoft.com

{}
```

Response:

```
const options = {
	authProvider,
};

const client = Client.init(options);

let res = await client.api('/users')
	.filter('mailNickName eq 'meganb'')
	.get();
```

Notice the string literal in the `.filter` call. It should be one of the following:

```
'mailNickName eq \'meganb\''
"mailNickName eq 'meganb'"
`mailNickName eq 'meganb'`
```
 
[AB#8090](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8090)

jasonjoh:
- Following existing design in `LanguageExpressions`, added a property `SingleQuotesEscapeSequence`. Value is just `"'"` for C#, Java, and Obj-C, but is `"\\'"` for JavaScript.
- Updated `CommonGenerator.GenerateQuerySection` to call `.Replace("'", languageExpressions.SingleQuotesEscapeSequence)` on all string literals.
- Updated `JavaScriptGenerator.GenerateRequestSection` to call `Replace("'", "\\'")` on the `snippetModel.QueryString` to handle any custom query parameters that might have single-quotes in them.

It's probably a good idea to chain the `.Replace("'", languageExpressions.SingleQuotesEscapeSequence)` calls in `CommonGenerator.GenerateQuerySection` with a `.Replace("\"", languageExpressions.DoubleQuotesEscapeSequence)` to cover all the bases, but as that was outside the scope of #460, I left it alone.

Fixes #460

### DIFF
--- a/CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
+++ b/CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Text;
 using Newtonsoft.Json.Linq;
+using CodeSnippetsReflection.StringExtensions;
 
 namespace CodeSnippetsReflection.LanguageGenerators
 {
@@ -43,15 +44,14 @@ namespace CodeSnippetsReflection.LanguageGenerators
                     continue;
                 //append the header to the snippet
                 var valueString = value.First()
-                    .Replace("\"", languageExpressions.DoubleQuotesEscapeSequence)
-                    .Replace("'", languageExpressions.SingleQuotesEscapeSequence);
+                    .EscapeQuotesInLiteral(languageExpressions.DoubleQuotesEscapeSequence, languageExpressions.SingleQuotesEscapeSequence);
                 snippetBuilder.Append(string.Format(languageExpressions.HeaderExpression, key, valueString));
             }
             //Append any filter queries
             if (snippetModel.FilterFieldList.Any())
             {
                 var filterResult = string.Join(languageExpressions.FilterExpressionDelimiter, snippetModel.FilterFieldList)
-                    .Replace("'", languageExpressions.SingleQuotesEscapeSequence);
+                    .EscapeQuotesInLiteral(languageExpressions.DoubleQuotesEscapeSequence, languageExpressions.SingleQuotesEscapeSequence);
                 //append the filter to the snippet
                 snippetBuilder.Append(string.Format(languageExpressions.FilterExpression, filterResult));
             }
@@ -60,7 +60,8 @@ namespace CodeSnippetsReflection.LanguageGenerators
             if (!string.IsNullOrEmpty(snippetModel.SearchExpression))
             {
                 snippetBuilder.Append(string.Format(languageExpressions.SearchExpression, 
-                    snippetModel.SearchExpression.Replace("'", languageExpressions.SingleQuotesEscapeSequence)));
+                    snippetModel.SearchExpression
+                    .EscapeQuotesInLiteral(languageExpressions.DoubleQuotesEscapeSequence, languageExpressions.SingleQuotesEscapeSequence)));
             }
 
             //Append the expand section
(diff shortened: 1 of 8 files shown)


CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
```diff
@@ -42,42 +42,49 @@ public static string GenerateQuerySection(SnippetModel snippetModel, LanguageExp
                 if (key.ToLower().Equals("host",StringComparison.Ordinal))
                     continue;
                 //append the header to the snippet
-                var valueString = value.First().Replace("\"", languageExpressions.DoubleQuotesEscapeSequence);
+                var valueString = value.First()
```
baywet, line 45: instead of having this at different places and potentially forgetting one or the other at some place, can you refactor to a method please?

baywet (changes requested):

zengin:
Snippets Diff: https://github.com/microsoftgraph/microsoft-graph-docs/compare/snippet-generation/40065...snippet-generation/40068?expand=1 (note that the diff is affected by #463)

Compilation Errors in Raptor: https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=40070&view=ms.vss-test-web.build-test-results-tab
15 failures (no regression there)

jasonjoh:
@zengin For the snippets diff, there seem to be a lot there that aren't related to this change. Did this change break something in that regard?

For the Raptor errors, I don't have access to that AzDO instance. Can I get added?

baywet (approved):
thanks for taking the time to make the change

zengin:
@jasonjoh, extra diff is because of #463. We should fix that soon as that causes inconsistent documentation and a lot of noise in snippet generation PRs.

Can you try again to see if you can access the Raptor results?

zengin (approved):

jasonjoh:
Yes I can see it now. I'll take a look.

On the diff, is there something broken because of this PR? I'm not sure what I'm looking for :D.

jasonjoh:
Those Raptor errors seem unrelated to this PR unless I'm missing something.

zengin:
@jasonjoh

Raptor failures are unrelated and there is no regression (we already have 15 failures in the base branch). I shared the result for reference, and noted that there is no regression ;)

As for the diff, extra diff is not because of your changes, it is because of the documentation having block names that are not unique. Our generation process uses block names to identify snippets and runs in parallel for all the snippets, so the order is indeterministic. So we end up with different results in each run. We follow the issue with #463.

So all is good here if all the related changes in the diff look as expected to you. Sorry for the confusion.

jasonjoh:
Gotcha. The changes in the diff that seem related all look good to me.

zengin:
@jasonjoh awesome. Do you have the merge rights here, or do you want me to merge?

jasonjoh:
I do have merge rights but didn't want to bypass your processes. Besides the automated checks, are there any other requirements (certain number of reviewers approved, etc.?)

zengin:
In most repos, we are OK with one approval. For snippet generation stuff, you can go ahead and merge with automated checks and an approval from me, @baywet or @andrueastman.

We will need more automated checks for contributors from outside our team as some of the dependencies are not enforced, such as Raptor regressions and full end-to-end generation to the docs repo. I opened #464 to make this process better.

zengin:
And thanks a lot for your contributions!

PR #465, opened 2021-02-10
Updated \n to \r\n in string literals for consistency

### LINKED ISSUE #459: Generated JavaScript snippets use a mix of line-ending styles
Line endings should be consistent within the snippet. Having mixed makes it harder to work with the returned snippet.

Repro:

```
POST https://localhost:44399/snippetgenerator?lang=javascript

POST /v1.0/groups/18505c12-3495-4f1b-8c63-29d5c21b0daa/members/$ref
Host: graph.microsoft.com
Content-Type: application/json

{
  "@odata.id": "https://graph.microsoft.com/v1.0/directoryObjects/87d349ed-44d7-43e1-9a83-5f2406dee5bd"
}
```

Response:

```
"const options = {\n  authProvider,\n};\n\nconst client = Client.init(options);\n\nconst directoryObject = {\r\n  @odata.id: \"https://graph.microsoft.com/v1.0/directoryObjects/87d349ed-44d7-43e1-9a83-5f2406dee5bd\"\r\n};\r\n\r\nlet res = await client.api('/groups/18505c12-3495-4f1b-8c63-29d5c21b0daa/members/$ref')\n  .post(directoryObject);"
```

Notice there are some `\n`, and some `\r\n`. 
[AB#8089](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8089)

jasonjoh:
Used `\r\n` as the majority of newline constants seemed to use that sequence.

Fixes #459

### DIFF
--- a/CodeSnippetsReflection.Test/CommonGeneratorShould.cs
+++ b/CodeSnippetsReflection.Test/CommonGeneratorShould.cs
@@ -163,7 +163,7 @@ namespace CodeSnippetsReflection.Test
             var result = CommonGenerator.GenerateQuerySection(snippetModel, expressions);
 
             //Assert string is empty
-            Assert.Equal("\n\t.Select(\"displayName,givenName,postalCode\")", result);
+            Assert.Equal("\r\n\t.Select(\"displayName,givenName,postalCode\")", result);
         }
 
         [Fact]
@@ -179,7 +179,7 @@ namespace CodeSnippetsReflection.Test
             var result = CommonGenerator.GenerateQuerySection(snippetModel, expressions);
 
             //Assert string is empty
-            Assert.Equal("\n\t.Filter(\"startswith(givenName, 'J')\")", result);
+            Assert.Equal("\r\n\t.Filter(\"startswith(givenName, 'J')\")", result);
         }
 
         [Fact]
@@ -195,7 +195,7 @@ namespace CodeSnippetsReflection.Test
             var result = CommonGenerator.GenerateQuerySection(snippetModel, expressions);
 
             //Assert string is empty
-            Assert.Equal("\n\t.Search(\"Irene McGowen\")", result);
+            Assert.Equal("\r\n\t.Search(\"Irene McGowen\")", result);
         }
 
         [Fact]
@@ -211,7 +211,7 @@ namespace CodeSnippetsReflection.Test
             var result = CommonGenerator.GenerateQuerySection(snippetModel, expressions);
 
             //Assert string is empty
-            Assert.Equal("\n\t.Skip(20)", result);
+            Assert.Equal("\r\n\t.Skip(20)", result);
         }
 
         [Fact]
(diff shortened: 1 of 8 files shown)


zengin:
In Raptor, I am normalizing these into one format, but I have to double check whether I have any hidden dependency so that this doesn't break the error output for the compiled snippets.

zengin:
cc: @baywet in case there is any dependency in the Java Raptor tests.

jasonjoh:
Awesome! Did you go with `/r/n` or something else?

zengin:
Normalized to `\r\n`: https://github.com/microsoftgraph/msgraph-sdk-raptor/blob/dev/TestsCommon/BaseTestRunner.cs#L21

zengin (approved):

andrueastman (approved):

nikithauc (approved):

CodeSnippetsReflection.Test/CommonGeneratorShould.cs
```diff
@@ -63,7 +63,7 @@ public void GenerateQuerySection_ShouldReturnAppropriateJavascriptFilterExpressi
             var result = CommonGenerator.GenerateQuerySection(snippetModel, expressions);
 
             //Assert string is empty
-            Assert.Equal("\r\n\t.filter('startswith(givenName, 'J')')", result);
+            Assert.Equal("\r\n\t.filter('startswith(givenName, \\'J\\')')", result);
```
baywet, line 66: I believe those should be single backslashes, not double.

baywet (commented):
@zengin I don't think it'll impact java raptor. What I'm more concerned about is the diff we'll get running snippets generation on a linux based machine. Won't all the line returns show as diff?

andrueastman:
Just as an FYI though (in case its helpful). 
From the docs perspective, the line endings are usually normalized (with a git config [here](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/.gitattributes)) plus [this PR check](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/.github/workflows/eol-blocker.yml) exists to ensure consistency (at least if we are basing off the master branch of the docs.).
cc @baywet @zengin

zengin:
@andrueastman, when I use sources directly to generate snippets on the fly and run the raptor tests in the same container, I don't get to benefit from the normalization in the master branch of the docs repo. So this fix is helpful for both that scenario and for the response in Graph Explorer.

zengin:
@baywet, I think @andrueastman's response covers your question. Sorry I saw the question after my first reply.

CodeSnippetsReflection.Test/CommonGeneratorShould.cs
zengin: I think it should be double based on this PR: https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/461

baywet: oh because we're escaping the \ in dotnet to get one \ in "javascript" which escapes the '... got it. It was too meta and too early.

baywet (approved):

CodeSnippetsReflection.Test/CommonGeneratorShould.cs
jasonjoh: No, they have to be double. It's escaping the backslash, not the quote. In other words, we want the string to contain `\'J\'`.

Issue #468, opened 2021-02-12
Synchronous IO are disallowed


irvinesunday:
When you run a `POST` on the path `/openapi`, you get back an error: "synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead."

Example: 

`POST` https://localhost:44399/openapi?operationIds=*&style=PowerShell

RequestBody:
```
<edmx:Edmx xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx" Version="4.0">
<edmx:DataServices>
<Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Namespace="microsoft.graph" >

    <EntityContainer Name="TodoService">
        <EntitySet Name="Todos" EntityType="microsoft.graph.Todo">
        </EntitySet>
    </EntityContainer>
    
    <EntityType Name="Todo" HasStream="true">
            <Key>
                <PropertyRef Name="Id"/>
            </Key>
        <Property Name="Id" Type="Edm.String"/>
        <Property Name="Logo" Type="Edm.Stream"/>
        <Property Name="Description" Type="Edm.String"/>
    </EntityType>

</Schema>
</edmx:DataServices>
</edmx:Edmx>
```

NB: This is only experienced with `POST` requests and not `GET` requests. 
[AB#8126](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8126)

PR #469, opened 2021-02-12
Allows synchronous IO

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/468

The exception call stack is as below: 

Line 1: https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/c7a21c9b2211cf506c6bebfcc7a20b85501d974b/OpenAPIService/OpenApiService.cs#L378
Line 2: https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/c7a21c9b2211cf506c6bebfcc7a20b85501d974b/GraphWebApi/Controllers/OpenApiController.cs#L113

### DIFF
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -110,7 +110,7 @@ namespace GraphWebApi.Controllers
                     throw new InvalidOperationException($"Unsupported {nameof(graphVersion)} provided: '{graphVersion}'");
                 }
 
-                OpenApiDocument source = OpenApiService.ConvertCsdlToOpenApi(Request.Body);
+                OpenApiDocument source = await OpenApiService.ConvertCsdlToOpenApiAsync(Request.Body);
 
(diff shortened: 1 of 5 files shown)


baywet (changes requested):
Is there any specific reason why we don't use async methods instead of blocking the main thread?

irvinesunday:
> Is there any specific reason why we don't use async methods instead of blocking the main thread?

For the specific issue that this PR is fixing, we don't have an alternate `async` option for this [method ](https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/c7a21c9b2211cf506c6bebfcc7a20b85501d974b/OpenAPIService/OpenApiService.cs#L378)from the package (`Microsoft.Odata.Edm`)

baywet:
what about doing the following?

```CSharp
using var reader = new StreamReader(csdl);
var cdslTxt = await reader.ReadToEndAsync();
var edmModel = CsdlReader.Parse(XElement.Parse(csdlTxt).CreateReader());
```

baywet (approved):
thanks for taking the time to implement the changes!

PR #478, opened 2021-02-23
Quote fixes in JavascriptGenerateObjectFromJson

### LINKED ISSUE #476: JavaScript snippet generator doesn't handle @odata fields correctly.
Example: https://docs.microsoft.com/en-us/graph/api/user-post-manager?view=graph-rest-1.0&tabs=javascript

The code ends up creating an object like this:

```
const directoryObject = {
  @odata.id: "https://graph.microsoft.com/v1.0/users/{id}"
};
```

But you can't use `@odata.id` as a property name. You have to surround it with quotes (or double-quotes):

```
const directoryObject = {
  '@odata.id': "https://graph.microsoft.com/v1.0/users/{id}"
};
``` 
[AB#8269](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8269)

### LINKED ISSUE #477: JavaScript snippet generator is inconsistent with quotes for string literals
Sometimes it uses double-quotes, sometimes single. Snippets should use a consistent approach.

For example: https://docs.microsoft.com/en-us/graph/api/user-post-manager?view=graph-rest-1.0&tabs=javascript

The value of `@odata.id` is specified with double-quotes, but the URL in the `client.api` call is specified with single-quotes.

```
const options = {
	authProvider,
};

const client = Client.init(options);

const directoryObject = {
  @odata.id: "https://graph.microsoft.com/v1.0/users/{id}"
};

let res = await client.api('/users/{id}/manager/$ref')
	.put(directoryObject);
``` 
[AB#8273](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8273)

jasonjoh:
- Changed regex that stripped quotes from key names to skip keys with punctuation.
- Added additional regex to change those to single-quotes.
- Added additional regex to replace surrounding double-quotes on values to single quotes.

These changes also force consistent spacing around the colon in JSON objects.

Fixes #476 
Fixes #477

### DIFF
--- a/CodeSnippetsReflection.Test/JavascriptGeneratorShould.cs
+++ b/CodeSnippetsReflection.Test/JavascriptGeneratorShould.cs
@@ -58,29 +58,29 @@ namespace CodeSnippetsReflection.Test
 
             //Assert code snippet string matches expectation
             const string expectedSnippet = "const user = {\r\n" +
-                                           "  subject: \"Let's go for lunch\",\r\n" +
+                                           "  subject: 'Let's go for lunch',\r\n" +
                                            "  body: {\r\n" +
-                                           "    contentType: \"HTML\",\r\n" +
-                                           "    content: \"Does mid month work for you?\"\r\n" +
+                                           "    contentType: 'HTML',\r\n" +
+                                           "    content: 'Does mid month work for you?'\r\n" +
                                            "  },\r\n" +
                                            "  start: {\r\n" +
-                                           "      dateTime: \"2019-03-15T12:00:00\",\r\n" +
-                                           "      timeZone: \"Pacific Standard Time\"\r\n" +
+                                           "      dateTime: '2019-03-15T12:00:00',\r\n" +
+                                           "      timeZone: 'Pacific Standard Time'\r\n" +
                                            "  },\r\n" +
                                            "  end: {\r\n" +
-                                           "      dateTime: \"2019-03-15T14:00:00\",\r\n" +
-                                           "      timeZone: \"Pacific Standard Time\"\r\n" +
+                                           "      dateTime: '2019-03-15T14:00:00',\r\n" +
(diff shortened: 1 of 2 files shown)


zengin (changes requested):
Changing all double quotes to single quotes requires us to escape single quotes in those strings. Example:

```js
'user@odata.bind': 'https://graph.microsoft.com/beta/users('18a80140-b0fb-4489-b360-2f6efaf225a0')'
```
https://github.com/microsoftgraph/microsoft-graph-docs/commit/cb0cf45a4c575690b90e7987366a22fd74547d53#diff-92aa5760d179674bcdf5b371ae706786bd80cd8df83246c8f59e342450ca659d

And for reference, this commit shows the full diff on the docs repo:
https://github.com/microsoftgraph/microsoft-graph-docs/commit/cb0cf45a4c575690b90e7987366a22fd74547d53

jasonjoh:
Nice catch!

zengin (approved):
Here is the last commit's effect on docs:
https://github.com/microsoftgraph/microsoft-graph-docs/commit/49a593cd0069e12349c315772aaf80dbdf6fdd97

Only [this file](https://github.com/microsoftgraph/microsoft-graph-docs/commit/49a593cd0069e12349c315772aaf80dbdf6fdd97#diff-621fc54962ae9c0755deb73db7585c30ed304667b66672ef777b4b93a0ac5808) seems wrong, but it is broken in other ways too. We should be converting the string into a stream object anyways. So I don't think it is a blocker.

Thanks for making the changes.

LGTM.

jasonjoh:
Ah yeah - it looks like the JSON is invalid in that file - it's using single-quotes around keys and values, which isn't correct.

PR #488, opened 2021-03-03
Consistent ID placeholders in C# snippets

zengin:
Fixes: https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/202

Generates consistent id output for C# snippets so that we can replace them with real values while running the snippets in Raptor.

ID placeholder value is extracted from previous segment's type. 

Examples:
- `graphClient.Me.Drive.Items["{driveItem-id}"]`
- `graphClient.Me.PendingAccessReviewInstances["{accessReviewInstance-id}"]`
- `graphClient.Me.MailFolders["{mailFolder-id}"].Messages`
- `graphClient.TermStore.Sets["{termStore.set-id}"]`

Namespaces are shortened by dropping `microsoft.graph` from the name as it is a shared prefix for all the types. Subnamespaces like `termStore` are kept to cover the case of name collisions between two different namespaces.

Full diff on the docs repo:
https://github.com/microsoftgraph/microsoft-graph-docs/commit/3257bee01450b32df17c592dc30477f92c334761

### DIFF
--- a/CodeSnippetsReflection.Test/CSharpGeneratorShould.cs
+++ b/CodeSnippetsReflection.Test/CSharpGeneratorShould.cs
@@ -207,8 +207,8 @@ namespace CodeSnippetsReflection.Test
 
             // Assert that stream is set as a property of a full object and full object is passed in in the PATCH request.
             Assert.Contains("pages.Content = stream;", result);
-            Assert.Contains("await graphClient.Me.Onenote.Pages[\"{id}\"]", result);
-            Assert.DoesNotContain("await graphClient.Me.Onenote.Pages[\"{id}\"].Content", result);
+            Assert.Contains("await graphClient.Me.Onenote.Pages[\"{onenotePage-id}\"]", result);
+            Assert.DoesNotContain("await graphClient.Me.Onenote.Pages[\"{onenotePage-id}\"].Content", result);
             Assert.Contains("UpdateAsync(pages)", result);
         }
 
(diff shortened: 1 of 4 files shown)


MIchaelMainer (dismissed):

baywet (dismissed):

zengin:
Latest 3 commits limit the change to the command line callers only (i.e. snippet generator for the documentation).

We are not using placeholder IDs for DevX API callers, instead we continue to use ID values from original HTTP snippet. This is especially useful for Graph Explorer "Code Snippets" tab.

zengin:
I reran the full snippet generation for the last 3 commits, [no diff](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=42619&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=e19b279f-2a41-5171-6e52-a43b27738914&l=21619) was generated.

zengin:
+@Lauragra for awareness.

After merging this, we will no longer transfer ID placeholders from HTTP snippets to C# snippets because there is inconsistency in naming, e.g.:
- guid values
- Base64 values (often shortened)
- {id} {item-id} etc..

We will be using type information to generate these IDs so that when we execute the snippets in our tests, we will find what type of IDs we need to replace with real values. See examples of generated placeholders in the PR description and the upcoming full change [here](https://github.com/microsoftgraph/microsoft-graph-docs/commit/3257bee01450b32df17c592dc30477f92c334761).

Please let us know if you have any concerns.

CodeSnippetsReflection.Test/CSharpGeneratorShould.cs
```diff
@@ -1049,5 +1049,40 @@ public void GeneratesSnippetsInOtherNamespace()
...
+            Assert.Contains(itemIdBasedOnTypeInformation, result);
+            Assert.DoesNotContain(itemIdFromOriginalSnippet, result);
+        }
+
+        [Fact]
+        public void UsesIDFromHTTPSnippetForDevXAPICallers()
+        {
+            const bool isCommandLine = false; // DevX API caller (e.g. Graph Explorer "Code Snippets" tab)
+
+            const string itemIdFromOriginalSnippet = "{item-id-from-original-snippet}";
```
MIchaelMainer, line 1075: nit: itemIdFromDevXApiSnippet would be a better name to let us know that this supports that scenario. I also suggest that the id look like a real id since we expect that from DevX API

MIchaelMainer (approved):

MIchaelMainer (approved):

PR #503, opened 2021-03-11
Fix handling of Enums in Snippet generation

### LINKED ISSUE #482: [Snippet Generation] Handling of enums is not correct
Example:
https://docs.microsoft.com/en-us/graph/api/authenticationmethodsroot-usersregisteredbymethod?view=graph-rest-beta&tabs=http#request

HTTP snippet:
```http
GET https://graph.microsoft.com/beta/reports/authenticationMethods/usersRegisteredByMethod(includedUserTypes='all',includedUserRoles='all')
```

Actual C# snippet:
```csharp

var userRegistrationMethodSummary = await graphClient.Reports.AuthenticationMethods
	.UsersRegisteredByMethod(.all,.all)
	.Request()
	.GetAsync();
```

Expected C# snippet:
```csharp
var userRegistrationMethodSummary = await client.Reports.AuthenticationMethods
	.UsersRegisteredByMethod(IncludedUserTypes.All, IncludedUserRoles.All)
	.Request()
	.GetAsync();
```
 
[AB#8302](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8302)

fey101:
Fixes #482 

This PR Fixes handling of Enums in C# Snippet generation by prepending the `enum` type name to the `enum` value in the common language agnostic generator if the `returnEnumTypeIfEnum` optional named param is set, then uppercasing the enum value in the Csharp generator. Using the optional named param makes it possible for languages that make use of the enum type to specify the need of prepending the enumtype in the returned parameter list.

### DIFF
--- a/CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
+++ b/CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
@@ -326,7 +326,9 @@ namespace CodeSnippetsReflection.LanguageGenerators
         /// <param name="collectionSuffix">Suffix to be added to elements that are proved to be members collections</param>
         /// <param name="isOrderedByOptionalParameters">Flag to show whether the parameters are ordered by the the metadata or optionality of params</param>
         /// <returns></returns>
-        public static IEnumerable<string> GetParameterListFromOperationSegment(OperationSegment operationSegment, SnippetModel snippetModel, string collectionSuffix = "", bool isOrderedByOptionalParameters = true)
+        public static IEnumerable<string> GetParameterListFromOperationSegment(
+            OperationSegment operationSegment, SnippetModel snippetModel, string collectionSuffix = "",
+            bool isOrderedByOptionalParameters = true, bool returnEnumTypeIfEnum = false)
         {
             var paramList = new List<string>();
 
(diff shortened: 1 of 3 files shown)


zengin (changes requested):
Looks good for C#, but it creates  a side effect for Java since part of the fix is in the `CommonGenerator`:

https://github.com/microsoftgraph/microsoft-graph-docs/commit/73137c89fb6916060e4dde3e2d025740513142e6

We don't have to fix it for Java in this PR, but it would be nice if you can remove the side effect and keep the state as is for Java.

CodeSnippetsReflection/LanguageGenerators/CommonGenerator.cs
```diff
@@ -326,7 +326,9 @@ public static string LowerCaseFirstLetter(string s)
         /// <param name="collectionSuffix">Suffix to be added to elements that are proved to be members collections</param>
         /// <param name="isOrderedByOptionalParameters">Flag to show whether the parameters are ordered by the the metadata or optionality of params</param>
         /// <returns></returns>
-        public static IEnumerable<string> GetParameterListFromOperationSegment(OperationSegment operationSegment, SnippetModel snippetModel, string collectionSuffix = "", bool isOrderedByOptionalParameters = true)
+        public static IEnumerable<string> GetParameterListFromOperationSegment(
+            OperationSegment operationSegment, SnippetModel snippetModel, string collectionSuffix = "",
+            bool isOrderedByOptionalParameters = true, bool returnEnumTypeIfEnum = false)
```
zengin, line 331: nit: style
please choose either "all parameters in one line" or "one parameter per line".
```suggestion
            OperationSegment operationSegment,
            SnippetModel snippetModel,
            string collectionSuffix = "",
            bool isOrderedByOptionalParameters = true,
            bool returnEnumTypeIfEnum = false)
```

zengin (approved):

PR #517, opened 2021-04-01
Escapes # character in graph.networkInterface schema

### LINKED ISSUE #514: Remove unexpected character "*<#>" in OpenAPI description
Related to: https://github.com/microsoftgraph/microsoft-graph-docs/issues/12110

This is affecting PowerShell code gen. 
[AB#8803](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8803) 
[AB#8804](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8804)

irvinesunday:
Fixes #514 

**Issue:**
The presence of the # character in the `Description` property of the `Properties` property of the `microsoft.graph.networkInterface` schema is breaking Powershell Comment Based Help System, and by extension breaking modules.

e.g.
MicrosoftTeams-image (12)

produces...
MicrosoftTeams-image (13)

**Fix:**
By appending a ` before the # to escape it.

**Test:**
- A unit test has been added to validate this fix.
- The mock test class has been updated to include the test schema and test path.

NB: This is a temporary fix as a more permanent solution is awaited from AutoRest. AutoRest should be able to parse the document without breaking if a # character is included within string characters ' '

### DIFF
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -334,6 +334,10 @@ namespace OpenAPIService
 
                     // Remove the root path to make AutoREST happy
                     subsetOpenApiDocument.Paths.Remove("/");
+
+                    // Temp. fix - Escape the # character from description in
+                    // 'microsoft.graph.networkInterface' schema
+                    EscapePoundCharacter(subsetOpenApiDocument.Components);
                 }
             }
 
@@ -398,7 +402,7 @@ namespace OpenAPIService
             return document;
         }
 
-        private static OpenApiDocument FixReferences(OpenApiDocument document)
+        public static OpenApiDocument FixReferences(OpenApiDocument document)
         {
             // This method is only needed because the output of ConvertToOpenApi isn't quite a valid OpenApiDocument instance.
             // So we write it out, and read it back in again to fix it up.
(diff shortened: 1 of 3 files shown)


OpenAPIService/OpenApiService.cs
```diff
@@ -529,5 +533,26 @@ string evaluator(Match match)
...
+        /// </remarks>
+        /// <param name="components">The <see cref="OpenApiComponents"/> object with the target schema.</param>
+        private static void EscapePoundCharacter(OpenApiComponents components)
+        {
+            if (components.Schemas.TryGetValue("microsoft.graph.networkInterface", out OpenApiSchema parentSchema))
+            {
+                if (parentSchema.Properties.TryGetValue("description", out OpenApiSchema descriptionSchema))
+                {
+                    // PowerShell uses ` to escape special characters
+                    descriptionSchema.Description = descriptionSchema.Description.Replace("#", "`#");
```
baywet, line 553: could we imagine relying on a list of characters than need to be escaped instead? and rename this method to "escape Special characters"? this is probably the first of a list and we probably want to have to call a single method, not set multiple.

peombwa, line 553:
```suggestion
                    descriptionSchema.Description = descriptionSchema.Description.Replace("<#>", "<#/>");
```
We should close the tag (`<#/>`) and not escape `#`. The issue is that an unclosed tag confuses PowerShell's comment-based help to think that `#>` in the description is a closing tag of the comment block. A comment block in PowerShell opens with a `<#` and closes with a `#>`.

ref - https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comment_based_help?view=powershell-7.1#syntax-for-comment-based-help

irvinesunday: In essence, this is a hotfix (and a temporary fix), addressing the specific schema in the OpenAPI doc that currently has this breaking character. It is possible that in the future a similar breaking character may be introduced in other sections of the doc. and which may need to be sanitized as well. The challenge would be covering all these areas, as this would further add a performance and maintenance cost - checking all string values in the OpenAPI doc for special characters. Ideally, AutoRest should be able to parse a string text which has comment characters without breaking, and this where a more permanent solution needs to be explored. @peombwa might you know whether newer versions of AutoRest already handle this? I understand the PowerShell code gen. is using an older version of AutoRest?

baywet: thanks for sharing the context! Do you anticipate pushing the fix in the OpenApi lib?

peombwa: @irvinesunday, the issue is still present with newer versions of AutoREST.

irvinesunday: @baywet since this is only being experienced by the PowerShell client - which consumes the OpenAPI doc. through this library - the fix needs only to be localized to the DevX API and not the OpenAPI lib. 
@peombwa thanks for the confirmation. Can we create an issue on their public repo for them to fix the bug?

peombwa (approved):

baywet (approved):

PR #525, opened 2021-04-07
ReEnable Application Insights for Code-based monitoring

### LINKED ISSUE #138: Add logging to capture events.
[AB#7008](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7008)

### LINKED ISSUE #425: Re-enable app insights for graph explorer api
[AB#7790](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7790)

### LINKED ISSUE #516: Clean out PII data from Custom Event Telemetry in AppInsights
[AB#8813](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8813)

MaggieKimani1:
This PR involves reenabling App Insights for code-based monitoring using the ASP.Net Core SDK.
It also captures PII data handling by cleaning out any from of GUIDS or usernames and email addresses from the request body.
Fixes #138,
Fixes #425,
Fixes #516


GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+        // Matches patterns like users('MeganB@M365x214355.onmicrosoft.com')
+        private static Regex _emailRegex = new Regex(@"([a-zA-Z0-9_\.-]+)@([\da-zA-Z\.-]+)\.([a-zA-Z\.]{2,6})",
+            RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+        // Matches usernames with firstname or lastname labels
+        private static Regex _username = new Regex("(fn|ln|lastname|firstname|name|fullname)");
+
+        public TelemetryProcessor(ITelemetryProcessor next)
+        {
+            _next = next;
```
baywet, line 32: defensive programming?

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <summary>
+        /// Filters request data and calls the next ITelemetryProcessor in the chain
+        /// </summary>
+        /// <param name="item"></param>
+        public void Process(ITelemetry item)
+        {
+            try
+            {
+                var request = item as RequestTelemetry;
+                if (request != null)
```
baywet, line 44:
```suggestion
                if (item is RequestTelemetry request)
```
(and remove the previous line)

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+                {
+                    // Regex-replace anything that looks like a GUID or email
+                    request.Name = _guidRegex.Replace(request.Name, "{customerID}");
```
baywet, line 47: could Name be null?

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+
+                    if (request.Url != null)
+                    {
+                        if (_guidRegex.IsMatch(request.Url?.AbsoluteUri))
```
baywet, line 51: no need for null propagation operator as you're testing for not null in parent block

baywet (changes requested):

GraphWebApi/appsettings.json
```diff
@@ -12,7 +12,8 @@
   },
   "ApplicationInsights": {
     "InstrumentationKey": "ENTER_KEY_HERE",
-    "AppInsightsApiKey": "ENTER_KEY_HERE"
+    "AppInsightsApiKey": "ENTER_KEY_HERE",
```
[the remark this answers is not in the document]
MaggieKimani1: They're not similar but I can do away with the API key

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+                        }
+                        // Now it's time to stop the operation (and track telemetry).
+                        client.StopOperation(operation);
+                    }
+                }
+            }            
+        }       
+
+        //GetOperationId method
+        public static string GetOperationId(string id)
```
nikithauc, line 91: Please add summary for this function and describe what id stands for.

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+                _next.Process(item);
+            }
+        }
+
+        /// <summary>
+        /// Takes in a requestUrl param and returns a sanitized url.
+        /// </summary>
+        /// <param name="requestUrl"> The request url.</param>
+        /// <returns> A sanitized url.</returns>
+        public Uri SanitizeUrl(Uri requestUrl)
```
nikithauc, line 66: Can you add unit tests for the changes in this PR?

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+        public void Process(ITelemetry item)
+        {
+            try
+            {
+                if (item is RequestTelemetry request)
+                {
+                    if(request.Name != null)
+                    {
+                        // Regex-replace anything that looks like a GUID or email
+                        request.Name = _guidRegex.Replace(request.Name, "{customerID}");
```
nikithauc, line 49: Do we need a _emailRegex.Replace here as the comment mentions - Regex-replace anything that looks like a GUID or email?

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
```diff
@@ -8,6 +8,7 @@
 using GraphExplorerPermissionsService.Interfaces;
 using GraphExplorerPermissionsService.Models;
 using GraphWebApi.Common;
+using GraphWebApi.Telemetry;
```
nikithauc, line 11: Is this import being used anywhere? As the file has no other changes?

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        private static Regex _guidRegex = new Regex(@"\b[A-F0-9]{8}(?:-[A-F0-9]{4}){3}-[A-F0-9]{12}\b",
+            RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+        // Matches patterns like users('MeganB@M365x214355.onmicrosoft.com')
+        private static Regex _emailRegex = new Regex(@"([a-zA-Z0-9_\.-]+)@([\da-zA-Z\.-]+)\.([a-zA-Z\.]{2,6})",
+            RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+        // Matches usernames with firstname or lastname labels
+        private static Regex _username = new Regex("(fn|ln|lastname|firstname|name|fullname)");
```
nikithauc, line 28: How does this work?

GraphWebApi/Startup.cs
```diff
@@ -87,15 +88,16 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddApplicationInsightsTelemetry(options =>
             {
-               // options.InstrumentationKey = Configuration["ApplicationInsights:InstrumentationKey"];
-                options.RequestCollectionOptions.InjectResponseHeaders = false;
+                options.InstrumentationKey = Configuration["ApplicationInsights:InstrumentationKey"];
+                options.RequestCollectionOptions.InjectResponseHeaders = true;
```
MIchaelMainer, line 92: I don't think this is applicable here but response headers can contain PII. Just an FYI, please resolve.

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        public async Task Invoke(HttpContext context, TelemetryClient client)
+        {
+            if(context != null)
+            {
+                // Let's create and start RequestTelemetry.
+                var requestTelemetry = new RequestTelemetry
+                {
+                    // "Request {method} {url} => {statusCode}" format
+                    Name = $"{context.Request.Method + " " + context.Request.Path.Value}"
```
MIchaelMainer, line 38: Comment doesn't match the format.

`context.Request.Path.Value` looks like a URL. Where is this URL coming from and could it contain PII? If so, it must be sanitized.

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+                };
+                requestTelemetry.Properties.Add("HttpMethod", context.Request.Method);
+
+                var requestHeaders = context.Request.Headers;
+
+                // If there is a Request-Id received from the upstream service, set the telemetry context accordingly.
+                if (requestHeaders.ContainsKey("Request-Id"))
```
MIchaelMainer, line 45: I don't think the header case is always `Request-Id`. In fact, Request-Id is returned in responses, not requests.  Please make this check `StringComparison.OrdinalIgnoreCase`. I'm not quite sure I understand the context of these changes in this PR so it may make sense to check for `client-request-id`.

GraphWebApi/Telemetry/TelemetryProcessor.cs
MIchaelMainer: Also, does this filter properties like `$filter(displayname -eq 'Michael Mainer')` in a URL?

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+namespace GraphWebApi.Telemetry
+{
+    /// <summary>
+    /// Initializes a telemetry processor to sanitize http request urls to remove sensitive data.
+    /// </summary>
+    public class TelemetryProcessor : ITelemetryProcessor
+    {
+        private readonly ITelemetryProcessor _next;
+
+        private static Regex _guidRegex = new Regex(@"\b[A-F0-9]{8}(?:-[A-F0-9]{4}){3}-[A-F0-9]{12}\b",
```
MIchaelMainer, line 20: Are these the same filters used in GE? @millicentachieng

MIchaelMainer (commented):
Will we be able to test the telemetry capture against a pre-release instance of this application? I want to see what we capture before we take this to production.

Also, what is the context of this PR? Where is the data coming from?

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+            RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+        // Matches patterns like users('MeganB@M365x214355.onmicrosoft.com')
+        private static Regex _emailRegex = new Regex(@"([a-zA-Z0-9_\.-]+)@([\da-zA-Z\.-]+)\.([a-zA-Z\.]{2,6})",
```
irvinesunday, line 24: With [C#9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#fit-and-finish-features) you can omit the type in your constructor invocations: e.g.:

```suggestion
        private static readonly Regex _emailRegex = new(@"([a-zA-Z0-9_\.-]+)@([\da-zA-Z\.-]+)\.([a-zA-Z\.]{2,6})",
```
 Also, update the private fields to be `readonly`

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+
+                    // Get the operation ID from the Request-Id (if you follow the HTTP Protocol for Correlation).
+                    requestTelemetry.Context.Operation.Id = GetOperationId(requestId);
+                    requestTelemetry.Context.Operation.ParentId = requestId;
+                }
+
+                // StartOperation is a helper method that allows correlation of 
+                // current operations with nested operations/telemetry
+                // and initializes start time and duration on telemetry items.
+                using (var operation = client.StartOperation(requestTelemetry))
```
irvinesunday, line 57: You may simplify the using statement:

```suggestion
                using var operation = client.StartOperation(requestTelemetry);
```
Then remove the redundant open and close braces that follow.

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+        */
+        private readonly RequestDelegate _next;
+
+        public TelemetryMiddleware(RequestDelegate next)
+        {
+            _next = next
+                ?? throw new ArgumentNullException(nameof(next), $"{ next }: { nameof(next) }");
+        }
+
+        public async Task Invoke(HttpContext context, TelemetryClient client)
```
irvinesunday, line 30: Please add a summary for this public method.

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
+﻿// ------------------------------------------------------------------------------------------------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+// ------------------------------------------------------------------------------------------------------------------------------------------------------
+
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.AspNetCore.Http;
+using System;
+using System.Collections.Generic;
```
irvinesunday, line 9: Unused import; you may remove.

GraphWebApi/Properties/ServiceDependencies/local/appInsights1.arm.json
```diff
@@ -0,0 +1,67 @@
...
+              "kind": "web",
+              "properties": {},
+              "apiVersion": "2015-05-01"
+            }
+          ]
+        }
+      }
+    }
+  ],
+  "metadata": {
```
irvinesunday, line 64: Should this property be here? 
See: https://docs.microsoft.com/en-us/azure/azure-monitor/app/resource-manager-app-resource

GraphWebApi/Properties/ServiceDependencies/local/appInsights1.arm.json
```diff
@@ -0,0 +1,67 @@
...
+        "template": {
+          "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+          "contentVersion": "1.0.0.0",
+          "resources": [
+            {
+              "name": "DevX-API-AppInsights-Testing",
+              "type": "microsoft.insights/components",
+              "location": "[parameters('resourceLocation')]",
+              "kind": "web",
+              "properties": {},
```
irvinesunday, line 56: Do we need to include the required missing property `Application_Type` under `properties`?
image

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
```diff
@@ -14,6 +14,7 @@
 using System.Linq;
 using GraphWebApi.Common;
 using GraphExplorerSamplesService.Interfaces;
+using Microsoft.ApplicationInsights;
```
irvinesunday, line 17: Unused import; you may remove.

GraphWebApi/Telemetry/TelemetryMiddleware.cs
baywet, line 57: you're also missing defensive programming for the client

baywet (changes requested):

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+                    }
+                }
+            }            
+        }       
+
+        //GetOperationId method
+        public static string GetOperationId(string id)
+        {
+            // Returns the root ID from the '|' to the first '.' if any.
+            int rootEnd = (int)(id?.IndexOf('.'));
```
baywet, line 94:
```suggestion
            int rootEnd = id?.IndexOf('.') ?? -1;
```

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+                        client.StopOperation(operation);
+                    }
+                }
+            }            
+        }       
+
+        //GetOperationId method
+        public static string GetOperationId(string id)
+        {
+            // Returns the root ID from the '|' to the first '.' if any.
```
baywet, line 93:
```suggestion
            if(string.IsNullOrEmpty(id)) return string.Empty;
            // Returns the root ID from the '|' to the first '.' if any.
```

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// Filters request data and calls the next ITelemetryProcessor in the chain
+        /// </summary>
+        /// <param name="item"> A telemetry Item.</param>
+        public void Process(ITelemetry item)
+        {
+            try
+            {
+                if (item is RequestTelemetry request)
+                {
+                    if(request.Name != null)
```
baywet, line 46:
```suggestion
                    if(!string.IsNullOrEmpty(request.Name))
```

GraphWebApi/Telemetry/TelemetryProcessor.cs
MaggieKimani1: That's a typo on my end. The regex only caters for GUIDs

GraphWebApi/Controllers/GraphExplorerPermissionsController.cs
MaggieKimani1: No it's not. Cleaning it up..

ddyett:
We should put this in a feature branch first so we can do validation we have no privacy or security issues.  I wouldn't want this to block deploying other work.

GraphWebApi/Startup.cs
```diff
@@ -87,15 +88,16 @@ public void ConfigureServices(IServiceCollection services)
...
                 options.RequestCollectionOptions.TrackExceptions = false;
                 options.EnableAuthenticationTrackingJavaScript = false;
-                options.EnableHeartbeat = false;
-                options.EnableAdaptiveSampling = false;
-                options.EnableQuickPulseMetricStream = false;
-                options.EnableDebugLogger = false;
+                options.EnableHeartbeat = true;
+                options.EnableAdaptiveSampling = true;    // Enable adaptive sampling
```
ddyett, line 96: the name is self explanatory.  it would be great if the comment better described what enabling adaptive sampling means.

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+                        client.TrackException(e);
+                        throw;
+                    }
+                    finally
+                    {
+                        // Update status code and success as appropriate.
+                        if (context.Response != null)
+                        {
+                            requestTelemetry.ResponseCode = context.Response.StatusCode.ToString();
+                            requestTelemetry.Success = context.Response.StatusCode >= 200 && context.Response.StatusCode <= 299;
```
ddyett, line 77:
```suggestion
                            requestTelemetry.Success = context.Response.StatusCode is >= 200 and <= 299;
```

GraphWebApi/Telemetry/TelemetryMiddleware.cs
ddyett, line 94: stay consistent with var usage.

GraphWebApi/Telemetry/TelemetryProcessor.cs
ddyett, line 20: public static readonly

GraphWebApi/Telemetry/TelemetryProcessor.cs
ddyett, line 66: looks like this can be private and static since it doesn't use any instance properties.

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        /// <param name="requestUrl"> The request url.</param>
+        /// <returns> A sanitized url.</returns>
+        public Uri SanitizeUrl(Uri requestUrl)
+        {
+            if (requestUrl != null)
+            {
+                if (_guidRegex.IsMatch(requestUrl.AbsoluteUri))
+                {
+                    requestUrl = new Uri(_guidRegex.Replace(requestUrl.ToString(), "{customerID}"));
```
ddyett, line 72: not true that any guid is a customerid.  Perhaps just Id

GraphWebApi/Telemetry/TelemetryMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace GraphWebApi.Telemetry
+{
+    /// <summary>
+    /// This class implements a telemetry middleware for correlating telemetry operations
+    /// </summary>
+    public class TelemetryMiddleware
```
ddyett, line 17: unit test this middleware.

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,87 @@
...
+        private readonly ITelemetryProcessor _next;
+
+        private static Regex _guidRegex = new Regex(@"\b[A-F0-9]{8}(?:-[A-F0-9]{4}){3}-[A-F0-9]{12}\b",
+            RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+        // Matches patterns like users('MeganB@M365x214355.onmicrosoft.com')
+        private static Regex _emailRegex = new Regex(@"([a-zA-Z0-9_\.-]+)@([\da-zA-Z\.-]+)\.([a-zA-Z\.]{2,6})",
+            RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+        // Matches usernames with firstname or lastname labels
```
thewahome, line 27: I have noticed a huge amount of comments on the code even in places where the code is self-explanatory. When there are many comments and a refactor happens, the comments can quickly get outdated and not match the methods they intend to explain.

When a statement/a set of statements require comments for them to make sense, it/they are ripe for refactoring into a method (or a better named variable in the case of a statement )that communicates intent. So that someone can easily read the function name and understand it. 

Summaries on public functions are helpful for the public clients that depend on them.

thewahome (changes requested):

GraphWebApi/Telemetry/TelemetryProcessor.cs
MaggieKimani1: I've implemented logic to filter that out

GraphWebApi/Telemetry/TelemetryMiddleware.cs
fey101: Is StatusCode **is** >= 200 **and** <= 299 valid C# code? Interesting coz it looks like a mix of syntax from different languages?

GraphWebApi/Telemetry/TelemetryProcessor.cs
```diff
@@ -0,0 +1,122 @@
...
+namespace GraphWebApi.Telemetry
+{
+    /// <summary>
+    /// Initializes a telemetry processor to sanitize http request urls to remove sensitive data.
+    /// </summary>
+    public class TelemetryProcessor : ITelemetryProcessor
+    {
+        private readonly ITelemetryProcessor _next;
+
+        private static readonly Regex _guidRegex = new(@"\b[A-F0-9]{8}(?:-[A-F0-9]{4}){3}-[A-F0-9]{12}\b",
```
fey101, line 20: This appears quite strange to me. Is there no package in C# that can handle this better than having one hard code the actual regex? Or why was this necessary?

MaggieKimani1: I haven't found a package that caters to my particular use case, hence the hard coding.

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -0,0 +1,143 @@
...
+                {"guidRegex", _guidRegex},
+                {"emailRegex", _emailRegex},
+            };
+
+            var requestPath = customEvent.Properties["RequestPath"];
+            var renderedMessage = customEvent.Properties["RenderedMessage"];
+
+            foreach (var value in piiRegex.Values)
+            {
+                if (value.IsMatch(requestPath) && value.IsMatch(renderedMessage))
```
baywet, line 74: shouldn't it be || instead?

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -0,0 +1,143 @@
...
+            var piiRegex = new Dictionary<string, Regex>
+            {
+                {"guidRegex", _guidRegex},
+                {"emailRegex", _emailRegex},
+            };
+
+            var requestPath = customEvent.Properties["RequestPath"];
+            var renderedMessage = customEvent.Properties["RenderedMessage"];
+
+            foreach (var value in piiRegex.Values)
```
baywet, line 72:
```suggestion
            foreach (var piiRegex in piiRegexes.Values)
```

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -0,0 +1,143 @@
...
+            _next.Process(item);
+        }
+
+        /// <summary>
+        /// Sanitizes a custom event telemetry's request path and the rendered message properties
+        /// </summary>
+        /// <param name="request"> An event telemetry item.</param>
+        public void SanitizeEventTelemetry(EventTelemetry customEvent)
+        {
+            var piiRegex = new Dictionary<string, Regex>
```
baywet, line 63:
```suggestion
            var piiRegexes = new Dictionary<string, Regex>
```

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <param name="requestUrl"> The url of the Http request.</param>
+        /// <returns>A sanitized request url.</returns>
+        private void SanitizeQueryString(EventTelemetry customEvent)
+        {
+            if (customEvent != null)
+            {
+                var requestPath = customEvent.Properties["RequestPath"];
+                var renderedMessage = customEvent.Properties["RenderedMessage"];
+
+                if (requestPath.Contains("filter") && requestPath.Contains("displayname"))
```
baywet, line 96: what about firstName, lastName, email?... maybe this should rely on a list instead and iterate over it

baywet (changes requested):

GraphWebApi/Telemetry/CustomPIIFilter.cs
MaggieKimani1: Oh yes, let me make that addition

GraphWebApi/Telemetry/CustomPIIFilter.cs
baywet: but won't this condition sanitize ONLY if BOTH contain PII information?

MaggieKimani1: I can change it to || , but my understanding is that if the request path has PII, then definitely the rendered message will also contain PII since it contains the request path string.

baywet: let me rephrase: is there a possibility for either requestPath or renderedMessage to contain PII but not both?

thewahome (approved):

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -0,0 +1,151 @@
...
+        /// </summary>
+        /// <param name="requestUrl"> The url of the Http request.</param>
+        /// <returns>A sanitized request url.</returns>
+        private void SanitizeQueryString(EventTelemetry customEvent)
+        {
+            if (customEvent != null)
+            {
+                var propertyNames = new List<string>
+                {
+                    "displayname",
```
baywet, line 95: id, givenName, surname, employeeId, phone numbers, all kind of upns and email addresses, and more probably need to be in there as well. Please cross check the [user properties](https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0#properties) and see which ones apply.

We might actually take an allow list approach where properties we allow and want to track are kept, and the rest is discarded. This will protect us from having a workload adding a new PII property, app developers filtering on it and us logging it when it wasn't intended to.

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        /// <summary>
+        /// Sanitizes a custom event telemetry's request path and the rendered message properties
+        /// </summary>
+        /// <param name="request"> An event telemetry item.</param>
+        public void SanitizeEventTelemetry(EventTelemetry customEvent)
+        {
+            var piiRegexes = new Dictionary<string, Regex>
+            {
+                {"guidRegex", _guidRegex},
```
baywet, line 65: you're never reading the key, switch to a list instead?

baywet (changes requested):

GraphWebApi/Telemetry/CustomPIIFilter.cs
MaggieKimani1: I'm already masking the Id, email addresses and UPNs.
I've added a check for more user properties that you've suggested after counter-checking with the link you shared.

GraphWebApi/Properties/ServiceDependencies/local/appInsights1.arm.json
```diff
@@ -53,14 +53,15 @@
               "type": "microsoft.insights/components",
               "location": "[parameters('resourceLocation')]",
               "kind": "web",
-              "properties": {
-                "Application_Type": "web"
-              },
+              "properties": {}
```
baywet, line 56:
```suggestion
              "properties": {},
```

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -129,21 +176,20 @@ private void SanitizeQueryString(EventTelemetry customEvent)
...
+
             if (_usernameRegex.IsMatch(propertyName))
             {
-                propertyName = propertyName.Replace(propertyName, " '****'");
+                propertyName = propertyName.Replace(propertyName, "'****'");
             }
 
             // Append sanitized property name to query string
-            var newQueryString = $"{queryString[0]}eq{propertyName})";
+            var newQueryString = $"{queryString[0] + propertyName + queryString[2]}";
```
baywet, line 192: comments for the magic numbers

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -113,14 +150,24 @@ private void SanitizeQueryString(EventTelemetry customEvent)
                         var text = renderedMessage.Split(separators, StringSplitOptions.RemoveEmptyEntries);
                         requestPath = text[1];
 
-                        var newQueryString = FormatString(requestPath);
+                        var newQueryString = RedactUserName(requestPath);
 
                         // Append sanitized property name to query string
-                        newQueryString = $"{text[0]}GET{newQueryString} responded{text[2]}";
+                        newQueryString = $"{text[0]}GET{newQueryString}responded{text[2]}";
```
baywet, line 156: comments for the magic numbers

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -58,52 +81,66 @@ public void Process(ITelemetry item)
...
         /// <summary>
         /// Parses the query string in a request url and sanitizes the username property.
         /// </summary>
         /// <param name="requestUrl"> The url of the Http request.</param>
         /// <returns>A sanitized request url.</returns>
-        private void SanitizeQueryString(EventTelemetry customEvent)
+        private void SanitizeQueryString(EventTelemetry customEvent = null, RequestTelemetry request = null)
         {
-            if (customEvent != null)
+            var propertyNames = new List<string>
```
baywet, line 124: move this to a static constant outside of the method to avoid initializing the list on every method call

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -58,52 +81,66 @@ public void Process(ITelemetry item)
...
-            foreach (var piiRegex in piiRegexes.Values)
-            {
-                if (piiRegex.IsMatch(requestPath) && piiRegex.IsMatch(renderedMessage))
+                foreach (var piiRegex in piiRegexes)
                 {
-                    customEvent.Properties["RequestPath"] = piiRegex.Replace(customEvent.Properties["RequestPath"], "****");
-                    customEvent.Properties["RenderedMessage"] = piiRegex.Replace(customEvent.Properties["RenderedMessage"], "****");
+                    if (piiRegex.IsMatch(requestPath) && piiRegex.IsMatch(renderedMessage))
+                    {
+                        customEvent.Properties["RequestPath"] = piiRegex.Replace(customEvent.Properties["RequestPath"], "****");
```
baywet, line 95: move the keys to constants to avoid duplication and facilitate maintenance

baywet (changes requested):

GraphWebApi/Telemetry/CustomPIIFilter.cs
```diff
@@ -0,0 +1,204 @@
...
+                                           string renderedMessage = null)
+        {
+            if(customEvent != null)
+            {
+                foreach (var piiRegex in piiRegexes)
+                {
+                    if (piiRegex.IsMatch(requestPath) && piiRegex.IsMatch(renderedMessage))
+                    {
+                        customEvent.Properties["RequestPath"] = piiRegex.Replace(requestPath, "****");
+                        customEvent.Properties["RenderedMessage"] = piiRegex.Replace(renderedMessage, "****");
```
baywet, line 109: Maybe I wasn't clear enough: I didn't mean move things as a parameter. The code still has "requestPath" and "renderedMessage" defined 6 times only in that file.
1. Move those keys into `private const string`
2. Use those constants instead of redefining by passing them to the indexers

baywet (approved):

PR #528, opened 2021-04-08
Support Casting of CollectionWithReferencesPage

### LINKED ISSUE #327: Snippet Generation needs casting support for *CollectionWithReferencesPage
In this sample:
https://docs.microsoft.com/en-us/graph/api/identitycontainer-post-b2cuserflows?view=graph-rest-beta&tabs=csharp#request-1

Cast to `IB2cIdentityUserFlowIdentityProvidersCollectionPage` should be replaced with `IB2cIdentityUserFlowIdentityProvidersCollectionWithReferencesPage`

In this sample:
https://docs.microsoft.com/en-us/graph/api/identitycontainer-post-b2xuserflows?view=graph-rest-beta&tabs=csharp#request-1
Cast to `IB2xIdentityUserFlowIdentityProvidersCollectionPage` should be replaced with `IB2xIdentityUserFlowIdentityProvidersCollectionWithReferencesPage`
 
[AB#6991](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6991)

fey101:
Fixes #327
 
Initially, casting to CollectionWithReferencesPage was not supported. Navigation properties with attribute `ContainsTarget=false` e.g `identityProviders` would be converted to type "ICollectionPage" instead of type `CollectionWithReferencesPage`.

This PR fixes this by checking the ContainsTarget attribute and deciding on the type to return base on the value of the attribute.

Changes have been tested and verified to affect only b2xuserflows and b2cuserflows as expected: [here](https://github.com/microsoftgraph/microsoft-graph-docs/compare/snippet-generation/47479...snippet-generation/47480?expand=1)

### DIFF
--- a/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
+++ b/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
@@ -775,7 +775,7 @@ namespace CodeSnippetsReflection.LanguageGenerators
         /// <returns>ICollectionPage Interface name</returns>
         private string GetCollectionPageClassName(ODataPathSegment pathSegment, ICollection<string> path)
         {
-            var type = GetCsharpClassName(pathSegment, path.ToList().GetRange(0, path.Count - 1));
+            var type = GetCsharpClassName(pathSegment, path.Take(path.Count-1).ToList());
             var property = path.Last();
             var uppercasedProperty = CommonGenerator.UppercaseFirstLetter(property);
 
--- a/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
+++ b/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
@@ -768,7 +768,7 @@ namespace CodeSnippetsReflection.LanguageGenerators
         }
 
         /// <summary>
-        /// Generates CollectionPage class name
+        /// Generates CollectionPage and CollectionWithReferencesPage class name
         /// </summary>
         /// <param name="pathSegment">The OdataPathSegment in use</param>
         /// <param name="path">path in edm model</param>
@@ -776,8 +776,16 @@ namespace CodeSnippetsReflection.LanguageGenerators
         private string GetCollectionPageClassName(ODataPathSegment pathSegment, ICollection<string> path)
         {
             var type = GetCsharpClassName(pathSegment, path.ToList().GetRange(0, path.Count - 1));
-            var property = CommonGenerator.UppercaseFirstLetter(path.Last());
-            return $"{type}{property}CollectionPage";
+            var property = path.Last();
+            var uppercasedProperty = CommonGenerator.UppercaseFirstLetter(property);
+
+            var structuredType = (pathSegment.EdmType as IEdmCollectionType)?.ElementType?.Definition as IEdmStructuredType;
+            var containsTarget = (structuredType?.FindProperty(property) as IEdmNavigationProperty)?.ContainsTarget
+                ?? throw new InvalidOperationException("Path doesn't match with the metadata property description!" +
+                    $"path: { string.Join(' ', path) }");
+
+            var navCollectionType = containsTarget ? "CollectionPage" : "CollectionWithReferencesPage";
+            return $"{type}{uppercasedProperty}{navCollectionType}";
         }
 
         /// <summary>
--- a/CodeSnippetsReflection.Test/CSharpGeneratorShould.cs
+++ b/CodeSnippetsReflection.Test/CSharpGeneratorShould.cs
@@ -1193,5 +1193,69 @@ namespace CodeSnippetsReflection.Test
             Assert.Contains("AdditionalData", result);
             Assert.Contains("\"languageId\", \"languageId-value\"", result);    // LanguageId as part of AdditionalData
         } 
+
+        [Fact]
+        public void ShouldSupportCastingToCollectionWithReferencesPage()
+        {
+            var expressions = new CSharpExpressions();
+            const string jsonObject = @"{
+                ""id"": ""Customer"",
+                ""userFlowType"": ""signUpOrSignIn"",
+                ""userFlowTypeVersion"": 3,
+                ""identityProviders"": [
+                    {
+                        ""id"": ""Facebook-OAuth"",
+                        ""type"": ""Facebook"",
+                        ""Name"": ""Facebook""
+                    }
+                ]
+            }";
+            var requestPayload = new HttpRequestMessage(HttpMethod.Post, "https://graph.microsoft.com/beta/identity/b2cUserFlows")
+            {
+                Content = new StringContent(jsonObject)
+            };
+            var snippetModel = new SnippetModel(requestPayload, ServiceRootUrlBeta, _edmModelBeta.Value);
+
+            var result = new CSharpGenerator(_edmModelBeta.Value).GenerateCodeSnippet(snippetModel, expressions);
+
+            var replacedType = "IdentityProviders = new B2cIdentityUserFlowIdentityProvidersCollectionPage()";
+            var expectedReplacement = "IdentityProviders = new B2cIdentityUserFlowIdentityProvidersCollectionWithReferencesPage";
+            Assert.DoesNotContain(replacedType, result);
+            Assert.Contains(expectedReplacement, result);
+        }
+
+        [Fact]
+        public void ShouldSupportCastingToCollectionPage()
+        {
+            var expressions = new CSharpExpressions();
+            const string jsonObject = @"{
+                ""displayName"": ""Books"",
+                ""columns"": [
+                    {
+                    ""name"": ""Author"",
+                    ""text"": { }
+                    },
+                    {
+                    ""name"": ""PageCount"",
+                    ""number"": { }
+                    }
+                ],
+                ""list"": {
+                    ""template"": ""genericList""
+                }
+            }";
+            var requestPayload = new HttpRequestMessage(HttpMethod.Post, "https://graph.microsoft.com/beta/sites/{site-id}/lists")
+            {
+                Content = new StringContent(jsonObject)
+            };
+            var snippetModel = new SnippetModel(requestPayload, ServiceRootUrlBeta, _edmModelBeta.Value);
+
+            var result = new CSharpGenerator(_edmModelBeta.Value).GenerateCodeSnippet(snippetModel, expressions);
+
+            var wrongType = "Columns = new ListColumnsCollectionWithReferencesPage()";
+            var expectedType = "Columns = new ListColumnsCollectionPage()";
+            Assert.DoesNotContain(wrongType, result);
+            Assert.Contains(expectedType, result);
+        }
     }
 }


CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -777,7 +777,7 @@ private string GetCollectionPageClassName(ODataPathSegment pathSegment, ICollect
         {
             var type = GetCsharpClassName(pathSegment, path.ToList().GetRange(0, path.Count - 1));
             var property = CommonGenerator.UppercaseFirstLetter(path.Last());
-            return $"{type}{property}CollectionPage";
+            return $"{type}{property}CollectionWithReferencesPage";
```
baywet, line 780: would that change affect all collections? even the ones that don't support references?

fey101: Probably, yes. I'll look into this. Do you know of any collection that doesn't support references, that I could use to test this?

baywet: the lists collection on a SharePoint site is a good example.

MIchaelMainer:
At first glance, we may need to add a GetCollectionWithReferencesPageClassName(). This would be called here:
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/851eea0381cf448d0671421fcd200079af23e4a2/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs#L558

When `typeProperties.IsNavigationProperty` and when the target is not contained.

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
fey101: It appears b2xUserflows, b2cUserflows as well as sharepoint columns list all contain the attribute ContainsTarget=True. https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_beta_metadata/cleanMetadataWithDescriptionsbeta.xml. Not sure what the differentiating factor will therefore be.

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -557,7 +557,7 @@ string dataRow(KeyValuePair<string, string> kvp)
 
                             if (typeProperties.IsNavigationProperty)
                             {
-                                stringBuilder.Append($"new {GetCollectionPageClassName(pathSegment, path)}()\r\n{tabSpace}{{\r\n");
+                                stringBuilder.Append($"new {GetCollectionPageClassName((NavigationPropertySegment)pathSegment, path)}()\r\n{tabSpace}{{\r\n");
```
baywet, line 560: use as instead

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -773,11 +773,13 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
         /// <param name="pathSegment">The OdataPathSegment in use</param>
         /// <param name="path">path in edm model</param>
         /// <returns>ICollectionPage Interface name</returns>
-        private string GetCollectionPageClassName(ODataPathSegment pathSegment, ICollection<string> path)
+        private string GetCollectionPageClassName(NavigationPropertySegment pathSegment, ICollection<string> path)
```
baywet, line 776: defensive programming

baywet (changes requested):

fey101:
> At first glance, we may need to add a GetCollectionWithReferencesPageClassName(). This would be called here:
> https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/851eea0381cf448d0671421fcd200079af23e4a2/CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs#L558
> 
> When `typeProperties.IsNavigationProperty` and when the target is not contained.

The `GetCollectionPageClassName` is only referenced once. Creating another class `GetCollectionWithReferencesPageClassName` would therefore render the former class stale. So instead, I chose to check within the class if `ContainsTarget` is true to differentiate between the 2. However, current problem is that both `b2cUserFlows`/`b2xUserFlows` and Sharepoint `columns` list appear to have the attribute `ContainsTarget=True`

zengin (approved):

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -767,19 +767,50 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
...
         }
 
+        /// <summary>
+        /// Return ElementType.Definition as an IEdmStructuredType
+        /// </summary>
+        /// <param name="navPathSegment">The NavigationPropertySegment in use</param>
+        /// <returns>ElementType.Definition as an IEdmStructuredType</returns>
+        private IEdmStructuredType GetElementTypeDefinitionProperties(NavigationPropertySegment navPathSegment)
+        {
+            if (navPathSegment.EdmType is IEdmCollectionType)
```
baywet:
```suggestion
            if (navPathSegment.EdmType is IEdmCollectionType edmCollectionType)
```

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -767,19 +767,50 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
...
+            {
+                var definition =  (navPathSegment.EdmType as IEdmCollectionType).ElementType.Definition;
```
baywet:
```suggestion
                var definition =  edmCollectionType.ElementType.Definition;
```

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -767,19 +767,50 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
...
+                return definition as IEdmStructuredType;
+            }
+            else
+            {
+                throw new Exception($"Failed to get ElementType.Definition properties for {navPathSegment}");
```
baywet:
```suggestion
                throw new InvalidOperationException($"Failed to get ElementType.Definition properties for {navPathSegment}");
```

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -767,19 +767,50 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
...
-            var type = GetCsharpClassName(pathSegment, path.ToList().GetRange(0, path.Count - 1));
-            var property = CommonGenerator.UppercaseFirstLetter(path.Last());
-            var navContainsTarget = pathSegment.NavigationProperty.ContainsTarget;
+            var type = GetCsharpClassName(navPathSegment, path.ToList().GetRange(0, path.Count - 1));
+            var property = path.Last();
+            var uppercasedProperty = CommonGenerator.UppercaseFirstLetter(property);
+            var structuredTypeProperty = GetElementTypeDefinitionProperties(navPathSegment).FindProperty(property);
+            if (structuredTypeProperty == null)
+            {
+                throw new Exception($"{navPathSegment.NavigationProperty.Name} does not contain property {property}");
```
baywet:
```suggestion
                throw new InvalidOperationException($"{navPathSegment.NavigationProperty.Name} does not contain property {property}");
```

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -767,19 +767,50 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
...
         /// <param name="pathSegment">The OdataPathSegment in use</param>
         /// <param name="path">path in edm model</param>
         /// <returns>ICollectionPage Interface name</returns>
-        private string GetCollectionPageClassName(NavigationPropertySegment pathSegment, ICollection<string> path)
+        private string GetCollectionPageClassName(NavigationPropertySegment navPathSegment, ICollection<string> path)
         {
-            var type = GetCsharpClassName(pathSegment, path.ToList().GetRange(0, path.Count - 1));
-            var property = CommonGenerator.UppercaseFirstLetter(path.Last());
-            var navContainsTarget = pathSegment.NavigationProperty.ContainsTarget;
+            var type = GetCsharpClassName(navPathSegment, path.ToList().GetRange(0, path.Count - 1));
```
baywet:
```suggestion
            var type = GetCsharpClassName(navPathSegment, path.Take(path.Count - 1));
```

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -767,19 +767,50 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
...
+        }
+
         /// <summary>
         /// Generates CollectionWithReferencesPage class name
         /// </summary>
         /// <param name="pathSegment">The OdataPathSegment in use</param>
         /// <param name="path">path in edm model</param>
         /// <returns>ICollectionPage Interface name</returns>
-        private string GetCollectionPageClassName(NavigationPropertySegment pathSegment, ICollection<string> path)
+        private string GetCollectionPageClassName(NavigationPropertySegment navPathSegment, ICollection<string> path)
```
baywet: defensive programming on path, if you don't have anything in the collection (or if it's null) the method will derail

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -767,19 +767,50 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
...
+            var property = path.Last();
+            var uppercasedProperty = CommonGenerator.UppercaseFirstLetter(property);
+            var structuredTypeProperty = GetElementTypeDefinitionProperties(navPathSegment).FindProperty(property);
+            if (structuredTypeProperty == null)
+            {
+                throw new Exception($"{navPathSegment.NavigationProperty.Name} does not contain property {property}");
+            }
+            if (!(structuredTypeProperty is IEdmNavigationProperty) )
+            {
+                throw new Exception(
```
baywet:
```suggestion
                throw new InvalidOperationException(
```

baywet (changes requested):

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
fey101: @baywet , are you reviewing the most recent changes? We worked with Mustafa to simplify some of the contained code in a C# friendly manner. The most recent updates, include the commit "Csharp Syntax simplification".

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
fey101: same as above. this code block is outdated.

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
fey101: Outdated code block

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
fey101: Thanks for the suggestions. Learnt on how to use a IEnumerable.Take. However, codeblock outdated as well.

fey101: Also `Path.Take` might not work especially in a situation where there's only 1 item in the path.

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -768,16 +768,24 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
...
         /// <summary>
-        /// Generates CollectionPage class name
+        /// Generates CollectionPage and CollectionWithReferencesPage class name
         /// </summary>
         /// <param name="pathSegment">The OdataPathSegment in use</param>
         /// <param name="path">path in edm model</param>
         /// <returns>ICollectionPage Interface name</returns>
         private string GetCollectionPageClassName(ODataPathSegment pathSegment, ICollection<string> path)
         {
             var type = GetCsharpClassName(pathSegment, path.ToList().GetRange(0, path.Count - 1));
```
baywet, line 778: I'm not sure what's the value of this get range since it's taking all the elements. Maybe a clone would be enough if the goal is to clone the collection?

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
```diff
@@ -768,16 +768,24 @@ private string GetCsharpClassName(ODataPathSegment pathSegment, ICollection<stri
...
-            var property = CommonGenerator.UppercaseFirstLetter(path.Last());
-            return $"{type}{property}CollectionPage";
+            var property = path.Last();
+            var uppercasedProperty = CommonGenerator.UppercaseFirstLetter(property);
+
+            var structuredType = (pathSegment.EdmType as IEdmCollectionType)?.ElementType?.Definition as IEdmStructuredType;
+            var containsTarget = (structuredType?.FindProperty(property) as IEdmNavigationProperty)?.ContainsTarget
+                ?? throw new Exception("Path doesn't match with the metadata property description!" +
```
baywet, line 784:
```suggestion
                ?? throw new InvalidOperationException("Path doesn't match with the metadata property description!" +
```

baywet (changes requested):
Ignore my last batch of comments, it seams that I have reviewed the merge 🤦‍♂️

CodeSnippetsReflection/LanguageGenerators/CSharpGenerator.cs
fey101: No, the goal is not to clone. The goal is to get the parent type( i.e. 1 level lower than the last element in the path). Using Take raises an error error CS1503: Argument 2: cannot convert from 'System.Collections.Generic.IEnumerable<string>' to 'System.Collections.Generic.ICollection<string>. On that line. Both with `var type = GetCsharpClassName(pathSegment, path.ToList().Take(path.Count - 1));` as well as `var type = GetCsharpClassName(pathSegment, path.Take(path.Count - 1));`

zengin: This doesn't take all the elements, it gets all the elements except the last one. Second parameter is not `endIndex` but `count`.

Maybe the following is less confusing:
```suggestion
            var type = GetCsharpClassName(pathSegment, path.ToList().Take(path.Count - 1));
```

baywet: alternatively you could also have used [skip last ](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.skiplast?view=net-5.0)

baywet (approved):

PR #585, opened 2021-05-26
Sanitize telemetry request urls using uri template

irvinesunday:
This PR proposes:
- Updating the `CustomPIIFilter.cs` class to redact numbers from requests urls. A test has been added and a few updated to validate this.
- Using the `UriTemplateMatcher` table populated by the `PermissionsService` to sanitize telemetry request urls by replacing them with matching uri templates from the table.
- All telemetry request urls with the pattern `/permissions?requestUrl=xyz` or `/openapi?url=xyz` will be sanitized using uri template matching:
  - Any query params present will be stripped off from the param values for `requestUrl` or `url`. This is required before performing the uri template matching.
  - The resulting base url will then be matched against the `PermissionsService` `UriTemplateMatcher` table. This table exposes a list of url templates for Graph endpoints. It is the same one used when matching permissions requests.
  - Any params that don't match the query param format -> `/permissions?requestUrl=xyz` or `/openapi?url=xyz`, for example `/samples?search='hello world'` (this is a valid query param for searching sample queries) or `/permissions?request=/abc?$filter(displayName eq 'Megan Bowen'` (this is an invalid query param - but still expected) will not be sanitized using the uri template matching method. (**NB**: For telemetry, we don't really care much about any subsequent query params coming in through the main query params (`requestUrl` or `url`), hence why we strip them off. We cannot account for all cases of incoming query params that are invalid, and thus they will be sanitized through the other available options.)
  - The above checks apply to Request telemetry and Event telemetry request urls. Trace telemetry will be sanitized using the other sanitization checks available.
  - Urls will still go through all other sanitization checks even after the above steps. 
- Telemetry tests have been updated in the `TelemetryService.Test` project.
- To enable to mock out the `UriTemplateMatcher` table from the `PermissionsService`, a `PermissionStoreFactoryMock` class has been created and used to create mocked instances of `IPermissionsStore` objects. This helps us get mocked instances of the `PermissionsService` class in both the `TelemetryService.Test` and `PermissionsService.Test` projects whilst maintaining code integrity and sanity. 
- The  `TelemetryService.Test` project now exposes sample JSON test files for the permissions endpoints mapping to be used for testing. 
- `PermissionsService.Test` has been updated to take advantage of the factory class as well. Some minor method renamings have been done as well to encourage consistency.
- A new method for retrieving the `UriTemplateMatcher` table from the `PermissionsService` has been added to the interface `IPermissionsStore`. This will be used by the `TelemetryService` to get access to the aforementioned table. Note that, `PermissionsService` implements `IPermissionsStore`.
- Permissions seeding now happens on object creation of `PermissionsStore`. This is important so as to ensure the `UriTemplateMatcher` table being used in the `TelemetryService` is always in a valid state.
- The `Extensions.cs` class in the `UtilityService` library has been updated with a new method. This method `RemoveParantheses` is used to sanitize a url before attempting to find a matching template in the `UriTemplateMatcher` table.

**Edit**
- Adding a new extension method for resolving Graph-centric paths by ensuring keys are represented as segments and action/function names are simplified by removing the namespace reference. example: `education/classes(educationClass-id)schools/microsoft.graph.delta()` resolves to `education/classes/educationClass-id/schools/delta`.
The `CustomPIIFilter.cs` and `PermissionsStore.cs` classes will use this extension method in resolving incoming request urls before attempting to match with the `UriTemplateMatcher`.
- Remove sanitization for non-target query param values: We are updating to sanitize only values for below target query params:
  `/permissions?requestUrl=xyz` or
   `/openapi?url=xyz`
 We expect that the above query param values contain Graph urls. It will be unfeasible sanitizing values from all query params,. some legitimate examples of query params that might be affected include '&**openApiVersion**=2&**graphVersion**=v1.0'. Hence, any PII values entered in query params not `requestUrl` or `url` shall not be sanitized.


MaggieKimani1 (approved):

UtilityService/Extensions.cs
```diff
@@ -42,5 +42,20 @@ public static string Query(this string uri)
             var regex = new Regex(@"(?<=\?)(.*)");
             return regex.Match(uri).Value;
         }
+
+        /// <summary>
+        /// Removes matching open and close parentheses (including the enclosed content) from a string.
+        /// </summary>
+        /// <param name="value">The target string value.</param>
+        /// <returns>The string value without the open and close parentheses.</returns>
+        public static string RemoveParantheses(this string value)
```
MIchaelMainer, line 51: The name should be changed to RemoveParenthesisKey() since the regex results in the removal of the entire key.

While we promote the [key-as-segment convention](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_KeyasSegmentConvention) in our docs and samples, we also need to support parenthesis-style convention.

As a consideration, instead of removing the parenthesis and all inside of it, it may make sense to just normalize it with key-as-segment.

UtilityService/Extensions.cs
```diff
@@ -42,5 +42,20 @@ public static string Query(this string uri)
...
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return value;
+            }
+
+            return Regex.Replace(value, @"\(.*?\)", string.Empty);
```
MIchaelMainer, line 58: Since few humans remember regex expression syntax, please provide an example like:

// example input: `/v1.0/users('MeganB@M365.onmicrosoft.com')`
// example input: `/v1.0/users('3bf94edc-10ba-45ab-9664-0dc645461f99')`
// returns: `/v1.0/users`

MIchaelMainer (dismissed):
This is more like an Approved with Suggestions

TelemetryService.Test/TelemetryService.Test.csproj
```diff
@@ -6,6 +6,24 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <ItemGroup>
```
baywet, line 9: I believe you don't need that block since you have a content block below for the same files.

MockTestUtility/PermissionStoreFactoryMock.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <param name="httpClientUtility">Optional: Mock <see cref="IHttpClientUtility"/> object.</param>
+        /// <param name="permissionsCache">Optional: Mock <see cref="IMemoryCache"/> object.</param>
+        /// <returns>A mocked IPermissionsStore object.</returns>
+        public static IPermissionsStore GetPermissionStore(string configFilePath = null,
+                                                           IConfigurationRoot configuration = null,
+                                                           IFileUtility fileUtility = null,
+                                                           IHttpClientUtility httpClientUtility = null,
+                                                           IMemoryCache permissionsCache = null)
+        {
+            if (string.IsNullOrEmpty(configFilePath) && configuration == null)
```
baywet, line 35: break apart, saying "one of these values might be null" is less useful thank saying "this value is null"

PermissionsService.Test/PermissionsStoreShould.cs
```diff
@@ -17,22 +13,11 @@ namespace PermissionsService.Test
...
         {
-            _fileUtility = new FileUtilityMock();
-            _httpClientUtility = new FileUtilityMock();
-            _permissionsCache = Create.MockedMemoryCache();
-            _configuration = new ConfigurationBuilder()
-               .AddJsonFile(".\\TestFiles\\appsettingstest-valid.json")
-               .Build();
-
-            _permissionsStore = new PermissionsStore(_configuration, _httpClientUtility, _fileUtility, _permissionsCache);
+            _permissionsStore = PermissionStoreFactoryMock.GetPermissionStore(".\\TestFiles\\appsettingstest-valid.json");
```
baywet, line 20: use a constant for the folder name

TelemetryService.Test/CustomPIIFilterShould.cs
```diff
@@ -13,24 +15,76 @@ namespace Telemetry.Test
...
+            var renderedMessage = $"HTTP {httpMethod + requestPath} responded {statusCode} in {elapsed} ms";
+
+            var eventTelemetry = new EventTelemetry();
+            eventTelemetry.Properties.Add("RequestPath", requestPath);
+            eventTelemetry.Properties.Add("RequestMethod", httpMethod);
+            eventTelemetry.Properties.Add("StatusCode", statusCode);
+            eventTelemetry.Properties.Add("RenderedMessage", renderedMessage);
+
+            // Act
+            if (eventTelemetry.Properties.ContainsKey("RequestPath") && eventTelemetry.Properties.ContainsKey("RenderedMessage"))
```
baywet, line 63: I'm not sure whether you're testing this since you added it right above?

TelemetryService/CustomPIIFilter.cs
```diff
@@ -138,17 +148,53 @@ public void Process(ITelemetry item)
...
+
+                if (queryValue.Contains("url=", StringComparison.OrdinalIgnoreCase))
+                {
+                    /* Sanitize first using uri templates paths supplied from below example paths:
+                        /permissions?requestUrl=xyz or
+                        /openapi?url=xyz
+                       Possibility that these could be valid Graph urls.
+                    */
+
+                    var valueIndex = queryValue.IndexOf('=');
```
baywet, line 177: maybe add 1 here instead so you don't have to repeat it below?

baywet (changes requested):

UtilityService/Extensions.cs
irvinesunday: Thanks for the heads up. I might have just discovered a better way around it.

UtilityService/Extensions.cs
irvinesunday: Since we don't want to remove paranthesis-styled keys from paths, I've added the below as examples instead in the XML documentation:
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/6377bdf7a4d3a25735a9e977ed31405a70bf5fef/UtilityService/Extensions.cs#L53-L61

TelemetryService.Test/CustomPIIFilterShould.cs
irvinesunday: That's really unnecessary. Thanks.

MockTestUtility/PermissionStoreFactoryMock.cs
irvinesunday: I've added helpful comment to try and convey the intention:
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/7ac4efe7d7d103e6dfda26fd39c4a7fe8f7cca58/MockTestUtility/PermissionStoreFactoryMock.cs#L35-L41

baywet (dismissed):

TelemetryService.Test/CustomPIIFilterShould.cs
```diff
@@ -259,21 +225,8 @@ public void RedactFirstNameFromEventTelemetry(string requestPath, string expecte
...
-        [InlineData("/openapi?endpoint=/xyz?$filter=displayName%20Eq%20%27Meghan%27",
-                    "/openapi?endpoint=/xyz?$filter=displayName Eq ****")]
-
-        [InlineData("/openapi?test=/students?$filter=startswith(displayName, 'a')&$count=true&$top=1&$orderby=displayName",
-                    "/openapi?test=/students?$filter=startswith****&$count=true&$top=****&$orderby=displayName")]
-
-        [InlineData("/permissions?foo=/students('MeganB@M365x214355.onmicrosoft.com')",
-                    "/permissions?foo=/students('****')")]
+        [InlineData("/permissions?requesturl=/students('MeganB@M365x214355.onmicrosoft.com')",
+                    "/permissions?requesturl=/students/'****'")]
```
MIchaelMainer, line 229: I'd expect the output to be `/permissions?requesturl=/students/****`, not `/permissions?requesturl=/students/'****'`

MIchaelMainer: Could we make this so that it tests that we properly form intermediate segments as well.

`/permissions?requesturl=/students('MeganB@M365x214355.onmicrosoft.com')/classes`
`/permissions?requesturl=/students/****/classes`

UtilityService.Test/UtilityServiceExtensionsShould.cs
```diff
@@ -39,5 +39,37 @@ public void GetUriQuery(string targetUri, string expectedQuery)
             // Assert
             Assert.Equal(expectedQuery, actualQuery);
         }
+
+        [Theory]
+        [InlineData("microsoft.graph.delta()",
+                    "microsoft.graph.delta")]
+        [InlineData("worksheets/microsoft.graph.range(address={address})",
+                    "worksheets/microsoft.graph.range")]
+        public void RemoveParantheses(string targetString, string expectedString)
```
MIchaelMainer, line 48: Global/Spelling. 

RemoveParantheses => RemoveParentheses

MIchaelMainer: Also, add a test case for composable function. It works as is.

        [InlineData("worksheets/microsoft.graph.range(address={address})/microsoft.graph.range(address={address})",
                    "worksheets/microsoft.graph.range/microsoft.graph.range")]

UtilityService.Test/UtilityServiceExtensionsShould.cs
```diff
@@ -39,5 +39,37 @@ public void GetUriQuery(string targetUri, string expectedQuery)
...
+        [Theory]
+        [InlineData("/users('MeganB@M365x214355.onmicrosoft.com')",
+                    "/users/'MeganB@M365x214355.onmicrosoft.com'")]
+        [InlineData("/education/schools(id)users/microsoft.graph.delta()",
+                    "/education/schools/id/users/delta")]
+        [InlineData("/users/{user-id}/insights/used(usedInsight-id)resource/microsoft.graph.workbookWorksheet/microsoft.graph.range(address={address})",
+                    "/users/{user-id}/insights/used/usedInsight-id/resource/workbookWorksheet/range")]
+        [InlineData("/groupLifecyclePolicies(groupLifecyclePolicy-id)microsoft.graph.remove",
+                    "/groupLifecyclePolicies/groupLifecyclePolicy-id/remove")]
+        public void GetUriTemplatePathFormat(string targetString, string expectedString)
```
MIchaelMainer, line 66: Add test case for composable function.

UtilityService/Extensions.cs
MIchaelMainer: That is different than parenthesis-styled keys. These are OData functions. We don't want to treat them the same.

MIchaelMainer (changes requested):

TelemetryService.Test/CustomPIIFilterShould.cs
irvinesunday: With regard to masking the email address value including the quotation marks as:
`/permissions?requesturl=/students('MeganB@M365x214355.onmicrosoft.com')` to be resolved to `/permissions?requesturl=/students/****` as opposed to `/permissions?requesturl=/students/'****'`
I am of the opinion that this shouldn't be a hard requirement as long as the crucial piece of information, the email address, is been masked successfully. 
Updating the email regex: https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/1d03de986e7f5496c3f2f09f3d927fc675a8f970/TelemetryService/CustomPIIFilter.cs#L31 to also match non-alphanumeric characters preceding or succeeding the email address value (except for the expected _  - and a .) would inadvertently also end up masking the parentheses or slashes - in such cases: `/students(MeganB@M365x214355.onmicrosoft.com)` or `/students/MeganB@M365x214355.onmicrosoft.com` inaccurately resolving to `/students****`

UtilityService.Test/UtilityServiceExtensionsShould.cs
irvinesunday: Added.

UtilityService.Test/UtilityServiceExtensionsShould.cs
irvinesunday: Added

TelemetryService.Test/CustomPIIFilterShould.cs
MIchaelMainer: Agreed about it not being a hard requirement. It does what it needs to do.

MIchaelMainer (approved):

Issue #590, opened 2021-06-02
Replace UriTemplateMatcher with OpenApiUrlTreeNode in OpenAPIService


irvinesunday:
The `OpenApiUrlTreeNode`is now available in `Microsoft.OpenApi.NET` `v1.3.0-preview`  
[AB#9669](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9669)

PR #592, opened 2021-06-04
Replaces UriTemplateMatcher with OpenApiUrlTreeNode in OpenAPIService

irvinesunday:
Closes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/590

This PR:
- Bumps the `Microsoft.OpenApi.Readers` lib. to `v1.3.0-preview`. This is to get the new `OpenApiUrlTreeNode` data structure.
- Replaces the `UriTemplateMatcher` with the `OpenApiUrlTreeNode` for matching urls. The latter provides an easier, efficient and faster matching experience compared to the former, which uses regexes. This should actualize faster response times for the Graph Explorer autocomplete feature or any client requesting a subset OpenAPI doc using a url, ex: `/openapi?url=/users/{user-id}`.
- Because we need to provide a label for each `OpenApiDocument` that gets appended to the `OpenApiUrlTreeNode` data structure, the method `CreatePredicate()` adds a new parameter `graphVersion` with a default value: `v1.0`. Consequently, the `OpenApiController` has been updated to provide this new argument when calling the aforementioned method. Respective tests have also been updated to account for the new parameter (where applicable). This param is defaulted because only calls to the `CreatePredicate()` method with a `url` value need only supply a value for the `graphVersion` param.
- Besides others, `OpenApiService.cs` adds a new `private` method  `GetOpenApiOperations()` that is responsible for matching a url string value to the children segments of the `OpenApiUrlTreeNode` data structure and pulling out the `OpenApiOperations` under the path.
- Unless a `forceRefresh` is requested, the `OpenApiUrlTreeNode` will persist paths and operations for `v1.0` and `beta` in the same data structure.
- Adds a new `Utils.cs` class in the `Extensions` library. This contains common utility functions `CheckArgumentNull()` and `CheckArgumentNullOrEmpty()` that can be used across the solution for a standardized way of checking for nullability of params. In `OpenApiService.cs`, its use has been employed in the method `GetOpenApiOperations()` 
- Updates the `UriTemplatePathFormat()` method: adds a new `bool` param - `simplifyNamespace` that provides an option on whether to simplify `action`/`functions` namespaces in path values. In the `OpenApiUrlTreeNode` data structure matching we don't necessarily want to simplify the namespace; however, in the `PermissionsStore.cs` and `CustomPIIFilter.cs` matching, we do need to simplify the namespace - because for the latter two they use the JSON permissions doc. urls as their source of templates for matching; and these don't expose the fully qualified names for `actions`/`functions` in their values. 
- `Extensions.cs` tests for the above updated feature has been added for validation.
- In the `Visit(OpenApiSchema schema)` method in the `PowershellFormatter.cs` class, we now have to account for the fact that `AdditionalProperties` in schemas are now being _walked_. To avoid endlessly creating and walking the `AdditionalProperties` schemas in an infinite recursion, we are using a `Stack` to keep track of the schemas already visited.
- Update the _OpenApi.yaml_ file version and server values.
- Updates some XML documentations.

### DIFF
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -121,7 +121,7 @@ namespace OpenAPIService
         /// <param name="graphVersion">Version of the target Microsoft Graph API.</param>
         /// <param name="forceRefresh">Whether to reload the OpenAPI document from source.</param>
         /// <returns>A predicate.</returns>
-        public static async Task<Func<OpenApiOperation, bool>> CreatePredicate(string operationIds, string tags, string url,
+        public static Func<OpenApiOperation, bool> CreatePredicate(string operationIds, string tags, string url,
             OpenApiDocument source, string graphVersion = "v1.0", bool forceRefresh = false)
         {
             if (url != null && (operationIds != null || tags != null))
@@ -191,7 +191,7 @@ namespace OpenAPIService
                 throw new InvalidOperationException("Either operationIds, tags or url need to be specified.");
             }
 
-            return await Task.FromResult(predicate);
+            return predicate;
         }
 
         /// <summary>
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -47,7 +47,6 @@ namespace OpenAPIService
         /// <returns>A partial OpenAPI document.</returns>
         public static OpenApiDocument CreateFilteredDocument(OpenApiDocument source, string title, string graphVersion, Func<OpenApiOperation, bool> predicate)
         {
-
             var subset = new OpenApiDocument
             {
                 Info = new OpenApiInfo()
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -20,7 +20,6 @@ using System.Threading.Tasks;
 using System.Collections.Concurrent;
 using System.Text;
 using OpenAPIService.Common;
-using UriMatchingService;
 using UtilityService;
 
 namespace OpenAPIService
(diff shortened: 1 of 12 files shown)


OpenAPIService/OpenApiService.cs
```diff
@@ -191,37 +191,115 @@ public static OpenApiDocument CreateFilteredDocument(OpenApiDocument source, str
                 throw new InvalidOperationException("Either operationIds, tags or url need to be specified.");
             }
 
-            return predicate;
+            return await Task.FromResult(predicate);
```
baywet, line 194: In general you don't need an await when calling Task.FromResult if you don't have other awaits in the method, just remove it as well as the async. You should even remove the Task from the return type all together as no actual async operation is performed in this method.

OpenAPIService/OpenApiService.cs
```diff
@@ -191,37 +191,115 @@ public static OpenApiDocument CreateFilteredDocument(OpenApiDocument source, str
...
+                        break;
+                    }
+
+                    /* Attempt to get the parameter segment from the children of the current node:
+                     * We are assuming a failed match because of different parameter namings
+                     * between the relative url segment and the corresponding OpenApiUrlTreeNode segment name
+                     * ex.: matching '/users/12345/messages' with '/users/{user-id}/messages'
+                     */
+                    tempTargetChild = targetChild.Children
+                                                 .FirstOrDefault(x => x.Value.IsParameter == true).Value;
```
baywet, line 266:
```suggestion
                                                 .FirstOrDefault(x => x.Value.IsParameter).Value;
```

OpenAPIService/PowershellFormatter.cs
```diff
@@ -79,13 +85,27 @@ public override void Visit(OpenApiOperation operation)
...
+        /// </summary>
+        /// <param name="schema">The target <see cref="OpenApiSchema"/></param>
         public override void Visit(OpenApiSchema schema)
         {
+            if (_schemaLoop.Contains(schema))
+            {
+                return; // loop detected, this schema has already been walked.
+            }
+
             if (schema?.Type == "object")
```
baywet, line 99:
```suggestion
            if ("object".Equals(schema?.Type, StringComparison.OrdinalIgnoreCase))
```

baywet (changes requested):

baywet (dismissed):

baywet (approved):

PR #593, opened 2021-06-07
Add Extra Instrumentation using Application Insights API

### LINKED ISSUE #560: Use App Insights API to instrument code
[AB#9351](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9351)

MaggieKimani1:
This PR proposes using the Application Insights API to send custom metrics by inserting a few lines of code in order to help better diagnose issues related to our API and obtain the stack trace whenever an exception is thrown.

We're using an instance of TelemetryClient from the dependency injection container and inserting methods such as:
1. `TrackTrace` by adding trace messages for better logging and to capture diagnostic data and inspect them in Diagnostic Search tab in the App Insights portal
2. `TrackException` to send exception messages to Application Insights.

These methods have been plugged into the following controllers and service libraries:

- SamplesStore and Samples Controller
- PermissionsStore and Permissions Controller
- ChangesStore, ChangesService and the ChangesController
- OpenAPIService and OpenApi Controller

This helps to enrich our data collection, enhance tracking and examination of individual occurrences of exceptions and improving QoS in DevX API.

Resource: https://docs.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics
Fixes #560


adhiambovivian:
@MaggieKimani1  @baywet any reason why the telemetry & app monitoring on devxAPI is tightly coupled? Wouldn't it be beneficial to decouple all the telemetry work on DevXAPI to allow the two to evolve separately without heavy dependency?

baywet:
@adhiambovivian i believe Maggie and @irvinesunday will have more context here.
My involvement with devx Api have only been have Java snippets generation, dotnet 5 migration and high level pull requests reviews.

ChangesService/Services/ChangesStore.cs
```diff
@@ -51,19 +58,36 @@ public async Task<ChangeLogRecords> FetchChangeLogRecordsAsync(CultureInfo cultu
         {
             var locale = cultureInfo.GetSupportedLocaleVariant().ToLower(); // lowercased in line with source files
 
+            _telemetry?.TrackTrace($"Retrieving changelog records for locale '{locale}' from in-memory cache '{locale}'",
```
fey101, line 61: I think it would be prudent to remove args from the TrackTrace.message and put them in TrackTrace.properties so as to have all related logs bundled up together and filterable. Otherwise the logs may be too much to process.

ChangesService/Services/ChangesStore.cs
```diff
@@ -51,19 +58,36 @@ public async Task<ChangeLogRecords> FetchChangeLogRecordsAsync(CultureInfo cultu
...
+                                   SeverityLevel.Information,
+                                   ChangesTraceProperties);
+
             // Fetch cached changelog records
             ChangeLogRecords changeLogRecords = await _changeLogCache.GetOrCreateAsync(locale, cacheEntry =>
             {
+                _telemetry?.TrackTrace($"In-memory cache '{locale}' empty. " +
```
fey101, line 68: remove args from the TrackTrace.message and put them in TrackTrace.properties

ChangesService/Services/ChangesStore.cs
```diff
@@ -51,19 +58,36 @@ public async Task<ChangeLogRecords> FetchChangeLogRecordsAsync(CultureInfo cultu
...
                     /* Check whether a previous thread already seeded an
                      * instance of the localized changelog records during the lock.
                      */
                     var lockedLocale = locale;
                     var seededChangeLogRecords = _changeLogCache.Get<ChangeLogRecords>(lockedLocale);
+                    var sourceMsg = $"Return locale '{locale}' changelog records from in-memory cache";
 
                     if (seededChangeLogRecords != null)
                     {
+                        _telemetry?.TrackTrace($"In-memory cache '{lockedLocale}' of changelog records " +
```
fey101, line 85: remove args from the TrackTrace.message and put them in TrackTrace.properties

ChangesService/Services/ChangesStore.cs
```diff
@@ -85,6 +109,12 @@ public async Task<ChangeLogRecords> FetchChangeLogRecordsAsync(CultureInfo cultu
                     var jsonFileContents = _httpClientUtility.ReadFromDocumentAsync(httpRequestMessage)
                                                 .GetAwaiter().GetResult();
 
+                    sourceMsg = $"Successfully seeded changelog records for locale '{lockedLocale}' from source";
```
fey101, line 112: remove args from the TrackTrace.message and put them in TrackTrace.properties

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -80,6 +86,10 @@ private void SeedPermissionsTables()
 
             foreach (string permissionFilePath in _permissionsBlobNames)
             {
+                _telemetry?.TrackTrace($"Seeding permissions table from file source '{permissionFilePath}'",
```
fey101, line 89: remove args from the TrackTrace.message and put them in TrackTrace.properties

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -127,8 +141,17 @@ private void SeedPermissionsTables()
         /// <returns>The localized instance of permissions descriptions.</returns>
         private async Task<IDictionary<string, IDictionary<string, ScopeInformation>>> GetOrCreatePermissionsDescriptionsAsync(string locale = DefaultLocale)
         {
+            _telemetry?.TrackTrace($"Retrieving permissions for locale '{locale}' from in-memory cache 'ScopesInfoList_{locale}'",
```
fey101, line 144: remove args from the TrackTrace.message and put them in TrackTrace.properties

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -139,21 +162,34 @@ private void SeedPermissionsTables()
...
                     var seededScopesInfoDictionary = _permissionsCache.Get<IDictionary<string, IDictionary<string, ScopeInformation>>>($"ScopesInfoList_{locale}");
+                    var sourceMsg = $"Return locale '{locale}' permissions from in-memory cache 'ScopesInfoList_{locale}'";
 
                     if (seededScopesInfoDictionary == null)
                     {
                         string relativeScopesInfoPath = FileServiceHelper.GetLocalizedFilePathSource(_permissionsContainerName, _scopesInformation, locale);
 
                         // Get file contents from source
                         string scopesInfoJson = _fileUtility.ReadFromFile(relativeScopesInfoPath).GetAwaiter().GetResult();
+                        _telemetry?.TrackTrace($"Successfully seeded permissions for locale '{locale}' from Azure blob resource",
```
fey101, line 173: remove args from the TrackTrace.message and put them in TrackTrace.properties

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -139,21 +162,34 @@ private void SeedPermissionsTables()
...
+                                               PermissionsTraceProperties);
 
                         seededScopesInfoDictionary = CreateScopesInformationTables(scopesInfoJson);
-
                         cacheEntry.AbsoluteExpirationRelativeToNow = TimeSpan.FromHours(_defaultRefreshTimeInHours);
+                        sourceMsg = $"Return locale '{locale}' permissions from Azure blob resource";
+                    }
+                    else
+                    {
+                        _telemetry?.TrackTrace($"In-memory cache 'ScopesInfoList_{locale}' of permissions " +
```
fey101, line 183: remove args from the TrackTrace.message and put them in TrackTrace.properties

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -172,6 +208,10 @@ private void SeedPermissionsTables()
                                                                                                                             string branchName,
                                                                                                                             string locale = DefaultLocale)
         {
+            _telemetry?.TrackTrace($"Retrieving permissions for locale '{locale}' from GitHub repository",
```
fey101, line 211: remove args eg. '{locale}' above, from the TrackTrace.message and put them in TrackTrace.properties

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -211,9 +255,16 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
         {
             if (string.IsNullOrEmpty(scopesInfoJson))
             {
+                _telemetry?.TrackTrace($"{nameof(scopesInfoJson)} empty or null when creating the scopes information tables",
```
fey101, line 258: remove args i.e. the {nameof(scopesInfoJson)} from the TrackTrace.message and put them in TrackTrace.properties

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -321,6 +386,10 @@ private bool RefreshPermissionsTables()
 
                 if (resultMatch == null)
                 {
+                    _telemetry?.TrackTrace($"Url '{requestUrl}' not found",
```
fey101, line 389: remove args i.e. {requestUrl} from the TrackTrace.message and put them in TrackTrace.properties

GraphWebApi/Controllers/ChangesController.cs
```diff
@@ -61,9 +69,13 @@ public ChangesController(IChangesStore changesStore, IConfiguration configuratio
                 var cultureFeature = HttpContext.Features.Get<IRequestCultureFeature>();
                 var cultureInfo = cultureFeature.RequestCulture.Culture;
 
+                _telemetry?.TrackTrace($"Request to fetch changelog records for the requested culture info '{cultureInfo}'",
```
fey101, line 72: remove args from the TrackTrace.message and put them in TrackTrace.properties

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
```diff
@@ -58,15 +63,21 @@ public async Task<IActionResult> GetSampleQueriesListAsync(string search, string
 
                 if (filteredSampleQueries.Count == 0)
                 {
-                    // Search parameter not found in list of sample queries
+                    _telemetry?.TrackTrace($"Search value: '{search}' not found in: category, humanName or tip properties of sample queries",
```
fey101, line 66: remove args from the TrackTrace.message and put them in TrackTrace.properties

GraphWebApi/Controllers/GraphExplorerSamplesController.cs
```diff
@@ -92,14 +103,19 @@ public async Task<IActionResult> GetSampleQueryByIdAsync(string id, string org,
 
                 if (sampleQueryById == null)
                 {
-                    return NotFound(); // sample query with the given id doesn't exist in the list of sample queries
+                    _telemetry?.TrackTrace($"Sample query with id: {id} doesn't exist in the list of sample queries",
```
fey101, line 106: remove args i.e. id from the TrackTrace.message and put them in TrackTrace.properties

fey101 (changes requested):

ChangesService/Services/ChangesStore.cs
MaggieKimani1: @fey101 as @irvinesunday has mentioned, according to [documentation](https://docs.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#properties), the TrackTrace properties will group the trace events that are related by default on App Insights

irvinesunday:
Need to add Trace telemetry for _CodeSnippets_ as well?

ChangesService/Services/ChangesStore.cs
fey101: @irvinesunday , Are you suggesting that TrackTrace.Properties is not the right place to place args or are you suggesting that even with the args in the message, tracktrace.Properties will group related logs by default? Not seeing where the latter is specified in the attached documentation.

ChangesService/Services/ChangesStore.cs
fey101: @MaggieKimani1 , this is what I see from the documentation



properties | Map of string to string: Additional data used to filter exceptions in the portal. Defaults to empty.
-- | --



Am wondering how it will know of related events if args in the message make the messages appear different for each unique instance.

fey101: If the case is as you stated @MaggieKimani1 , then why did they not have the db.id in the following example as part of the message?
```C#
var telemetry = new Microsoft.ApplicationInsights.TelemetryClient();
telemetry.TrackTrace("Slow database response",
                SeverityLevel.Warning,
                new Dictionary<string,string> { {"database", db.ID} });
```

MaggieKimani1: @fey101 in the above example, the db.id is used to enable filtering of all messages relating to a particular db whereas in our case, we're using the name of the controller or service library for us to keep track of all trace events fired by calling an endpoint.

ChangesService/Services/ChangesStore.cs
MIchaelMainer: Args in TrackTrace.message and TrackTrace.properties serve two related but different purposes. 

Regarding adding args to TrackTrace.properties, I think the question to ask ourselves is whether the argument is intended to be filtered or maybe queried. If I can answer in the affirmative to "is there a scenario where I want to filter on locale?", then we we should put that in TrackTrace.properties.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -89,8 +99,8 @@ private void SeedPermissionsTables()
 
                     if (permissionsObject.Count < 1)
                     {
-                        throw new InvalidOperationException($"The permissions data sources cannot be empty." +
-                            $"Check the source file or check whether the file path is properly set. File path: " +
+                        throw new InvalidOperationException("The permissions data sources cannot be empty." +
```
MIchaelMainer, line 102: nit: the other controllers explicitly handle InvalidOperationException. PermissionsController implicitly handles this with the final catch(Exception). 

This would be a good place to trigger an alert that our permission data source is empty.

MIchaelMainer (approved):
approved pending others' signoff.

irvinesunday:
Unfortunately we are sanitizing all numerical data.

image

ChangesService/Services/ChangesStore.cs
fey101: Resolving this since it makes sense that there will be no value in filtering based on locale, for now and in the foreseen future. And is useful in the message as seen [devX-API-AppInsights-Testing](https://ms.portal.azure.com/#blade/AppInsightsExtension/DetailsV2Blade/ComponentId/%7B%22SubscriptionId%22%3A%2268397ef5-9754-46e7-9572-9b15d12367f1%22%2C%22ResourceGroup%22%3A%22Graph-Explorer-API%22%2C%22Name%22%3A%22DevX-API-AppInsights-Testing%22%2C%22LinkedApplicationType%22%3A0%2C%22ResourceId%22%3A%22%252Fsubscriptions%252F68397ef5-9754-46e7-9572-9b15d12367f1%252FresourceGroups%252FGraph-Explorer-API%252Fproviders%252Fmicrosoft.insights%252Fcomponents%252FDevX-API-AppInsights-Testing%22%2C%22ResourceType%22%3A%22microsoft.insights%252Fcomponents%22%2C%22IsAzureFirst%22%3Afalse%7D/DataModel/%7B%22cacheId%22%3A%2237340994-198a-4857-9d14-6c94f21b17bf%22%2C%22eventTable%22%3A%22traces%22%2C%22timeContext%22%3A%7B%22durationMs%22%3A86400000%2C%22endTime%22%3A%222021-06-09T18%3A08%3A42.565Z%22%7D%7D)

GraphExplorerPermissionsService/Services/PermissionsStore.cs
fey101: is there a scenario where I want to filter on requestUrl? I feel like this is a valid case but if it's not, @MaggieKimani1 , feel free to resolve it.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
MaggieKimani1: I'll definitely add that trace message where that exception is thrown. Thank you for noticing that!

fey101 (approved):

ChangesService/Services/ChangesService.cs
```diff
@@ -23,6 +25,22 @@ public static class ChangesService
     {
         // Field to hold key-value pairs of url and workload names
         private static readonly Dictionary<string, string> _urlWorkloadDict = new();
+        private static readonly IDictionary<string, string> _changesTraceProperties = new Dictionary<string, string> { { "Changes", "ChangesService" } };
+        private static readonly object _telemetrySetLock = new();
+        private static TelemetryClient _telemetryClient;
+        public static TelemetryClient TelemetryClient
```
baywet, line 31: just out of curiosity, have you considered using a Lazy if you want a singleton?

OpenAPIService/OpenApiService.cs
```diff
@@ -36,6 +38,22 @@ public class OpenApiService
     {
         private static readonly ConcurrentDictionary<Uri, OpenApiDocument> _OpenApiDocuments = new ConcurrentDictionary<Uri, OpenApiDocument>();
         private static OpenApiUrlTreeNode _openApiRootNode = OpenApiUrlTreeNode.Create();
+        private static readonly IDictionary<string, string> _openApiTraceProperties = new Dictionary<string, string> { { "OpenApi", "OpenApiService" } };
+        private static readonly object _telemetrySetLock = new();
+        private static TelemetryClient _telemetryClient;
+        public static TelemetryClient TelemetryClient
```
baywet, line 44: why is this repeated? couldn't we encapsulate it away in a telemetry service or something similar to make it reusable?

baywet (changes requested):

ChangesService/Services/ChangesService.cs
MaggieKimani1: I hadn't considered implementing it that way.
Thank you for the suggestion

TelemetryClient/ITelemetryClientSingleton.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace TelemetryClientFactory
+{
+    public interface ITelemetryClientSingleton
+    {
+      //  TelemetryClient GetTelemetryClient();
```
baywet, line 12: ?

TelemetryClient/TelemetryClientSingleton.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Microsoft.ApplicationInsights;
+
+namespace TelemetryClientFactory
+{
+    public sealed class TelemetryClientSingleton
+    {
+        public static TelemetryClient TelemetryClient { get; private set; }
```
baywet, line 7: duplicated with the same class in TelemetryClientFactory folder, remove

GraphWebApi/Startup.cs
```diff
@@ -56,7 +57,33 @@ public void ConfigureServices(IServiceCollection services)
...
+            if (!_env.IsDevelopment())
+            {
+                services.ConfigureTelemetryModule<QuickPulseTelemetryModule>((module, o) =>
+                    module.AuthenticationApiKey = Configuration["ApplicationInsights:AppInsightsApiKey"]);
+            }
+
+            #endregion
+
             services.AddMemoryCache();
+            services.AddSingleton<TelemetryClientSingleton>();
```
baywet, line 86: shouldn't the interface be present here as well so it can be injected into dependant services?

ChangesService/Services/ChangesService.cs
```diff
@@ -58,6 +61,13 @@ public static ChangeLogRecords DeserializeChangeLogRecords(string jsonString)
                                                               MicrosoftGraphProxyConfigs graphProxyConfigs,
                                                               IHttpClientUtility httpClientUtility = null)
         {
+            TelemetryClientSingleton.TelemetryClient?
```
baywet, line 64: why make a static reference? inject it in the constructor instead and have it as an instance property to get better decoupling.

TelemetryClientFactory/TelemetryClientSingleton.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace TelemetryClientWrapper
+{
+    /// <summary>
+    /// Provides access to a singleton instance of a <see cref="Microsoft.ApplicationInsights.TelemetryClient"/>.
+    /// </summary>
+    public sealed class TelemetryClientSingleton
+    {
+        public static TelemetryClient TelemetryClient { get; private set; }
+
+        public TelemetryClientSingleton(TelemetryClient telemetryClient)
```
baywet, line 17: Why would you implement the singleton pattern yourself? the framework does this for you in the startup configuration. Remove

baywet (changes requested):

TelemetryClientFactory/TelemetryClientSingleton.cs
MaggieKimani1: There's a cyclic DI issue caused by injecting the IPermissionsStore into the constructor of the CustomPIIFilter class(it relies on the **UrlTemplateMatcher** from our PermissionsStore class for sanitization), and the TelemetryClient into the constructor of the PermissionsStore.
Implementing it this way helps to decouple the two classes as I couldn't reference the telemetry client via DI in the PermissionStore class without causing the DI issue. 

Is there a better approach that you can recommend?

baywet: I think we're coming back to @adhiambovivian 's comment about telemetry and APM being decoupled, maybe an opportunity for us to clean this as well?

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -49,15 +51,20 @@ public class PermissionsStore : IPermissionsStore
...
 
-        public PermissionsStore(IConfiguration configuration, IHttpClientUtility httpClientUtility,
-                                IFileUtility fileUtility, IMemoryCache permissionsCache)
+        public PermissionsStore(IConfiguration configuration, 
+            IHttpClientUtility httpClientUtility, 
+            IFileUtility fileUtility, 
+            IMemoryCache permissionsCache, 
+            TelemetryClient telemetryClient)
         {
+            _telemetryClient = telemetryClient;
```
baywet, line 60: defensive programming, otherwise the methods will fail as it's readonly

MockTestUtility/PermissionStoreFactoryMock.cs
```diff
@@ -47,7 +47,7 @@ public static class PermissionStoreFactoryMock
                                     .AddJsonFile(configFilePath)
                                     .Build();
 
-            return new PermissionsStore(configuration, httpClientUtility, fileUtility, permissionsCache);
+            return new PermissionsStore(configuration, httpClientUtility, fileUtility, permissionsCache, null);
```
baywet, line 50: mock the telemetry client

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -49,15 +51,20 @@ public class PermissionsStore : IPermissionsStore
         private const string PermissionsContainerBlobConfig = "BlobStorage:Containers:Permissions";
         private const string NullValueError = "Value cannot be null";
 
-        public PermissionsStore(IConfiguration configuration, IHttpClientUtility httpClientUtility,
-                                IFileUtility fileUtility, IMemoryCache permissionsCache)
+        public PermissionsStore(IConfiguration configuration,
```
baywet, line 54: The changes service needs a similar modification, the singleton needs to be removed

baywet (changes requested):

MockTestUtility/PermissionStoreFactoryMock.cs
MaggieKimani1: This won't be necessary as the telemetry client param is optional

TelemetryService/CustomPIIFilter.cs
```diff
@@ -69,17 +70,15 @@ public class CustomPIIFilter : ITelemetryProcessor
...
-
-            if (uriTemplateMatcher == null)
+            if (serviceProvider == null)
             {
-                throw new ArgumentNullException(nameof(uriTemplateMatcher), $"{ uriTemplateMatcher }: { nameof(uriTemplateMatcher) }");
+                throw new ArgumentNullException(nameof(serviceProvider), $"{ serviceProvider }: { nameof(serviceProvider) }");
             }
-
-            _uriTemplateMatcher = uriTemplateMatcher;
+            _serviceProvider = serviceProvider;
```
baywet, line 81:
```suggestion
            _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider), $"{ serviceProvider }: { nameof(serviceProvider) }");;
```

And delete the block above

TelemetryService/CustomPIIFilter.cs
```diff
@@ -159,6 +158,10 @@ public void Process(ITelemetry item)
         /// <returns>The string url with PII sanitized from its query path.</returns>
         private string SanitizeUrlQueryPath(string url)
         {
+            // use the service provider to lazily get an IPermissionsStore instance
+            var permissionsStore = _serviceProvider.GetRequiredService<IPermissionsStore>();
```
baywet, line 162: why don't you pass the IPermissionStore into the constructor directly and let the dependency injection framework do the work for you?

baywet (changes requested):

TelemetryService/CustomPIIFilter.cs
MaggieKimani1: This will cause a circular dependency which is what we are trying to avoid by using this approach

GraphWebApi/Startup.cs
baywet: in the latest update this has been removed, but I believe we're not adding the telemetry client to the list of services anymore.

TelemetryClientFactory/TelemetryClientSingleton.cs
baywet: resolving as this as been worked around by calling the service provider within the service itself. Not ideal but works for now without requiring a major revamp.

GraphWebApi/Startup.cs
baywet: nevermind, the Telemetry client class in use is the one provided by application insights and injected by the AI section. resolving

baywet (approved):
Thanks for making all the changes! 🚢

PR #628, opened 2021-07-07
Merges dev into master

irvinesunday:
* Update how permissions values are fetched

* Update path to the new permissions files container

* Update test files structure

* Update file path

* Make linq query more readable

* Refactor code

Co-authored-by: Irvine Sunday <irochand@microsoft.com>
Co-authored-by: Irvine Sunday <40403681+irvinesunday@users.noreply.github.com>

### DIFF
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -397,13 +397,22 @@ namespace GraphExplorerPermissionsService
                     return null;
                 }
 
-                JArray resultValue = new JArray();
-                resultValue = (JArray)_scopesListTable[int.Parse(resultMatch.Key)];
-
-                var scopes = resultValue.FirstOrDefault(x => x.Value<string>("HttpVerb") == method)?
-                    .SelectToken(scopeType)?
-                    .Select(s => (string)s)
-                    .ToArray();
+                string[] scopes = null;
+                var resultValue = _scopesListTable[int.Parse(resultMatch.Key)] as JToken;
+                foreach (JProperty httpVerb in resultValue)
+                {
+                    if (httpVerb.Name.Equals(method, StringComparison.OrdinalIgnoreCase))
+                    {
+                        foreach (JProperty scopeCategory in httpVerb.Value)
+                        {
+                            if (scopeCategory.Name.Equals(scopeType, StringComparison.OrdinalIgnoreCase))
+                            {
+                                scopes = scopeCategory.Value?.ToObject<string[]>();
+                                break;
+                            }
+                        }
+                    }
+                }
 
                 if (scopes == null)
                 {
--- a/PermissionsService.Test/TestFiles/permissions-test-file-v1.0_ver1.json
+++ b/PermissionsService.Test/TestFiles/permissions-test-file-v1.0_ver1.json
@@ -1,8 +1,7 @@
 {
   "ApiPermissions": {
-    "/security/alerts/{id}": [
-      {
-        "HttpVerb": "GET",
+    "/security/alerts/{id}": {
+      "GET": {
         "DelegatedWork": [
           "SecurityEvents.Read.All",
           "SecurityEvents.ReadWrite.All"
@@ -13,8 +12,7 @@
           "SecurityEvents.ReadWrite.All"
         ]
       },
-      {
-        "HttpVerb": "PATCH",
+      "PATCH": {
         "DelegatedWork": [
           "SecurityEvents.ReadWrite.All"
         ],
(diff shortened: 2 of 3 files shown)


GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -397,13 +397,22 @@ private bool RefreshPermissionsTables()
...
 
-                JArray resultValue = new JArray();
-                resultValue = (JArray)_scopesListTable[int.Parse(resultMatch.Key)];
-
-                var scopes = resultValue.FirstOrDefault(x => x.Value<string>("HttpVerb") == method)?
-                    .SelectToken(scopeType)?
-                    .Select(s => (string)s)
-                    .ToArray();
+                string[] scopes = null;
+                var resultValue = _scopesListTable[int.Parse(resultMatch.Key)] as JToken;
```
baywet, line 401: 1. the result might be null, you might want to check before the foreach
2. parse could throw an exception in case the Key is not an int, you might want to defend against that too.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -397,13 +397,22 @@ private bool RefreshPermissionsTables()
...
+                foreach (JProperty httpVerb in resultValue)
+                {
+                    if (httpVerb.Name.Equals(method, StringComparison.OrdinalIgnoreCase))
```
baywet, line 404: could be collapsed in the foreach (Where(x => ...))

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -397,13 +397,22 @@ private bool RefreshPermissionsTables()
...
+                    {
+                        foreach (JProperty scopeCategory in httpVerb.Value)
```
baywet, line 406: could be collapsed in the parent foreach (SelectMany)

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -397,13 +397,22 @@ private bool RefreshPermissionsTables()
...
+                        {
+                            if (scopeCategory.Name.Equals(scopeType, StringComparison.OrdinalIgnoreCase))
```
baywet, line 408: could be collapsed in the parent foreach (another where)

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -397,13 +397,22 @@ private bool RefreshPermissionsTables()
...
+                            {
+                                scopes = scopeCategory.Value?.ToObject<string[]>();
```
baywet, line 410: could be collapsed in the linq I'm describing and both for each removed.
```CSharp
scopes = resultValue.Where(x => method.Equals(x.Name, StringComparison.OrdinalIgnoreCase))
.SelectMany(x => x.Value)
.FirstOrDefault(x => scopeType.Equals(x.Name, StringComparison.OrdinalIgnoreCase))
?.Value?.ToObject<string[]>();
```

baywet (approved):
I'm leaving comments for improvements I didn't catch in the initial review. This doesn't need to be done now but can be done in a subsequent PR.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: 1. You mean the `resultMatch` variable? Nullability is being checked for here: https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/a8eed5b607d10c84f96720404a1f303cfddf35c8/GraphExplorerPermissionsService/Services/PermissionsStore.cs#L391-L398

2. This is covered here: https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/a8eed5b607d10c84f96720404a1f303cfddf35c8/GraphExplorerPermissionsService/Services/PermissionsStore.cs#L123-L126

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: I can add these in in a different PR before this one gets merged.

irvinesunday: Unfortunately, the suggested code is broken:

image

"`JToken` does not contain a definition for 'Name'... "

GraphExplorerPermissionsService/Services/PermissionsStore.cs
baywet: 1. I meant the resultValue
2. Sure, but a string can be empty, contain "abc" and int.Parse Is going to fail on both values.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
baywet: try adding `OfType<JProperty>()` between resultValue, and the Where. You have an implicit cast happening in the foreach.

baywet: > Note: you might have to add one after the select many as well

irvinesunday: Okay, these will def. go into separate PR in a separate workflow 😄

Issue #635, opened 2021-07-12
Not all scopes from the permissions mapping files are returned when Get all permissions are requested


irvinesunday:
**Describe the bug**
On this example request: `/permissions?scopeType=Application` we would expect all the scopes under `Application` permission types to be returned.

**To Reproduce**
- Run: https://graphexplorerapi.azurewebsites.net/permissions?requestUrl=/teams/{id}/channels&scopeType=Application&method=POST
- Search for the scope: `Channel.Create.Group`
- Run: https://graphexplorerapi.azurewebsites.net/permissions?scopeType=Application
- Search for the scope: `Channel.Create.Group`

**Expected behavior**
We expect to find the scope `Channel.Create.Group` in both searches above.

**Additional context**
The gap is because DevX API services the fetch-all permissions by returning all the scopes (and their descriptions) from the Service Principal and not from the [permissions mapping](https://raw.githubusercontent.com/microsoftgraph/microsoft-graph-devx-content/dev/permissions/permissions-v1.0.json) files. The solution is to pick all the unique scopes from the map files and then add the ones from the [Service Principal](https://raw.githubusercontent.com/microsoftgraph/microsoft-graph-devx-content/dev/permissions/permissions-descriptions.json). 
 
[AB#10188](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10188)

Issue #633, opened 2021-07-09
Permissions returned are repeated


thewahome:
When I request for permissions for the `/me` endpoint, this is what I see

![MicrosoftTeams-image (4)](https://user-images.githubusercontent.com/58787602/125070817-23de1900-e0c1-11eb-8de4-6c38bacbf494.png)

I should get only one copy per permission 
[AB#10172](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10172)

PR #636, opened 2021-07-12
Fix: Get all distinct permissions

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/635
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/633

## Overview

Gets all distinct scopes from the permissions mapping files and from the Service Principal descriptions when servicing fetch-all permissions requests from both the [permissions-mapping](https://raw.githubusercontent.com/microsoftgraph/microsoft-graph-devx-content/dev/permissions/permissions-v1.0.json) and the [Service Principal](https://raw.githubusercontent.com/microsoftgraph/microsoft-graph-devx-content/dev/permissions/permissions-descriptions.json) descriptions.

### Notes

Resolves issues raised in this PR as well: https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/628
@baywet I felt this syntax provides better readability and contextualization:
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/9835dbb0e5b9410e3db8b9178124d443bd4fbe7b/GraphExplorerPermissionsService/Services/PermissionsStore.cs#L407-L411

## Testing Instructions

- Run: https://localhost:44399/permissions?scopeType=Application
- Search for: `Channel.Create.Group`
- You should be able to see this sample query from the list.
- Run: https://localhost:44399/permissions?requesturl=/me&method=GET
- Search for: `Directory.AccessAsUser.All`
- There should only be one instance of this.

### DIFF
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -349,7 +349,11 @@ namespace GraphExplorerPermissionsService
                                              SeverityLevel.Information,
                                              _permissionsTraceProperties);
 
-                var scopes = new List<string>();
+                var scopes = new List<string>(from JToken operations in _scopesListTable.Values
+                                              from JProperty httpVerb in operations
+                                              from JProperty scopeCategory in httpVerb.Value
+                                              where scopeCategory.Name.Equals(scopeType, StringComparison.OrdinalIgnoreCase)
+                                              from scopeCategory.Value?.ToObject<List<string>>());
                 foreach (var scopeCategory in from JToken operations in _scopesListTable.Values
                                               from JProperty httpVerb in operations
                                               from JProperty scopeCategory in httpVerb.Value
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -506,7 +506,7 @@ namespace GraphExplorerPermissionsService
             foreach (var scope in scopes)
             {
                 ScopeInformation scopeInfo = null;
-                if (scopesInformationDictionary[key].ContainsKey(scope))
+scopesInformationDictionary[key].TryGetValue(scope, out var scopeInfo);
                 {
                     scopeInfo = scopesInformationDictionary[key][scope];
                 }
--- a/GraphExplorerPermissionsService/Services/PermissionsStore.cs
+++ b/GraphExplorerPermissionsService/Services/PermissionsStore.cs
@@ -269,31 +269,31 @@ namespace GraphExplorerPermissionsService
 
             ScopesInformationList scopesInformationList = JsonConvert.DeserializeObject<ScopesInformationList>(scopesInfoJson);
 
-            var _delegatedScopesInfoTable = new Dictionary<string, ScopeInformation>();
-            var _applicationScopesInfoTable = new Dictionary<string, ScopeInformation>();
+            var delegatedScopesInfoTable = new Dictionary<string, ScopeInformation>();
+            var applicationScopesInfoTable = new Dictionary<string, ScopeInformation>();
 
             foreach (ScopeInformation delegatedScopeInfo in scopesInformationList.DelegatedScopesList)
             {
-                _delegatedScopesInfoTable.Add(delegatedScopeInfo.ScopeName, delegatedScopeInfo);
+                delegatedScopesInfoTable.Add(delegatedScopeInfo.ScopeName, delegatedScopeInfo);
             }
 
             foreach (ScopeInformation applicationScopeInfo in scopesInformationList.ApplicationScopesList)
             {
-                _applicationScopesInfoTable.Add(applicationScopeInfo.ScopeName, applicationScopeInfo);
+                applicationScopesInfoTable.Add(applicationScopeInfo.ScopeName, applicationScopeInfo);
             }
 
             _permissionsTraceProperties.Add(UtilityConstants.TelemetryPropertyKey_SanitizeIgnore, nameof(PermissionsStore));
             _telemetryClient?.TrackTrace("Finished creating the scopes information tables. " +
-                                         $"Delegated scopes count: {_delegatedScopesInfoTable.Count}. " +
-                                         $"Application scopes count: {_applicationScopesInfoTable.Count}",
+                                         $"Delegated scopes count: {delegatedScopesInfoTable.Count}. " +
+                                         $"Application scopes count: {applicationScopesInfoTable.Count}",
                                          SeverityLevel.Information,
                                          _permissionsTraceProperties);
             _permissionsTraceProperties.Remove(UtilityConstants.TelemetryPropertyKey_SanitizeIgnore);
 
             return new Dictionary<string, IDictionary<string, ScopeInformation>>
             {
-                { Delegated, _delegatedScopesInfoTable },
-                { Application, _applicationScopesInfoTable }
+                { Delegated, delegatedScopesInfoTable },
+                { Application, applicationScopesInfoTable }
             };
         }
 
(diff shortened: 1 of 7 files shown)


GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -341,40 +341,32 @@ private bool RefreshPermissionsTables()
...
             if (string.IsNullOrEmpty(requestUrl))  // fetch all permissions
             {
                 _telemetryClient?.TrackTrace("Fetching all permissions",
                                              SeverityLevel.Information,
                                              _permissionsTraceProperties);
 
-                List<ScopeInformation> scopesListInfo = new List<ScopeInformation>();
-
-                if (scopeType.Contains(Delegated))
+                var scopes = new List<string>();
```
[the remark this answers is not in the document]
baywet: you could even append the distinct at the end of the statement here

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -388,81 +380,60 @@ private bool RefreshPermissionsTables()
...
 
-                List<ScopeInformation> scopesList = new List<ScopeInformation>();
+                List<string> scopes = null;
+                foreach (JProperty scopeCategory in from JProperty httpVerb in resultValue
+                                                    where httpVerb.Name.Equals(method, StringComparison.OrdinalIgnoreCase)
+                                                    from JProperty scopeCategory in httpVerb.Value
+                                                    where scopeCategory.Name.Equals(scopeType, StringComparison.OrdinalIgnoreCase)
+                                                    select scopeCategory)
+                {
+                    scopes = scopeCategory.Value?
```
baywet, line 413: First or default instead of foreach

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -529,6 +500,68 @@ private static string CleanRequestUrl(string requestUrl)
...
+            {
+                throw new ArgumentException($"'{nameof(scopeType)}' cannot be null or empty.", nameof(scopeType));
+            }
+
+            var scopesInfo = new List<ScopeInformation>();
+            foreach (var scope in scopes)
+            {
+                ScopeInformation scopeInfo = null;
+
+                if (scopeType.Contains(Delegated))
```
baywet, line 534: doing the comparison at each loop is not very optimal, get it out into a boolean var

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -269,31 +269,31 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
 
             ScopesInformationList scopesInformationList = JsonConvert.DeserializeObject<ScopesInformationList>(scopesInfoJson);
 
-            var _delegatedScopesInfoTable = new Dictionary<string, ScopeInformation>();
-            var _applicationScopesInfoTable = new Dictionary<string, ScopeInformation>();
+            var delegatedScopesInfoTable = new Dictionary<string, ScopeInformation>();
+            var applicationScopesInfoTable = new Dictionary<string, ScopeInformation>();
```
baywet, line 273: same comment as previous for the initialization

baywet (changes requested):
Thanks for following up on this one.
IMHO the method syntax is vastly superior to the query syntax as some operations (distinct, firstordefault and many others) are simply not available with the query syntax last time I checked.
Plus the method syntax gives you a fluent style that "reads like english" except for the lambdas and gets close to functional style programming.

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: True. 
I've gone for the option of using a single variable for the key: https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/181dcd4479d06b94bddbaf6994f8de40396e2218/GraphExplorerPermissionsService/Services/PermissionsStore.cs#L504

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -529,6 +500,68 @@ private static string CleanRequestUrl(string requestUrl)
...
+            }
+
+            var scopesInfo = new List<ScopeInformation>();
+            foreach (var scope in scopes)
+            {
+                ScopeInformation scopeInfo = null;
+
+                if (scopeType.Contains(Delegated))
+                {
+                    if (scopesInformationDictionary[Delegated].ContainsKey(scope))
```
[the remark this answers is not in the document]
irvinesunday: This works too in breaking up the duplication
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/181dcd4479d06b94bddbaf6994f8de40396e2218/GraphExplorerPermissionsService/Services/PermissionsStore.cs#L509-L512

irvinesunday:
> Thanks for following up on this one.
> IMHO the method syntax is vastly superior to the query syntax as some operations (distinct, firstordefault and many others) are simply not available with the query syntax last time I checked.
> Plus the method syntax gives you a fluent style that "reads like english" except for the lambdas and gets close to functional style programming.

Thank you for your feedback; much is learnt from these. 💪

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -529,11 +475,63 @@ private static string CleanRequestUrl(string requestUrl)
...
+            {
+                ScopeInformation scopeInfo = null;
+                if (scopesInformationDictionary[key].ContainsKey(scope))
+                {
+                    scopeInfo = scopesInformationDictionary[key][scope];
+                }
+
+                if (scopeInfo is not null)
+                {
+                    scopesInfo.Add(scopeInfo);
```
baywet, line 516:
```suggestion
                    scopesInfo.Add(scopeInfo ?? new() {
                        ScopeName = scope
                    });
```

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -529,11 +475,63 @@ private static string CleanRequestUrl(string requestUrl)
...
+            var scopesInfo = new List<ScopeInformation>();
+            foreach (var scope in scopes)
+            {
+                ScopeInformation scopeInfo = null;
+                if (scopesInformationDictionary[key].ContainsKey(scope))
+                {
+                    scopeInfo = scopesInformationDictionary[key][scope];
+                }
+
+                if (scopeInfo is not null)
```
baywet, line 514: remove if/else statement

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -529,11 +475,63 @@ private static string CleanRequestUrl(string requestUrl)
...
+            {
+                throw new ArgumentException($"'{nameof(scopeType)}' cannot be null or empty.", nameof(scopeType));
+            }
+
+            string key = scopeType.Contains(Delegated) ? Delegated : Application;
+            var scopesInfo = new List<ScopeInformation>();
+            foreach (var scope in scopes)
+            {
+                ScopeInformation scopeInfo = null;
+                if (scopesInformationDictionary[key].ContainsKey(scope))
```
baywet, line 509:
```suggestion
scopesInformationDictionary[key].TryGetValue(scope, out var scopeInfo);
```

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -529,11 +475,63 @@ private static string CleanRequestUrl(string requestUrl)
...
+            if (string.IsNullOrEmpty(scopeType))
+            {
+                throw new ArgumentException($"'{nameof(scopeType)}' cannot be null or empty.", nameof(scopeType));
+            }
+
+            string key = scopeType.Contains(Delegated) ? Delegated : Application;
+            var scopesInfo = new List<ScopeInformation>();
+            foreach (var scope in scopes)
+            {
+                ScopeInformation scopeInfo = null;
```
baywet, line 508: remove variable declaration

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -529,11 +475,63 @@ private static string CleanRequestUrl(string requestUrl)
...
+                if (scopesInformationDictionary[key].ContainsKey(scope))
+                {
```
baywet, line 510: remove block

GraphExplorerPermissionsService/Services/PermissionsStore.cs
```diff
@@ -529,11 +475,63 @@ private static string CleanRequestUrl(string requestUrl)
...
+            }
+
+            if (string.IsNullOrEmpty(scopeType))
+            {
+                throw new ArgumentException($"'{nameof(scopeType)}' cannot be null or empty.", nameof(scopeType));
+            }
+
+            string key = scopeType.Contains(Delegated) ? Delegated : Application;
+            var scopesInfo = new List<ScopeInformation>();
+            foreach (var scope in scopes)
```
baywet, line 506: after applying all the changes I've suggested below, you'll find the foreach body is much simpler and easier to fit in a Select statement.
So I'd suggest
```CSharp
return scopes.Select(x => ....).ToList();
```
... being the updated body of the foreach.

baywet (changes requested):
thanks for taking in all the changes!
Now that we've simplified a lot of the code, there's another opportunity to simplify a last section greatly.

irvinesunday:
> thanks for taking in all the changes!
> Now that we've simplified a lot of the code, there's another opportunity to simplify a last section greatly.

You are quite the minimalist 😄

GraphExplorerPermissionsService/Services/PermissionsStore.cs
irvinesunday: All is fine and dandy now. But I have a question about switching up the `foreach` for the `Select`. What gains are we getting from this? IMO the `foreach` gives an easy-to-debug option if someone wants to _step into_ the code.

baywet: strictly from a performance standpoint the approach has multiple benefits:
- it avoids assigning and growing multiple times a list, the tolist will create a list with the right number of memory spaces
- select can take advantage of vectorialization (SI MD) and other hardware instructions benefits on some platforms, improving performance with large datasets (although I doubt this will make a measurable change here)
- loops instructions are not necessarily aligned by default, leading to CPU "overhead" https://devblogs.microsoft.com/dotnet/loop-alignment-in-net-6/


from a semantic code approach: foreach to me translate into an "action" (for each of those items, I'm going to change/apply something), but select translate into data mapping (I want to get this subset/facet of that existing dataset). This makes it easier to parse mentally as your brain is already in the mode "we're mapping data" and not in the mode "we're performing an action, but wait, we're actually mapping data".

baywet: another advantage I forgot to mention from a performance standpoint which partially applies here: let's assume we return IEnumerable instead of list, and we don't call toList after the select: this would enable deferred execution of the select/streaming of data. i.e.: the selection will only execute whenever a caller tries to enumerate the data. This is particularly useful on large datasets/whenever multiple transformations need to be applied (based on conditions on the data set (order, select, distinct,...)

irvinesunday: I hear you

baywet (approved):

PR #690, opened 2021-09-08
Update snippets url

MaggieKimani1:
Following the recent project renaming, the `/graphexplorersnippets` endpoint on Graph Explorer was failing.
This PR adds a `/graphexplorersnippets` route to the controller to fix this issue


irvinesunday (approved):

baywet:
Alternatively, you could have added a redirect middleware to support both urls until graph explorer is updated. https://docs.microsoft.com/en-us/aspnet/core/fundamentals/url-rewriting?view=aspnetcore-5.0

PR #694, opened 2021-09-08
Revert Serilog.Sinks.Console lib. version from 4.0.0 to 3.1.1

irvinesunday:
The latest version of `Serilog.Sinks.Console` `v4.0.0` breaks the AzDo build pipeline.

See the below snapshot of the logged build failure:
image

Full log details can be found here: https://o365exchange.visualstudio.com/2f15a0d3-3217-4778-8334-4e61181c4029/_apis/build/builds/6269887/logs/20


MaggieKimani1 (approved):

baywet:
Looking at the error, this environment variable is set by ADO, I don't think it has anything to do with serilog but rather with the fact the build had too many commits and ADO is not trimming like it should for some reason.
https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables-devops-services

I believe we should revert this pr

PR #696, opened 2021-09-08
Pipeline Failure

georgend:


MaggieKimani1 (approved):

pipelines/release.yml
```diff
@@ -35,27 +35,43 @@ steps:
 
 - task: VSBuild@1
```
baywet, line 36: while we're at it replace this by dotnet build and the nuget ones by dotnet restor

baywet (changes requested):

pipelines/release.yml
georgend: We have migrated to dotnetcore fully and a linux build agent. Thanks for the pointer.

pipelines/release.yml
```diff
@@ -26,36 +26,43 @@ steps:
...
-    solution: '$(solution)'
-    msbuildArgs: '/p:DeployOnBuild=true /p:WebPublishMethod=Package /p:PackageAsSingleFile=true /p:SkipInvalidConfigurations=true /p:DesktopBuildPackageLocation="$(build.artifactStagingDirectory)\WebApp.zip" /p:DeployIisAppPath="Default Web Site"'
-    platform: '$(buildPlatform)'
-    configuration: '$(buildConfiguration)'
+    command: 'publish'
+    publishWebProjects: true
+    zipAfterPublish: true
+    arguments: '--configuration $(buildConfiguration) --nologo --output $(Build.ArtifactStagingDirectory) '
 
 - task: YodLabs.O365PostMessage.O365PostMessageBuild.O365PostMessageBuild@0
```
baywet, line 67: I believe this task only works on windows. I also think we can remove it since now ADO has wekbooks for teams that are very specific and would allow us to subscribe to the failures of that pipeline without having to have a task for that.

pipelines/release.yml
```diff
@@ -26,36 +26,43 @@ steps:
...
-    testAssemblyVer2: |
-      **/net5.0/**.Test.dll
-      !**/*TestAdapter.dll
-      !**/obj/**
-    diagnosticsEnabled: true
+    testResultsFormat: xUnit
+    testResultsFiles: '**/*.trx' 
+    codeCoverageTool: 'cobertura'
+    summaryFileLocation: '$(System.DefaultWorkingDirectory)/**/coverage.cobertura.xml'
+    failTaskOnFailedTests: false # Optional
```
baywet, line 56:
```suggestion
    failTaskOnFailedTests: true # Optional
```
We most likely want to know when a test starts failing

georgend: We want the test results to be published even when tests fail.
From the [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=trx%2Cyaml):-
image

pipelines/release.yml
georgend: Remoed.

pipelines/release.yml
baywet: then you need to set this task to always run (default being only if previous task succeeds). That's because dotnet test will fail if a unit test fails.

georgend: We removed the params and defaulted to default behavior, given your points.

baywet: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml to be specific:
`condition: succeededOrFailed()` on the publish test results task

baywet (approved):

PR #712, opened 2021-09-22
Merges dev into master

irvinesunday:
## Overview

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description


MaggieKimani1 (approved):

baywet (commented):
you're getting a lot of changes that were already merged with #684
I've noticed this is a recurring pattern with the pull request you make and I'm worried at some point it'll lead to conflicts. Can you double check how your local repo is configured and whether it's tracking additional head changes for some reason please?

irvinesunday:
> you're getting a lot of changes that were already merged with #684
> I've noticed this is a recurring pattern with the pull request you make and I'm worried at some point it'll lead to conflicts. Can you double check how your local repo is configured and whether it's tracking additional head changes for some reason please?

The issue is not just localized to the PRs I create, but all PRs from the `dev` branch into `master`.

See other references: 
- https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/695
- https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/691
- https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/684

baywet:
right, thanks for the precision, I had missed this aspect.
I'm curious to how you create the dev to master PRs @irvinesunday and @MaggieKimani1 ?

And to sync locally, what tool do you use? What's your workflow?

MaggieKimani1:
@baywet I use GitKraken for my Git workflow

baywet:
I wonder whether you could be committing locally before pulling? Which would cause a commit, pull, local merge, push and bring the history like we're seeing. This is a common "bad practice" I've seen many times.

irvinesunday:
I usually create PRs from `dev` into `master` directly from the GitHub repo UI. I've noticed this unusual pattern happen only with this repo though.

zengin:
I think this is happening because `dev` and `master` don't share the same history despite having the same set of files.

When the two branches are compared, the comparison is done based on how many commits one branch is ahead of the other, not based on the actual file comparison.

It is the difference between
1. `git diff master...dev` (commit history comparison) and
2. `git diff master..dev` (actual file comparison)

Note the 3 vs 2 dots.

Creating a PR in GitHub (and in other review tools like Azure DevOps) is always a 3 dotted action, in other words it takes you to this link with 3 dots: https://github.com/microsoftgraph/microsoft-graph-devx-api/compare/master...dev. There are reasons for this but the gist is that a PR is a proposal of commits since the time two branches diverged, irrespective of whether the same changes are already merged by different commit hashes into the target branch.

A file based comparison can also be done on the web like the following, which shows 0 diff as the files are now in sync although history is not. Of course, 0 diff holds true as long as the link is clicked before `dev` gets more changes :).
https://github.com/microsoftgraph/microsoft-graph-devx-api/compare/master..dev

 One way to avoid this problem is
- not to squash merge from `dev` to `master` and vice versa so that the commits are the same (i.e. commit hashes are the same so that git knows how to compare these 2 different sets of histories)
- always have a forward integration merge to sync `master` back to `dev` with the merge commit.

baywet:
aaaaah thanks @zengin for shedding some light over this repo, I had not thought about a diverging commit history for some reason.
@irvinesunday can you run the following and force push to dev please? this way we'll stop having this behavior with that repo.

PR #709, opened 2021-09-21
[Feature] Add new path to support Resource Explorer

irvinesunday:
This PR: 
- Creates an additional path on the `OpenApiController` --> `/openapi/tree` for surfacing the `OpenApiUrlTreeNode`. This will be used by Graph Explorer for the new Resource Explorer feature.
- Updates the OpenAPI description to describe the above new path.
- Adds support for serializing the  `OpenApiUrlTreeNode` into JSON. The output result will look like this: https://microsoft.sharepoint-df.com/:u:/t/GraphTooling/EaaJlJIk21xIpI8_pbZQim8Bjy16CJgadxTc8G7JD0QcJg
The prettier version: https://microsoft.sharepoint-df.com/:u:/t/GraphTooling/ERL1J2BviFRBnLqbgIzjq90Bf3eryzE8sM-fFU-C3zY0Ww?e=A8qURb
- Refactors the code that is responsible for creating new or attaching on to an `OpenApiUrlTreeNode` into a common reusable method.
- Adds tests to validate the above capabilities.

PR #713, opened 2021-09-23
[New feature bug fix] Avoids caching OpenAPIUrlTreeNode document

irvinesunday:
This PR:

- Attempts to fix a bug that was surfaced during staging tests of the new Resource Explorer endpoint feature. The `OpenApiService` was not correctly returning the expected `OpenApiUrlTreeNode` data for a requested version/label, e.g. 
  - These calls: 
  https://graphexplorerapi-staging.azurewebsites.net/openapi/tree?graphVersions=v1.0 or 
  https://graphexplorerapi-staging.azurewebsites.net/openapi/tree?graphVersions=beta 
fetches the `OpenApiUrlTreeNode` document for both `v1.0` and `beta` versions instead of the specific requested version, `v1.0` or `beta`. This is because, according to the implementation logic, we were improperly caching the `OpenApiUrlTreeNode` document that gets created. 
  - The proposed implementation in this PR is that the  `OpenApiUrlTreeNode` gets created every time it is requested, without storing it in a cache. The `CreateOpenApiUrlTreeNode` method accepts a `Collection` of `OpenApiDocument` and simply creates an `OpenApiUrlTreeNode` without caching the result. This will help in ensuring that the tree nodes do not go stale over time. Instead, the tree nodes will always be a representation of the cached versions of the `OpenApiDocument`.
- Actions user feedback that proposes calling the path `/openapi/tree` without providing any query params should default to fetching the `OpenApiUrlTreeNode` containing both `v1.0` and `beta` paths, and not the initial default `v1.0` paths.

Related to: https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/709

### DIFF
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -160,7 +160,7 @@ namespace GraphWebApi.Controllers
 
         [Route("openapi/tree")]
         [HttpGet]
-        public async Task<IActionResult> Get([FromQuery] string graphVersions,
+        public async Task<IActionResult> Get([FromQuery] string graphVersions = "*",
                                              [FromQuery] bool forceRefresh = false)
         {
             graphVersions = string.IsNullOrEmpty(graphVersions) ? "*" : graphVersions.ToLower();
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -163,6 +163,11 @@ namespace GraphWebApi.Controllers
         public async Task<IActionResult> Get([FromQuery] string graphVersions = "*",
                                              [FromQuery] bool forceRefresh = false)
         {
+            if (string.IsNullOrEmpty(graphVersions))
+            {
+                throw new InvalidOperationException($"{nameof(graphVersions)} parameter has an invalid value.");
+            }
+
             HashSet<string> graphVersionsList = new();
             if (graphVersions == "*")
             {
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -173,7 +173,7 @@ namespace GraphWebApi.Controllers
             }
             else
             {
-                graphVersionsList.Add(graphVersions);
+                graphVersionsList.Add(graphVersions.ToLower());
             }
 
             var sources = new Dictionary<string, OpenApiDocument>();
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -174,10 +174,10 @@ namespace GraphWebApi.Controllers
             }
             else
             {
-                // Either v1.0 or beta explicitly defined, or both
-                graphVersionsList = graphVersions.Split(',', StringSplitOptions.TrimEntries).ToList();
+                graphVersionsList.Add(graphVersions);
             }
 
+            var sources = new Dictionary<string, OpenApiDocument>();
             foreach (var graphVersion in graphVersionsList)
             {
                 var graphUri = GetVersionUri(graphVersion);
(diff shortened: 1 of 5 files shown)


GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -161,11 +160,11 @@ public OpenApiController(IConfiguration configuration, IOpenApiService openApiSe
 
         [Route("openapi/tree")]
         [HttpGet]
-        public async Task<IActionResult> Get([FromQuery] string graphVersions = Constants.OpenApiConstants.GraphVersion_V1,
+        public async Task<IActionResult> Get([FromQuery] string graphVersions,
                                              [FromQuery] bool forceRefresh = false)
         {
+            graphVersions = string.IsNullOrEmpty(graphVersions) ? "*" : graphVersions.ToLower();
```
thewahome, line 166: Nice that the parameter is not required if someone wants both versions

irvinesunday: Thank you for your earlier suggestion 😉

GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -161,11 +160,11 @@ public OpenApiController(IConfiguration configuration, IOpenApiService openApiSe
 
         [Route("openapi/tree")]
         [HttpGet]
-        public async Task<IActionResult> Get([FromQuery] string graphVersions = Constants.OpenApiConstants.GraphVersion_V1,
+        public async Task<IActionResult> Get([FromQuery] string graphVersions,
```
baywet, line 163:
```suggestion
        public async Task<IActionResult> Get([FromQuery] string graphVersions = "*",
```

GraphWebApi/Controllers/OpenApiController.cs
baywet:
```suggestion
```

GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -174,25 +173,24 @@ public OpenApiController(IConfiguration configuration, IOpenApiService openApiSe
             }
             else
             {
-                // Either v1.0 or beta explicitly defined, or both
-                graphVersionsList = graphVersions.Split(',', StringSplitOptions.TrimEntries).ToList();
+                graphVersionsList.Add(graphVersions);
```
baywet, line 176:
```suggestion
                graphVersionsList.Add(graphVersions.ToLower());
```

GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -161,11 +160,11 @@ public OpenApiController(IConfiguration configuration, IOpenApiService openApiSe
 
         [Route("openapi/tree")]
         [HttpGet]
-        public async Task<IActionResult> Get([FromQuery] string graphVersions = Constants.OpenApiConstants.GraphVersion_V1,
+        public async Task<IActionResult> Get([FromQuery] string graphVersions,
                                              [FromQuery] bool forceRefresh = false)
         {
+            graphVersions = string.IsNullOrEmpty(graphVersions) ? "*" : graphVersions.ToLower();
             List<string> graphVersionsList = new();
```
baywet, line 167:
```suggestion
            HashSet<string> graphVersionsList = new();
```

baywet (changes requested):

GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -174,25 +172,24 @@ public OpenApiController(IConfiguration configuration, IOpenApiService openApiSe
             }
             else
```
baywet, line 173:
```suggestion
            else if(!string.IsNullOrEmpty(graphVersions))
```

GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -174,25 +172,24 @@ public OpenApiController(IConfiguration configuration, IOpenApiService openApiSe
...
             {
-                // Either v1.0 or beta explicitly defined, or both
-                graphVersionsList = graphVersions.Split(',', StringSplitOptions.TrimEntries).ToList();
+                graphVersionsList.Add(graphVersions.ToLower());
             }
```
baywet, line 176: else throw? or return a bad request?

baywet (changes requested):

GraphWebApi/Controllers/OpenApiController.cs
irvinesunday: I already made a similar change before reading your suggestion here 😄

baywet (approved):

PR #719, opened 2021-09-29
Fixes Code Smells

### LINKED ISSUE #707: Address Critical code smells

MaggieKimani1:
This PR proposes:
- Fixing some of the code smells surfaced by SonarCloud
- Removing deprecated authentication handler code from `Startup.cs` and the configs from `appsettings.json`
- Removes deprecated `SampleQueriesCategories` class
- Addresses compiler warnings

Fixes #707


GraphWebApi/Startup.cs
```diff
@@ -49,20 +49,6 @@ public Startup(IConfiguration configuration, IWebHostEnvironment hostingEnvironm
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddAuthentication(option =>
```
baywet, line 52: just double checking we're not using any auth here and the removal was intended, not just a mechanical removal from static analysis

MaggieKimani1: Yeah we're not using auth for any of our endpoints.
The initial implementation used it for the Create, update and delete operations for the samples endpoint but that functionality is already handled by the DevX Content repo, thus rendering auth for DevX API obsolote.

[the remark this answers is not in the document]
MaggieKimani1: Yes. Resolving..

baywet: thanks for confirming, I wanted to double check we were not pulling something that was actually needed. Nit: there's probably a bunch of usings you can clean after removing this.

TelemetryService/CustomPIIFilter.cs
```diff
@@ -69,11 +69,11 @@ public class CustomPIIFilter : ITelemetryProcessor
         private const string ODataFilterOperator = "$filter";
         private const string SecurityMask = "****";
 
-        public CustomPIIFilter(ITelemetryProcessor next, IServiceProvider serviceProvider)
```
zengin, line 72: I think you can choose to suppress the warning for this particular instance. `CustomPIIFilter` represents the acronym `PII` better.

baywet: Or alternatively go with a phonic spelling to respect conventions CustomPeaAyeAyeFilter 🤣 (don't listen to me)

baywet (approved):

andrueastman (approved):

PR #725, opened 2021-09-30
updates open api conventions to align snippets on kiota latest

baywet:
This PR reflects the updates in kiota done for open api conventions in that PR https://github.com/microsoft/kiota/pull/573

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/KiotaOpenApiSchemaExtensions.cs
+++ b/CodeSnippetsReflection.OpenAPI/KiotaOpenApiSchemaExtensions.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.OpenApi.Models;
 
-// THIS CLASS IS COPIED FROM KIOTA TO GET THE SAME NAMEING CONVENTIONS, WE SHOULD FIND A WAY TO MUTUALIZE THE CODE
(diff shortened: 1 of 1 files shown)


CodeSnippetsReflection.OpenAPI/KiotaOpenApiSchemaExtensions.cs
```diff
@@ -6,7 +6,7 @@
 // THIS CLASS IS COPIED FROM KIOTA TO GET THE SAME NAMEING CONVENTIONS, WE SHOULD FIND A WAY TO MUTUALIZE THE CODE
```
irvinesunday, line 6: [Nitpick] *NAMEING* typo

irvinesunday:
Should we add tests to this class?

irvinesunday:
Add license info.

CodeSnippetsReflection.OpenAPI/KiotaOpenApiSchemaExtensions.cs
baywet:
```suggestion
// THIS CLASS IS COPIED FROM KIOTA TO GET THE SAME NAMING CONVENTIONS, WE SHOULD FIND A WAY TO MUTUALIZE THE CODE
```

baywet:
@irvinesunday about the unit tests, this is already unit tested in kiota. I don't see value in duplicating the tests.
Hopefully at some point we can just take a dependency on kiota an avoid duplicating code, but so far the impact is limited IMHO.

irvinesunday (approved):

Issue #721, opened 2021-09-29
Use service name in place of workload when querying for changes in the changes endpoint


irvinesunday:
The [changelog.json](https://graphprodblobstorage.blob.core.windows.net/content/graph/en-us/json/changelog.json) defines changes by Service names, e.g.: 

![image](https://user-images.githubusercontent.com/40403681/139022829-9b435409-e7f9-497b-8d27-e634722cb32e.png)

We are currently using the _TargetWorkloadId_ to search for changes (in cases where the client provides a url in the query param, e.g. https://graphexplorerapi.azurewebsites.net/changes?requestUrl=/me/messages):
![image](https://user-images.githubusercontent.com/40403681/139023222-26a94ca3-4800-4aa6-b0f0-bab09cb39683.png)

This _TargetWorkloadId_ is obtained from Graph by appending $whatif as a query option to the provided request url.

We need to map the _TargetWorkloadId_ to their respective service names, which can then be used to query the changelog data. The available mapping for this is this, [workload-mapping.json](https://microsoft.sharepoint-df.com/:u:/t/GraphTooling/EfH2TzE_zoBKqBZae0Ar-nQBxsbPYVfrRFjXkaIUqeqVMQ?e=Ldo56y), which is maintained by the content team and specific workload owners.

PR #747, opened 2021-10-27
Fetches changelog data by service name

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/721

PR proposals:
- Fetches and caches the workload-mapping.json file from the Azure blob. This will be handled by the `ChangesStore.cs` class.
- Adds tests to validate the above.
- Changes the public facing parameter name from `workload` to `service`.
- Updates how we call the Graph proxy to retrieve the workload information. Graph proxy urls were changed, and we needed to make corresponding changes.
- Updates and additions to `appsettings.json` were done because of the above.
- Updates the  `ChangesService.cs` class to include filtering for service names in the changelog data using the Graph version provided.
- Uses the cached workload-mapping.json file to to infer the service names of workloads and use the retrieved name to query the changelog data source. 
  - In the case of changes request for a given requestUrl, e.g.: https://graphexplorerapi.azurewebsites.net/changes?requestUrl=/me/messages,  the url segments --> _me_ and _messages_ will be matched against the comma separated values of the `Target` property of a matched service name (that corresponds to the WorkloadId of the provided request url). Refer to this [comment](https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/721#issuecomment-952709386) for more details.
- Updates and adds tests to validate the above.
- Updates the return statement of the `ReadFromDocumentAsync()` method of the `HttpClientUtility.cs` class so as to be able to capture any error messages returned from the Http service through the `HttpResponseMessage` object, instead of the defined constant exception message.
- Moves the `Constant.cs` class defined in the `FileService.Test` project into the `MocktestUtility` project and rename it to `MockHttpConstants` as this holds data used by the `MockHttpMessageHandler.cs` class.


irvinesunday:
@baywet your approval is required to unblock this PR.

ChangesService/Services/ChangesStore.cs
```diff
@@ -36,9 +40,14 @@ public class ChangesStore : IChangesStore
...
         private readonly TelemetryClient _telemetryClient;
         private readonly IChangesService _changesService;
+        private readonly string _workloadMappingContainerName;
+        private readonly string _workloadMappingBlobName;
+        private const string WorkloadMappingContainerConfig = "BlobStorage:Containers:Changelog";
+        private const string WorkloadMappingBlobConfig = "BlobStorage:Blobs:WorkloadMapping";
 
         public ChangesStore(IConfiguration configuration, IMemoryCache changeLogCache, IChangesService changesService,
-                            IHttpClientUtility httpClientUtility, TelemetryClient telemetryClient = null)
+                            IHttpClientUtility httpClientUtility, IFileUtility fileUtility = null,
```
baywet, line 49: why have the file utility as optional if you defend against it being null down below, make it mandatory

baywet (changes requested):

ChangesService/Services/ChangesStore.cs
irvinesunday: Resolved

baywet (approved):

thewahome (commented):
The call to the `endpoint` URL is unnecessary with every request. This means that with every changelog request, the API has to make two calls.
1. to the endpoint URL to get a proxy URL
2. to the returned proxy URL to do the $whatif query

Consider doing this instead:
1. Place a currently working proxy URL (returned from a call the endpoint URL) into the cache.
2. Make all  $whatif calls directly to the cached URL
3. If you make the call to the proxy URL and it fails, make the call to the endpoint URL to get a new URL and then place this in the cache (back to step 1)
4. Repeat step 3 each time there is a change in the proxy URL

irvinesunday:
Merging 

> The call to the `endpoint` URL is unnecessary with every request. This means that with every changelog request, the API has to make two calls.
> 
> 1. to the endpoint URL to get a proxy URL
> 2. to the returned proxy URL to do the $whatif query
> 
> Consider doing this instead:
> 
> 1. Place a currently working proxy URL (returned from a call the endpoint URL) into the cache.
> 2. Make all  $whatif calls directly to the cached URL
> 3. If you make the call to the proxy URL and it fails, make the call to the endpoint URL to get a new URL and then place this in the cache (back to step 1)
> 4. Repeat step 3 each time there is a change in the proxy URL

@thewahome calls to the proxy can fail for other reasons besides a changed proxy url; do you have the relevant error code and/or message that would indicate that the given proxy is no longer valid?

baywet (approved):

Issue #728, opened 2021-10-06
Invalid `operationIds` for certain OData functions under `reports`.


peombwa:
**Describe the bug**
Certain operationIds for OData functions under `reports` entitySet have a weird suffix to them e.g., `reports_getSharePointSiteUsageDetail-204b`.

**To Reproduce**
- Go to https://graphexplorerapi.azurewebsites.net/openapi?tags=reports.Functions&openApiVersion=3&graphVersion=v1.0&format=yaml&style=PowerShell
- Inspect the returned OpenAPI document.
![image](https://user-images.githubusercontent.com/7061532/136273472-11606858-bc68-4eb8-ac6c-4e7097bc41d8.png)

**Expected behavior**
OperationIds should match the function names in the OData metadata.
![image](https://user-images.githubusercontent.com/7061532/136273983-5cbb0f73-c580-4a8f-a425-3b83e690c679.png)

darrelmiller:
Possible suggestion: Could we add WithDate and WithPeriod to the end of the OperationId?  i.e. we use the set of parameter names as a unique suffix.

irvinesunday:
@peombwa the suffix after the `operationId` is a hash value of the function name. This is appended to the `operationId`s of two or more overloaded `Functions` so as to avoid duplicates. 

@darrelmiller some [functions](https://graphexplorerapi.azurewebsites.net/openapi?operationIds=drive.list.items.listItem.getActivitiesByInterval-96b0&openApiVersion=3&graphVersion=v1.0&format=yaml&style=PowerShell) are parameter-less, so the option of using parameter name might not be feasible.
![image](https://user-images.githubusercontent.com/40403681/142846073-3b93fe3d-8919-42d0-9787-f3597fdcd58c.png)

Does PowerShell care much about this suffix or the naming of this suffix. If not, we can just remove these suffixes in the DevX API when processing the subset OpenAPI documents.

darrelmiller:
@irvinesunday If the function is parameterless then there will not be a name clash.  OData requires the parameters to disambiguate between function names.  Therefore if there are no parameters then no suffix is required.

peombwa:
@irvinesunday, is there a reason we need the hash to disambiguate duplicate `operationIds`?  Having `GET` operations that have duplicate `operationIds` and different parameter names is not an issue in PowerShell. AutoREST will generate an overload (parameterSet) if the `operationIds` are identical and the parameters are different.

[...]

PR #797, opened 2021-11-24
Removes hash suffix values from the operationIds of function paths

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/728

This PR:
- Uses regex to remove hash suffix values from the `operationId`s of certain _function_ paths for the `PowerShell` `style` option.
- Adds a mock _function_ path and a test to validate the above.

### DIFF
--- a/OpenAPIService/PowershellFormatter.cs
+++ b/OpenAPIService/PowershellFormatter.cs
@@ -116,16 +116,8 @@ namespace OpenAPIService
         }
 
         /// <summary>
-        /// Resolves action and function OperationIds by reverting their signatures
-        /// to how they were being defined in package Microsoft.OpenApi.OData ver. 1.0.6 and below.
+        /// Resolves the OperationIds of action and function paths.
         /// </summary>
-        /// <remarks>
-        /// This is to prevent change of cmdlet names already defined using the previous version's format.
-        /// </remarks>
-        /// <example>
-        /// Default OperationId --> communications.calls.call_keepAlive
-        /// Resolved OperationId --> communications.calls_keepAlive
-        /// </example>
         /// <param name="operation">The target OpenAPI operation.</param>
         /// <returns>The resolved OperationId.</returns>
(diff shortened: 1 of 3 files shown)


OpenAPIService/PowershellFormatter.cs
```diff
@@ -150,7 +148,16 @@ private static string ResolveActionFunctionOperationId(OpenApiOperation operatio
                 }
             }
 
-            return string.Join(".", segments);
+            var updatedOperationId = string.Join(".", segments);
+
+            // Remove hash suffix values from OperationIds of function paths.
+            // For example,
+            // Default OperationId --> reports_getEmailActivityUserDetail-fe32
+            // Resolved OperationId --> reports_getEmailActivityUserDetail
```
baywet, line 156: it sounds to me like the hash was added in case of multiple actions with same name different parameter set to avoid operation id conflicts (they must be unique). I think with this change we are at risk of running into collisions.
For Kiota, for the namespace names (which rely on the path, same data source as the operation id in the conversion lib), we chose the convention of FunctionNameWithParameter1WithParameter2 etc... (only including the parameters that are in path) to avoid collisions

irvinesunday: I should have being more explicit in my description, sorry. This particular change only affects the `PowerShell` `style` option
All other `style` options remain unaffected.

baywet: thanks for the precision, won't that still lead to conflicting cmdlets?

irvinesunday: Check @peombwa's [comment](https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/728#issuecomment-975863954) on this with regard to how AutoREST handles this.

baywet: Thanks for the context!

baywet (approved):

peombwa (approved):

Issue #441, opened 2021-02-04
[Feature Request] Support regex-based filtering on operationIds query parameter


peombwa:
It would be nice to support a regex-based filtering on operationIds query parameter. This will make it easier for PS SDK and future SDKs to narrow down on a subset of APIs. This can be similar to what we currently have for tags.

e.g. 
This is currently not supported `https://graphexplorerapi.azurewebsites.net/openapi?operationIds=deviceManagement.reports*`.

darrelmiller:
@peombwa   Can you give an example of where you might want this?

millicentachieng:
This will be done in Hidi https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/802#issuecomment-983722081

PR #802, opened 2021-11-25
[Feature]: Support regex-based filtering on operationIds query parameter

danielmbaluka:
## Overview
Issue: https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/441

This PR adds the support for filtering Open API by operationIds query parameter.
 For example: GET `https://graphexplorerapi.azurewebsites.net/openapi?operationIds=deviceManagement.reports*`

### Demo
filter-by-operation-ids

### DIFF
--- a/OpenAPIService.Test/OpenAPIServiceShould.cs
+++ b/OpenAPIService.Test/OpenAPIServiceShould.cs
@@ -199,6 +199,28 @@ namespace OpenAPIService.Test
         }
 
         [Theory]
+        [InlineData("^users.", null, 3)]
+        [InlineData("users.user*", null, 2)]
+        [InlineData("^users.|^applications.", null, 5)]
+        [InlineData(null, "^users.", 3)]
+        [InlineData(null, "users.user*", 2)]
+        [InlineData(null, "^users.|^applications.", 5)]
+        public void SupportFilteringOpenApiDocumentByRegexOnOperationIdsAndTags(string operationIds, string tags, int pathCount)
+        {
+            // Act
+            var predicate = _openApiService.CreatePredicate(operationIds: operationIds,
+                                                          tags: tags,
+                                                          url: null,
+                                                          source: _graphMockSource,
+                                                          graphVersion: GraphVersion);
+
+            var subsetOpenApiDocument = _openApiService.CreateFilteredDocument(_graphMockSource, Title, GraphVersion, predicate);
+
+            // Assert
+            Assert.Equal(pathCount, subsetOpenApiDocument.Paths.Count);
+        }
+
+        [Theory]
         [InlineData(OpenApiStyle.Plain, "/users/{user-id}", OperationType.Get)]
         [InlineData(OpenApiStyle.Plain, "/users/12345", OperationType.Get)]
         [InlineData(OpenApiStyle.GEAutocomplete, "/users(user-id)messages(message-id)", OperationType.Get)]
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -160,13 +160,18 @@ namespace OpenAPIService
             Func<OpenApiOperation, bool> predicate;
             if (operationIds != null)
             {
-                if (operationIds == "*")
+                var operationIdsArray = operationIds.Split(',');
+                if (operationIds == "*") // Note: string literal "*" is not valid regex, so we are handling it separately
                 {
                     predicate = (o) => true;  // All operations
                 }
+                else if (operationIdsArray.Length == 1)
+                {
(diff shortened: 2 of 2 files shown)


OpenAPIService/OpenApiService.cs
```diff
@@ -160,13 +160,18 @@ public OpenApiDocument CreateFilteredDocument(OpenApiDocument source, string tit
...
             {
-                if (operationIds == "*")
+                var operationIdsArray = operationIds.Split(',');
+                if (operationIds == "*") // Note: string literal "*" is not valid regex, so we are handling it separately
                 {
                     predicate = (o) => true;  // All operations
                 }
+                else if (operationIdsArray.Length == 1)
+                {
+                    var regex = new Regex(operationIdsArray[0]);
```
baywet, line 170: we should probably surround this with a try catch in case the regex is invalid to return a sanitized exception.

baywet (changes requested):

OpenAPIService/OpenApiService.cs
irvinesunday: What's an example of an invalid regex in this case?

Ideally we should just return a `404` with a relevant message. Example,
image

OpenAPIService.Test/OpenAPIServiceShould.cs
```diff
@@ -198,6 +198,24 @@ public void ReturnOpenApiDocumentInCreateFilteredDocumentWhenValidArgumentsAreSp
             }
         }
 
+        [Theory]
+        [InlineData("users.user*", null)]
+        [InlineData(null, "users.user*")]
+        public void SupportFilteringOpenApiDocumentByRegexOnOperationIdsAndTags(string operationIds, string tags)
```
irvinesunday, line 204: Improvement suggestion:
```
 [Theory]
        [InlineData("^users.", null, 3)]
        [InlineData("^users.|^applications.", null, 5)]
        [InlineData(null, "^users.", 3)]
        [InlineData(null, "^users.user$", 2)]
        [InlineData(null, "^users.|^communications.|^applications.", 6)]
        public void SupportFilteringOpenApiDocumentByRegexOnOperationIdsAndTags(string operationIds, string tags,int pathCount)
        {
            // Act
            var predicate = _openApiService.CreatePredicate(operationIds: operationIds,
                                                          tags: tags,
                                                          url: null,
                                                          source: _graphMockSource,
                                                          graphVersion: GraphVersion);

            var subsetOpenApiDocument = _openApiService.CreateFilteredDocument(_graphMockSource, Title, GraphVersion, predicate);

            // Assert
            Assert.Equal(pathCount, subsetOpenApiDocument.Paths.Count);
        }
```

danielmbaluka: Thank you @irvinesunday, I have implemented the suggestion.

OpenAPIService/OpenApiService.cs
baywet: [[ would be an invalid regex for example

danielmbaluka: @irvinesunday here is an example of an invalid regex.
invalid-regex-example

darrelmiller:
We just finished moving slicing capabilities to Hidi.  The goal is to stop using devX for slicing.  This should have been implemented in the OpenAPI.NET slicing service, not in DevX API.

OpenAPIService/OpenApiService.cs
baywet: works for me, we're not leaking the stack trace and responding with a comprehensive error message.

baywet (approved):

irvinesunday:
> We just finished moving slicing capabilities to Hidi. The goal is to stop using devX for slicing. This should have been implemented in the OpenAPI.NET slicing service, not in DevX API. /cc @irvinesunday @adhiambovivian @MaggieKimani1

Hidi can now easily just port this to its codebase 😁. In the meantime, I think DevX API can continue exposing this new feature until a full switch to Hidi is realised.

irvinesunday (approved):

MaggieKimani1:
> > We just finished moving slicing capabilities to Hidi. The goal is to stop using devX for slicing. This should have been implemented in the OpenAPI.NET slicing service, not in DevX API. /cc @irvinesunday @adhiambovivian @MaggieKimani1
> 
> Hidi can now easily just port this to its codebase 😁. In the meantime, I think DevX API can continue exposing this new feature until a full switch to Hidi is realised.

In the meantime, I can import this functionality into Hidi. 😊

danielmbaluka:
This feature is not needed in DevX API for now. It will be ported to Hidi.

PR #806, opened 2021-12-01
- switches to feed for api doctor instead of local build

baywet:
This removes the apidoctor resource in the pipeline definition to use the nuget package instead.

A few limitations here because the nuget package is not published as a tool :
- cannot simply install it with `dotnet tool install -g ApiDoctor`, hence the reference project
- is not published for linux, hence the switch to windows

I'll go ahead and create an issue over there to get the teams attention, we can always revert/clean-up some of the changes made here after they unblock those aspects.

Result from the generation from that PR https://github.com/microsoftgraph/microsoft-graph-docs/pull/new/snippet-generation/60352

### DIFF
--- a/pipelines/snippets.yml
+++ b/pipelines/snippets.yml
@@ -20,19 +20,12 @@ resources:
      endpoint: microsoftgraph
      name: microsoftgraph/microsoft-graph-docs
      ref: main
-   - repository: apidoctor
-     type: github
-     endpoint: microsoftgraph
-     name: microsoftgraph/apidoctor
-     ref: master
 
 pool:
   vmImage: 'ubuntu-latest'
 
 variables:
   buildConfiguration: 'Release'
-  apidoctorPath: 'apidoctor'
-  apidoctorProjects: '$(apidoctorPath)/**/*.csproj'
   snippetLanguages: 'C#,JavaScript,Objective-C,Java,Go'
 
 steps:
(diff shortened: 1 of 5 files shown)


pipelines/snippets.yml
```diff
@@ -20,20 +20,14 @@ resources:
...
      ref: main
-   - repository: apidoctor
-     type: github
-     endpoint: microsoftgraph
-     name: microsoftgraph/apidoctor
-     ref: master
 
 pool:
-  vmImage: 'ubuntu-latest'
+  vmImage: 'windows-latest'
```
zengin, line 25: is there any performance loss with this change?

zengin (approved):

pipelines/snippets.yml
baywet: comparing [this one](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=60352&view=results) to [that one](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=60250&view=results) I'd say about 100% increase. to 7/8m. Not terrible but not great either.
Let's hold the merge for now and see how fast the api doctor team can publish a tool instead.
Raptor is much more performance sensible and I don't want to have to push it to windows because of this.

baywet:
closing in favor to submodule, less impact on the workflow, closer to the resource approach.

Issue #911, opened 2022-03-16
Resolve structured and collection-valued function parameters for PowerShell style


irvinesunday:
Resolve structured and collection-valued function parameters with `content` properties defined as below: 

![image](https://user-images.githubusercontent.com/40403681/158632844-5170ea05-d5e9-47bb-a10f-68dbb025f04e.png)

to:

![image](https://user-images.githubusercontent.com/40403681/158633185-3edfc40c-1acf-43f9-be18-dbcc310bf8d5.png)

Refer to this [issue](https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/902) thread for more details.

Issue #902, opened 2022-03-07
Test PowerShell integration with `Microsoft.OpenAPI.OData - v1.0.10-preview2` lib. 


irvinesunday:
The `Microsoft.OpenAPI.OData - v1.0.10-preview2` OData - OpenAPI conversion lib. is out! 
We need to test how this lib. integrates with PowerShell through DevX API via this test url: https://devxapitest.azurewebsites.net/ and bump up the lib version in this repo if successful.

peombwa:
Here are the results from the first test with PowerShell SDK:
- All but three modules build after adding directives to fix circular references and broken commands
- DeviceManagement.Functions fails because we are using `content` instead of `schema` in the parameter object for `GET /deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids)` operation.
![image](https://user-images.githubusercontent.com/7061532/157346607-89f1375a-8349-4d60-ab6a-2cdf2052611f.png)
AutoREST.PowerShell doesn't support parameter objects with content. In this case, there is no need to no need for a content-type in a parameter. The ideal parameter object should be:
  ``` yml
  /deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids):
    get:
      tags:
        - deviceManagement.Functions
      summary: Invoke function getRoleScopeTagsByIds
      operationId: deviceManagement_getRoleScopeTagsByIds
      parameters:
        - name: ids
          in: query
          description: 'Usage: ids={ids}'
          required: true
          schema:
            type: array
            items:
              type: string
        ...
  ```
- Sites and Files timeout after 8+ hours due to the number of paths and schemas. We will need to review the paths individually and remove those that are invalid.
  - Sites
     - v1.0
          ``` powershell
			➜ hidi validate --openapi .\v1.0\Sites.yml
			Path Items: 715
			Operations: 1257
			Parameters: 7309
			Request Bodies: 388
			Responses: 1257
			Links: 1558
			Callbacks: 0
			Schemas: 11632
          ```
     - beta
          ``` powershell
			➜ hidi validate --openapi .\beta\Sites.yml
			Path Items: 507
			Operations: 880
			Parameters: 4521
			Request Bodies: 282
			Responses: 880
			Links: 1069
			Callbacks: 0
			Schemas: 9820
          ``` 
   - Files
     - v1.0
          ``` powershell
			➜ hidi validate --openapi .\v1.0\Files.yml
			Path Items: 546
[...]

PR #913, opened 2022-03-17
Resolves structured and collection-valued function parameters for PowerShell style

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/911

This PR:
- Updates the properties of structured and collection-valued function parameters for PowerShell style by nulling the `content` property and setting the `schema` property appropriately as: 

```
  "schema": {
    "type": "array",
    "items": {
      "type": "string"
    }
```

- Adds a test to validate the above fix.
- Adds a string extension method in the `UtilityService` project and have it referenced in the above test.
- Add reference to `Microsoft.OpenApi` lib `v1.3.1-preview5` in the OpenAPIService tests. This is needed for the above test.
- Fixes a nullable `schema` bug.
- Other minor code refactorings.

### DIFF
--- a/OpenAPIService/PowershellFormatter.cs
+++ b/OpenAPIService/PowershellFormatter.cs
@@ -72,6 +72,11 @@ namespace OpenAPIService
                     // OperationIds to how they were being constructed in earlier versions of the lib.
                     operationId = ResolveActionFunctionOperationId(operation);
                 }
+
+                if ("function".Equals(operationType, StringComparison.OrdinalIgnoreCase))
+                {
+                    ResolveFunctionOperationParameterSchema(operation);
+                }
             }
 
             var charPos = operationId.LastIndexOf('.', operationId.Length - 1);
@@ -158,6 +163,26 @@ namespace OpenAPIService
             updatedOperationId = regex.Match(updatedOperationId).Value;
 
             return updatedOperationId;
+        }
+
+        private static void ResolveFunctionOperationParameterSchema(OpenApiOperation operation)
+        {
+            foreach (var parameter in operation.Parameters)
+            {
+                if (parameter.Content != null)
+                {
+                    // Replace with schema object
+                    parameter.Content = null;
+                    parameter.Schema = new OpenApiSchema
+                    {
+                        Type = "array",
+                        Items = new OpenApiSchema
+                        {
+                            Type = "string"
+                        }
+                    };
+                }
+            }
         }
     }
 }
--- a/OpenAPIService/PowershellFormatter.cs
+++ b/OpenAPIService/PowershellFormatter.cs
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------------------------------------------------------------------------------------------------
+// ------------------------------------------------------------------------------------------------------------------------------------------------------
 //  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 // ------------------------------------------------------------------------------------------------------------------------------------------------------
 
@@ -75,7 +75,7 @@ namespace OpenAPIService
 
                 if ("function".Equals(operationType, StringComparison.OrdinalIgnoreCase))
                 {
-                    ResolveFunctionOperationParameterSchema(operation);
+                    ResolveFunctionParameters(operation);
                 }
             }
 
@@ -165,13 +165,18 @@ namespace OpenAPIService
             return updatedOperationId;
         }
 
-        private static void ResolveFunctionOperationParameterSchema(OpenApiOperation operation)
+        /// <summary>
+        /// Resolves structured or collection-valued function parameters.
+        /// </summary>
+        /// <param name="operation">The target OpenAPI operation of the function.</param>
+        private static void ResolveFunctionParameters(OpenApiOperation operation)
         {
             foreach (var parameter in operation.Parameters)
             {
-                if (parameter.Content != null)
+                if (parameter.Content?.Any() ?? false)
                 {
-                    // Replace with schema object
+                    // Replace content with a schema object of type array
+                    // for structured or collection-valued function parameters
                     parameter.Content = null;
                     parameter.Schema = new OpenApiSchema
                     {
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -42,7 +42,7 @@ jobs:
         uses: actions/setup-dotnet@v2
         with:
           dotnet-version: 6.0.x
-      - uses: actions/checkout@v2.4.0
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
       - name: Cache SonarCloud packages
--- a/OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
+++ b/OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
@@ -771,7 +771,7 @@ namespace OpenAPIService.Test
                             }
                         }
                     },
-                    ["/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids)"] = new OpenApiPathItem()
+                    ["/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids={ids})"] = new OpenApiPathItem()
                     {
                         Operations = new Dictionary<OperationType, OpenApiOperation>
                         {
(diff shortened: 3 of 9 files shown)


OpenAPIService/PowershellFormatter.cs
```diff
@@ -159,5 +164,30 @@ private static string ResolveActionFunctionOperationId(OpenApiOperation operatio
...
+
+        /// <summary>
+        /// Resolves structured or collection-valued function parameters.
+        /// </summary>
+        /// <param name="operation">The target OpenAPI operation of the function.</param>
+        private static void ResolveFunctionParameters(OpenApiOperation operation)
+        {
+            foreach (var parameter in operation.Parameters)
+            {
+                if (parameter.Content?.Any() ?? false)
```
baywet, line 176: from https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/902#issuecomment-1069133352

> If the later works with OData, then we can definitely change the output of the CSDL conversion

My understanding is this will make powershell send `/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=1,2,3)` but kiota based sdks will still try to send `/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=["1","2","3"])`.

I'd argue we change the behavior in the conversion lib instead of this one off patch for powershell

irvinesunday: We know the second option works with OData convention. The 1st option is what I'm not sure about it working with OData. I had to change this for PowerShell to enable AutoREST build successfully.

peombwa: @baywet, @irvinesunday, as I understand it, the `ResolveFunctionParameters` method should only apply to `style=prowershell` as it is implemented in the PowerShell formatter. I see this change as a temp fix to allow PS to use 1.0.10 (without breaking), but the long-term fix is to add `content` parameter support to AutoREST.PowerShell so PS can also send `/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=["1","2","3"])`.

I'd say let's keep the fallback logic to `parameter.schema` in DevX API as a temporary solution as we wait for AutoREST.PowerShell to support `parameter.Content`. I'll open a feature request in the AutoREST.PowerShell repo.

baywet: but there's something I'm not clear on at this point
```
https://graph.microsoft.com/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids)
```
is not a valid RFC 6570 template

```
https://graph.microsoft.com/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids={ids})
```
is and results in
```
https://graph.microsoft.com/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=1,2,3)
```

so any kiota based SDK is NOT going to be able to do anything with this at the moment.

Before committing to any given solution, could you test this format with the service (which should work from memory) so we can implement a solution that is closer to standards and works for all SDKs please?

peombwa: Good catch @baywet! Also, the `in:` field of the parameter object should be `in: path` and not `in: query` if we want to send https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/902#issuecomment-1069133352.

Unfortunately, the service fails with a syntax error for `/beta/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids="1","2")` as it only expects `/beta/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=["1","2","3"])`. For PS, we can add a directive to fix this by forcing the cmdlet to send `...(ids=["1", "2"])` as we wait for https://github.com/Azure/autorest.powershell/issues/919.

irvinesunday: @peombwa from the [reference doc](http://docs.oasis-open.org/odata/odata-openapi/v1.0/cnprd01/odata-openapi-v1.0-cnprd01.html#_Toc472347995). it specifies that it should be a query option:

image

@baywet when you say that this path template is not valid: 
`https://graph.microsoft.com/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids)` does that imply that the doc. referenced above is in conflict with the RFC 6570 template? Where it specifies that the parameter alias should be represented as:

image


See [reference](http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part2-url-conventions/odata-v4.0-errata03-os-part2-url-conventions-complete.html#_Complex_and_Collection)

peombwa: > @peombwa from the [reference doc](http://docs.oasis-open.org/odata/odata-openapi/v1.0/cnprd01/odata-openapi-v1.0-cnprd01.html#_Toc472347995). it specifies that it should be a query option:
> 
> image

Got it! So in this case, the request url will be `/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids)?@ids=["1","2","3"]` and not `/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=["1","2","3"])`.

baywet: but shouldn't it be `/beta/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=1,2)` or `/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids)?@ids=[1,2,3]` (no quotes around values) since those are ints?

irvinesunday: According to the spec example --> `http://host/service/ProductsByColors(colors=@c)?@c=["red","green"]` it will be as @peombwa highlighted above --> `/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids)?@ids=["1","2","3"]`

It could be that Graph service implements it differently.

irvinesunday: Yes, you're right @baywet.
I propose we leave the conversion lib. as is with its current OData spec format: `https://graph.microsoft.com/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids)` 
and update DevX API to reflect the RFC 6570 format: `https://graph.microsoft.com/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids={ids})`

What's your opinion? 
cc: @darrelmiller

irvinesunday: I mean according to the [OData - OpenAPI Mapping](http://docs.oasis-open.org/odata/odata-openapi/v1.0/cnprd01/odata-openapi-v1.0-cnprd01.html) spec. format. The conversion lib. is designed to follow that spec. document and I think it should follow the specified path format as stipulated there. We should not change it to suit how the Graph service expects the path to be formatted, but rather we can do that manipulation here in DevX API.  The conversion lib. is used by other non-Graph clients.

darrelmiller: The OData - OpenAPI Mapping spec is intended as guidance as to how to map to OpenAPI. If somewhere in that document it says that this `https://graph.microsoft.com/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=@ids)` should be the output then it is wrong, because that's not a valid OpenAPI path for defining a path parameter.

irvinesunday: In that regard, the following is expected:
1. Update the conversion lib to reflect the correct path for structured and collection-valued function parameters `https://graph.microsoft.com/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids={ids})`.
2. Do not remove the `content` property added to functions with structured and collection-valued parameters in the conversion lib. so that this `/beta/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids=["1","2","3"])` can be sent to the service.
3. Leave this PR as is as a temp. fix for PowerShell as we wait on https://github.com/Azure/autorest.powershell/issues/919.
4. **Question**: The reference doc. outlines that the `in:` field of structured and collection-valued parameter objects be `in: query`. Should this be updated to be `in: path` ?

cc: @baywet @peombwa

baywet: makes sense to me, and yes it feels like the in property should be path. Thanks for the extended conversation to get to the bottom of this!

peombwa: Yes, the `in` property should be a `path`. Thanks for the fix!

baywet: resolving as the fixes in the conversion library have been implemented. Thanks for getting to the bottom of this everyone!

baywet (approved):

PR #925, opened 2022-04-04
Add request body content to improve autocomplete experience

### LINKED ISSUE #926: Return Request body Content for GE Autocomplete style
In order to improve on Graph Explorer's autocomplete experience, we need to surface the request body content for sample APIs whose content is not provided in the samples template.

MaggieKimani1:
## Overview
This PR adds an optional request body parameter that populates the request body content when using autocomplete on Graph Explorer.

Fixes #926 

### Before
image

### After
image

### DIFF
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -53,7 +53,7 @@ namespace GraphWebApi.Controllers
                                     [FromQuery]OpenApiStyle style = OpenApiStyle.Plain,
                                     [FromQuery]string format = null,
                                     [FromQuery]string graphVersion = null,
-                                    [FromQuery] bool requestBody = false,
+                                    [FromQuery]bool requestBody = false,
                                     [FromQuery]bool forceRefresh = false)
         {
             var styleOptions = new OpenApiStyleOptions(style, openApiVersion, graphVersion, format);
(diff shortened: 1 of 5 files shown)


baywet (approved):

danielmbaluka (dismissed):

OpenAPIService.Test/OpenAPIServiceShould.cs
```diff
@@ -292,6 +292,44 @@ public void ReturnOpenApiDocumentInCreateFilteredDocumentWhenValidArgumentsAreSp
             }
         }
 
+        [Fact]
+        public void ReturnContentForGEAutoCompleteStyleIfReQuestBodyIsTrue()
```
irvinesunday, line 296: We can update this test method to test for both `requestBody = true` and `requestBody = false`

irvinesunday (dismissed):

GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -53,6 +53,7 @@ public OpenApiController(IConfiguration configuration, IOpenApiService openApiSe
                                     [FromQuery]OpenApiStyle style = OpenApiStyle.Plain,
                                     [FromQuery]string format = null,
                                     [FromQuery]string graphVersion = null,
+                                    [FromQuery]bool requestBody = false,
```
zengin, line 56: I am late to the party, but could this be renamed to something like `includeRequestBody`? Current naming doesn't imply that it is a boolean query string parameter and the type information may not be available at the caller side.

zengin (approved):
Thank you @MaggieKimani1 for applying the change on parameter name.

irvinesunday (approved):

Issue #928, opened 2022-04-05
Missing key segments in the parameter object in the latest OpenAPI refresh.


peombwa:
**Describe the bug**
Missing key segments in the parameter object in latest OpenAPI refresh. This must be something recent as it wasn't there when we last tested OpenAPI.Net.OData preview 1.0.10 in https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/902.

**To Reproduce**
1. Download the latest OpenAPI file. e.g., [v1.0 Application.yml](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/po/OpenAPIRefresh/openApiDocs/v1.0/Applications.yml#L207-L289).
2. Go to a path with an indexer e.g., `'/applications/{application-id}'`.
3. Scroll down to any operation object and inspect its parameter object.
4. The indexer will be missing in the parameter object.
![image](https://user-images.githubusercontent.com/7061532/161864806-bc63d17b-f68b-4293-b94c-867c0d9f0b42.png)

**Expected behavior**
Indexers should be present in the parameter object e.g.,

**Screenshots**
![image](https://user-images.githubusercontent.com/7061532/161865962-6c5c482e-4dbd-4007-b5db-a7a4d7d3a72c.png)

baywet:
The parameter should be declared on the path item but not just missing from the operation. Are you not seeing it at the path item level?

irvinesunday:
@peombwa, does AutoREST care whether params are declared on the path item or on the parameters object?

peombwa:
@irvinesunday, I'm not entirely sure as it is not covered in their tests. We will have to test to verify if it's even supported.

irvinesunday:
Sorry @peombwa, @baywet, I missed this, the parameters appear at the bottom of the path item, after the operations:

![image](https://user-images.githubusercontent.com/40403681/162157283-ade139ae-ac3a-4c52-897d-7cad23396de3.png)

@peombwa  with this, AutoREST wasn't able to identify the segment param?

PR #929, opened 2022-04-06
Remove convert setting `DeclarePathParametersOnPathItem`

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/928

This PR:
- Removes the OpenAPI convert setting `DeclarePathParametersOnPathItem`. This setting removes path parameters from the parameter object

### DIFF
--- a/OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
+++ b/OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
@@ -133,6 +133,12 @@ namespace OpenAPIService.Test
                                                 }
                                             }
                                         }
+                                    },
+                                    Extensions = new Dictionary<string, IOpenApiExtension>
+                                    {
+                                        {
+                                            "x-ms-docs-operation-type", new OpenApiString("function")
+                                        }
                                     }
                                 }
                             }
@@ -191,6 +197,12 @@ namespace OpenAPIService.Test
                                                     }
                                                 }
                                             }
+                                        }
+                                    },
+                                    Extensions = new Dictionary<string, IOpenApiExtension>
+                                    {
+                                        {
+                                            "x-ms-docs-operation-type", new OpenApiString("function")
                                         }
                                     }
                                 }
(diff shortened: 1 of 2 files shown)


baywet (changes requested):
From the description in #928 it seems that the parameter is missing all together instead of being present at the path item level. I don't think we should disable this option but rather fix the behavior in the conversion library

irvinesunday:
> From the description in #928 it seems that the parameter is missing all together instead of being present at the path item level. I don't think we should disable this option but rather fix the behavior in the conversion library

This is currently blocking PowerShell generation. 
Also, I don't think we need this behaviour in DevX API.

irvinesunday:
> From the description in #928 it seems that the parameter is missing all together instead of being present at the path item level. I don't think we should disable this option but rather fix the behavior in the conversion library

This is currently blocking PowerShell generation. 
Also, I don't think we need this behaviour in DevX API.

baywet:
this feature has been introduced to significantly reduce the size of the OpenAPI description, which has been a challenge for generators, converters, etc...
the right thing to do here is to fix the feature, not toss the can down the road as it'll not only benefit kiota but also hidi, other generators, devx API AND PowerShell.

irvinesunday:
This behaviour actually works on the conversion lib. Need to figure out why DevX API is ejecting the parameters from the path item:

image

The parameters are at the bottom of the path item, after the operations.

baywet (changes requested):

danielmbaluka (approved):

irvinesunday:
DevX API is not translating the path item parameters because currently the slicing only walks the `OpenAPIOperations` object of the `OpenAPIPathItem`. This should be updated in Hidi, in which DevX API will take a dependency on. Issue is being tracked here: https://github.com/microsoft/OpenAPI.NET/issues/834

adhiambovivian:
@irvinesunday  Let's move with this PR so that we can unblock PowerShell to use 1.0.10 preview

Issue #927, opened 2022-04-04
OpenAPI code generators timing out 


rkodev:
**Describe the bug**
Investigate failing Snippet generation for Typescript and Go

Possible Causes: 
Failing SSL Validation when attempting to download OpenAPI yml file

Sample Run: 
https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=71427&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=e19b279f-2a41-5171-6e52-a43b27738914

error log 

```
Exception while processing /tmp/86314ac5-8d41-42d3-99f4-f8c0d3a6cfed/get-organization-1-beta-httpSnippet.The SSL connection could not be established, see inner exception.   at System.Net.Http.ConnectHelper.EstablishSslConnectionAsync(SslClientAuthenticationOptions sslOptions, HttpRequestMessage request, Boolean async, Stream stream, CancellationToken cancellationToken)   at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken) 
```

ddyett:
@rkodev is this work still blocked, if so what's the block

rkodev:
@ddyett We are still getting timeouts. Updated the ran to tests generation without validating SSL certificates as well as increasing timeouts but the validation is still failing https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=72095&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=12f1170f-54f2-53f3-20dd-22fc7dff55f9

zengin:
@rkodev I had a quick look at this issue, this doesn't seem to be failing because of SSL per se, the inner exception is still a timeout. I think, it might be worth looking into how lazy initialization is set up for open api documents. Ideally the document should be initialized only once, but it calls an async method in a synchronous manner through GetAwaiter().GetResult(). Please reach out to me if you want to investigate this together.

rkodev:
@zengin Thank you for this, removing the lazy initialization worked. Here is the test run https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=72307&view=results

rkodev:
@zengin Here is the PR for the changes https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/934

[...]

PR #934, opened 2022-04-11
Removes lazy initialization causing failing builds

rkodev:
## Overview
Resolves https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/927
This PR removes Lazy Initialization of GetOpenAPIDocument Method

### Notes

## Testing Instructions
[Sample generation run](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=72310&view=results)

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/OpenAPISnippetsGenerator.cs
+++ b/CodeSnippetsReflection.OpenAPI/OpenAPISnippetsGenerator.cs
@@ -33,10 +33,10 @@ namespace CodeSnippetsReflection.OpenAPI
             if(string.IsNullOrEmpty(betaOpenApiDocumentUrl)) throw new ArgumentNullException(nameof(betaOpenApiDocumentUrl));
 
             _telemetryClient = telemetryClient;
-            _v1OpenApiDocument = new Lazy<OpenApiUrlTreeNode>(() => GetOpenApiDocument(v1OpenApiDocumentUrl).GetAwaiter().GetResult(), LazyThreadSafetyMode.PublicationOnly);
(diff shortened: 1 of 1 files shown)


MaggieKimani1 (commented):
I could argue that we're not necessarily removing the lazy initialization, but the [lazy thread safety mode](https://docs.microsoft.com/en-us/dotnet/api/system.threading.lazythreadsafetymode?view=net-6.0#remarks).
Also may I ask why we're removing this? If we remove this, the exception caching behavior resumes and Snippets on GE will fail due to timing out.
I added it in order to prevent caching of the HttpClient timeout exceptions that were being rethrown on multiple subsequent requests when multiple threads try to access the value set by the `Lazy<T>` instance.

rkodev:
@MaggieKimani1 Just updated the PR to address the issue with cache you mentioned. The changes were necessitated to address the issue [927](https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/927).

We traced the timeouts to this code when fetching OpenAPI docs to this specific change, looks like the Lazy Object was not getting initialized possibly because of a race condition during execution and resulting to the pipeline getting runs stuck there. Changing this to `LazyThreadSafetyMode.ExecutionAndPublication` also addresses the problem

baywet (approved):

zengin (changes requested):
I agree with @andrueastman. On a side note, the private async method of GetOpenApiDocument is still being called synchronously. If we want to initialize the snippet generator classes asynchronously, I'd recommend this factory pattern as there is no way to mark constructors async: https://blog.stephencleary.com/2013/01/async-oop-2-constructors.html

baywet:
for everyone, we've had a long discussion about that with DeVere and Ronald a couple of minutes ago, I've synthetized some of the technical aspects here https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/927#issuecomment-1095462453

PR #936, opened 2022-04-11
Update convert settings to show links for autocomplete experience

### LINKED ISSUE #935: Missing Links from OpenAPI spec used for autocomplete experience

MaggieKimani1:
Fixes #935

### DIFF
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -633,8 +633,8 @@ namespace OpenAPIService
                 EnableDiscriminatorValue = false,
                 EnableDerivedTypesReferencesForRequestBody = false,
                 EnableDerivedTypesReferencesForResponses = false,
-                ShowRootPath = false,
-                ShowLinks = false
+                ShowRootPath = true,
+                ShowLinks = true
             };
             OpenApiDocument document = edmModel.ConvertToOpenApi(settings);
 


baywet (changes requested):
the style parameter on the controller should be driving those settings, and we should have presets for each style. This is why we've been pulling the blanket in every direction for the past couple of weeks and breaking different experiences.

MaggieKimani1:
> the style parameter on the controller should be driving those settings, and we should have presets for each style. This is why we've been pulling the blanket in every direction for the past couple of weeks and breaking different experiences.

I have added some convert settings set in the controller for GEAutocomplete which should unblock GE for now till @irvinesunday is back and some Engineering time is allocated for this work item.

darrelmiller:
The style parameter should be the driver of the behavior in DevXAPI.  It encapsulates the scenarios we are trying to support.  I think Maggie's fix is a necessary temporary fix to make Autocomplete in GE work again.  PowerShell/PowerPlatform/Plain will not be impacted by having the links present.  It just makes the document bigger.

The next steps should be:
1) Identify what the OpenApiConvertSettings should be for each style
2) Update the caching mechanism in DevX API so that we have a cached copy of the OpenAPI for each style

I can help with the first item.  @adhiambovivian Where would you like that to be documented?  Can you create an issue for the second item?

baywet (approved):

Issue #924, opened 2022-04-01
Namespace-simplified function urls don't return permissions


irvinesunday:
Urls of namespace-simplified function paths don't return permissions, example: `/reports/getemailactivityusercounts(period={value})`

**Expected result**:
`[{"value":"Reports.Read.All","consentDisplayName":"Read all usage reports","consentDescription":"Allows an app to read all service usage reports on your behalf. Services that provide usage reports include Office 365 and Azure Active Directory.","isAdmin":true}]`

**Actual result**:
`{"type":"https://tools.ietf.org/html/rfc7231#section-6.5.4","title":"Not Found","status":404,"traceId":"00-4bcebe86a8f0ae70dc308046724859c2-33ae785ec7260290-00"}`

Related to https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/1575

PR #948, opened 2022-04-18
Removes function parameters from incoming urls when querying for permissions

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/924

This PR addresses the below when querying for permissions:
- Fixes a bug that was translating namespace-simplified function parameters as path segments.
- Removes function parameters from incoming request urls.
- Updates the permission test appropriately.

### DIFF
--- a/PermissionsService.Test/PermissionsStoreShould.cs
+++ b/PermissionsService.Test/PermissionsStoreShould.cs
@@ -183,32 +183,19 @@ namespace PermissionsService.Test
               });
         }
 
-        [Fact]
-        public void RemoveFunctionParametersFromRequestUrlsDuringLoadingAndQueryingOfPermissionsFiles()
+        [Theory]
+        [InlineData("/workbook/worksheets/{id}/charts/{id}/image")]
+        [InlineData("/workbook/worksheets/{id}/charts/{id}/image(width={value})")]
+        public void RemoveFunctionParametersFromRequestUrlsDuringLoadingAndQueryingOfPermissionsFiles(string url)
         {
             // Act
-            // RequestUrl in permission file: "/workbook/worksheets/{id}/charts/{id}/image(width=640)"
             List<ScopeInformation> result =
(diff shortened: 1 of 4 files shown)


baywet (dismissed):

zengin (approved):

PermissionsService.Test/PermissionsStoreShould.cs
```diff
@@ -184,11 +184,15 @@ public void ReturnScopesForRequestUrlsInEitherPermissionFilesProvided()
         }
 
         [Fact]
```
millicentachieng, line 186: This test can be written as follows (Line 186 - 219):

```
[Theory]
[InlineData("/workbook/worksheets/{id}/charts/{id}/image")]
[InlineData("/workbook/worksheets/{id}/charts/{id}/image(width={value})")]
public void RemoveFunctionParametersFromRequestUrlsDuringLoadingAndQueryingOfPermissionsFiles(string url)
{
	// Act
	List<ScopeInformation> result =
		_permissionsStore.GetScopesAsync(scopeType: DelegatedWork,
										 requestUrl: url,
										 method: "GET").GetAwaiter().GetResult();

	/* Assert */
	Assert.Collection(result,
		item =>
		{
			Assert.Equal("Files.ReadWrite", item.ScopeName);
			Assert.Equal("Have full access to your files", item.DisplayName);
			Assert.Equal("Allows the app to read, create, update, and delete your files.", item.Description);
			Assert.False(item.IsAdmin);
		});
}
```
Sorry, a GitHub suggestion doesn't work here.

millicentachieng (changes requested):

PermissionsService.Test/PermissionsStoreShould.cs
irvinesunday: I was wondering how I'd do that. Thanks 😄

millicentachieng (approved):

PR #1061, opened 2022-07-12
Fix Sonarcloud issue

SilasKenneth:
## Overview

- Fix SonarCloud issue with unescaped regex pattern. (Still don't know the negative effect this fix might have).
### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/ModelGraph/SnippetCodeGraph.cs
+++ b/CodeSnippetsReflection.OpenAPI/ModelGraph/SnippetCodeGraph.cs
@@ -271,11 +271,10 @@ namespace CodeSnippetsReflection.OpenAPI.ModelGraph
                 return new CodeProperty { Name = propertyName, Value = value.GetString(), PropertyType = PropertyType.Guid, Children = new List<CodeProperty>() };
 
             var enumSchema = propSchema?.AnyOf.FirstOrDefault(x => x.Enum.Count > 0);
-            if (enumSchema == null)
+            if ((propSchema == null || propSchema.Enum.Count == 0) && enumSchema == null)
                 return new CodeProperty { Name = propertyName, Value = escapeSpecialCharacters(value.GetString()), PropertyType = PropertyType.String, Children = new List<CodeProperty>() };
-
-
-            var propValue = String.IsNullOrWhiteSpace(value.GetString()) ? null : $"{enumSchema.Title.ToFirstCharacterUpperCase()}.{value.GetString().ToFirstCharacterUpperCase()}";
+            enumSchema ??= propSchema;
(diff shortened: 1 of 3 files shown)


rkodev (dismissed):

pipelines/snippets.yml
```diff
@@ -28,7 +28,7 @@ variables:
   buildConfiguration: 'Release'
   apidoctorPath: 'microsoft-graph-devx-api/apidoctor'
   apidoctorProjects: '$(apidoctorPath)/**/*.csproj'
-  snippetLanguages: 'C#,JavaScript,Java,Go,PowerShell,PHP'
+  snippetLanguages: 'C#,JavaScript,Java,Go,PowerShell'
```
baywet, line 31: I'd argue we should keep PHP since we don't have snippets for the older version of the SDK AFAIK and since Go has snippets (but with a warning message) https://docs.microsoft.com/en-us/graph/api/user-list?view=graph-rest-1.0&tabs=go

baywet (changes requested):

pipelines/snippets.yml
rkodev: @baywet Should this apply to Typescript as well?

baywet: TypeScript is a different story since we already have a JavaScript tab here.
Same for C#, or Java when the preview comes out.
PMs are working on a design for languages that already have snippets.

baywet (dismissed):
I can see that you manually did the opposite change. In case you didn't know you could:
1. git revert commit-sha (to apply an automatically generated opposite commit)
2. git rebase -i (and then drop in front of the commit to drop it from history, requires a force push afterwards)

CodeSnippetsReflection.OpenAPI/ModelGraph/SnippetCodeGraph.cs
```diff
@@ -271,11 +271,10 @@ private static CodeProperty evaluateStringProperty(string propertyName, JsonElem
                 return new CodeProperty { Name = propertyName, Value = value.GetString(), PropertyType = PropertyType.Guid, Children = new List<CodeProperty>() };
 
             var enumSchema = propSchema?.AnyOf.FirstOrDefault(x => x.Enum.Count > 0);
-            if (enumSchema == null)
+            if ((propSchema == null || propSchema.Enum.Count == 0) && enumSchema == null)
```
baywet, line 274:
```suggestion
            if (enumSchema == null || propSchema?.Enum.Count == 0)
```

baywet (changes requested):

baywet (approved):

PR #1071, opened 2022-07-13
Add PHP tag for syntax highlighting + Fix minor bug

SilasKenneth:
## Overview

The current PHP snippets don't have PHP Syntax highlighting.

- Add PHP tag to beginning of snippet for GE syntax highlighting.
- Fix issue with handling endpoints with `()` leading to `()()` in generating snippet fluent path.

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/LanguageGenerators/PhpGenerator.cs
+++ b/CodeSnippetsReflection.OpenAPI/LanguageGenerators/PhpGenerator.cs
@@ -23,7 +23,8 @@ public class PhpGenerator : ILanguageGenerator<SnippetModel, OpenApiUrlTreeNode>
         var indentManager = new IndentManager();
         var codeGraph = new SnippetCodeGraph(snippetModel);
         var payloadSb = new StringBuilder(
-            "//THIS SNIPPET IS A PREVIEW FOR THE KIOTA BASED SDK. NON-PRODUCTION USE ONLY" + Environment.NewLine +
+            "<?php" + Environment.NewLine + Environment.NewLine +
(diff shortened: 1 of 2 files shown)


CodeSnippetsReflection.OpenAPI/LanguageGenerators/PhpGenerator.cs
```diff
@@ -23,7 +23,8 @@ public string GenerateCodeSnippet(SnippetModel snippetModel)
         var indentManager = new IndentManager();
         var codeGraph = new SnippetCodeGraph(snippetModel);
         var payloadSb = new StringBuilder(
-            "//THIS SNIPPET IS A PREVIEW FOR THE KIOTA BASED SDK. NON-PRODUCTION USE ONLY" + Environment.NewLine +
+            "<?php" + Environment.NewLine + Environment.NewLine +
```
baywet, line 26: If you're adding the php opening tag, shouldn't the php closing tag be added as well for consitency?

SilasKenneth: The end tag is only required if we are having code that is not PHP code in the snippet. From [this](https://www.php.net/manual/en/language.basic-syntax.phptags.php)

> If a file contains only PHP code, it is preferable to omit the PHP closing tag at the end of the file. This prevents accidental whitespace or new lines being added after the PHP closing tag, which may cause unwanted effects because PHP will start output buffering when there is no intention from the programmer to send any output at that point in the script.

baywet: Ahh I didn't know that, last time I wrote PHP, php4 was "new".
ImInMyPrimeFunnyGIF

baywet (approved):

PR #1086, opened 2022-07-20
Fix issue of broken Regex slicing by tags

### LINKED ISSUE #1082: Slicing by tags using REGEX is broken.
**Describe the bug**
Slicing of OpenAPI documents by tags via REGEX is broken and fails with `{"StatusCode":404,"Message":"No paths found for the supplied parameters."}`.

Last successful run was on `07/13/2022`. 

**To Reproduce**
- https://graphexplorerapi.azurewebsites.net/$openapi?tags=^groups.group$&title=Groups&openapiversion=3&style=Powershell&graphVersion=v1.0
- https://graphexplorerapi.azurewebsites.net/$openapi?tags=^teams\.&title=Teams&openapiversion=3&style=Powershell&graphVersion=v1.0
- https://graphexplorerapi.azurewebsites.net/$openapi?tags=^security\.&title=Security&openapiversion=3&style=Powershell&graphVersion=v1.0

Fail with:
`{"StatusCode":404,"Message":"No paths found for the supplied parameters."}`

**Expected behavior**
A sliced OpenAPI document should be returned based on the provided REGEX in the tags query parameter.

PowerShell uses a REGEX based mapping to slicing Microsoft Graph APIs into modules. See https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/config/ModulesMapping.jsonc.

MaryannGitonga:
## Overview

Fixes #1082 

This PR:

* Fixes the broken OpenAPI regex slicing by removing the `Regex.Escape()` function. The `Regex.Escape()` function was flagged as a security vulnerability by SonarCloud so as to prevent malicious attacks via regex. However, using this approach breaks the regex matching engine and thus, a better way needs to be investigated for escaping malicious regex metacharacters.

* Updates the ReturnOpenApiDocumentInCreateFilteredDocumentWhenValidArgumentsAreSpecified test by adding a test case for a regex-formatted tag.

### DIFF
--- a/OpenAPIService.Test/OpenAPIServiceShould.cs
+++ b/OpenAPIService.Test/OpenAPIServiceShould.cs
@@ -12,6 +12,7 @@ using System.Collections.Concurrent;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using UtilityService;
 using Xunit;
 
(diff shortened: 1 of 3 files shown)


MaggieKimani1:
@irvinesunday I think we can get this fix merged in order to unblock Powershell.

MaggieKimani1 (approved):

Issue #1101, opened 2022-07-22
Implement Threadlocking for OpenApiService during OpenAPI conversion


MaggieKimani1:
- DevX API is recording multiple failed requests with `SystemOutOfMemory` exceptions when calling the `/openapi` endpoint.
- On Inspecting telemetry logs, we discover that we're experiencing a high CPU and memory usage percentage especially during 
deployments.
- One probable cause for this would be that multiple threads are trying to implement the `CsdlToOpenApiConversion` process that includes fixing references, which is resource intensive.
- A potential fix for this issue would be to implement thread locking such that one single thread initializes the conversion process and fixes the references in the resulting OpenApi document instance in a thread-safe manner, while subsequent threads/requests use the cached instance

Issue #1055, opened 2022-07-05
DevX API Reliability Issue


peombwa:
**Describe the bug**
`500 out-of-memory` error when calling DevX API with `forceRefresh=true` when `graphVersion` is `beta` and intermittently in `v1.0`.

**To Reproduce**
1. Open https://graphexplorerapi.azurewebsites.net/openapi/operations?openapiversion=3&style=Powershell&graphVersion=beta&forceRefresh=true from your browser.
2. See error: `{"StatusCode":500,"Message":"Exception of type 'System.OutOfMemoryException' was thrown."}`

**Expected behavior**
An OpenAPI document should be returned when `forceRefresh=true`.

**Additional context**
The issue is blocking PowerShell's weekly OpenAPI refreshes from getting the latest API refreshes.

irvinesunday:
According to App Insights telemetry, this is where the exception gets thrown: https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/7114d5f9441d1a8f3edae59326a18b6b9e418749/OpenAPIService/OpenApiService.cs#L665. 
During fixing of references in the OpenAPI document, the document is first serialized into a `StringBuilder` [object](https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/7114d5f9441d1a8f3edae59326a18b6b9e418749/OpenAPIService/OpenApiService.cs#L663). The conversion of the `StringBuilder` object to a string and the `OpenAPI.NET` library method that reads this string (`OpenApiStringReader().Read()`) ends up allocating a significant amount of memory for this process, especially for `beta` paths, which are appr. double the size of `v1.0` paths. There is need to review this fixing of references.

irvinesunday:
This has been temporarily resolved by scaling up the DevX API app service plan to allocate more memory.

Issue #851, opened 2022-01-12
Investigate rise in failure anomalies 


MaggieKimani1:
The `OpenApi/Get` endpoint has been recording a lot of failed requests coming in.
The failure rate has spiked to 74% over the last 12hrs which is quite high compared to the normal failure rate of 28.4%
As per our telemetry board, the most prevalent error codes are error `404` and `400s`.
This ticket serves to investigate the root cause of the failures and come up with a solution to reduce the incoming errors.

zengin:
possibly related: https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/1335

MaggieKimani1:
> possibly related: [microsoftgraph/microsoft-graph-explorer-v4#1335](https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/1335)

Thank you for raising this. Investigating it further..

zengin:
I think it is relatively cheap to validate the version information on the client side to avoid calls that are guaranteed to fail. Currently GE is sending version information by extracting the first 4 letters that comes after graph.microsoft.com, e.g. `graph.microsoft.com/randomversion` sends `rand` as the version.

GE already has a constant list of versions, i.e. the version dropdown. I think we should add that validation to avoid the roundtrip to the service.

From DevX API standpoint, I think the service is doing the right thing and returning a 400 response for invalid version parameters.

cc: @thewahome

RabebOthmani:
@irvinesunday from your comment above, I believe the issue is not solved. Is that correct? In which case, let's prioritize this.

zengin:
@irvinesunday @MaggieKimani1 is there a chance we can see the correlation of this data with the service startup events? As part of an incident last week, I had to restart the service and observed high latency and larger request times right after that, which settled after about half an hour. It looks like we can do optimizations around service startup (this is likely a cache warm-up issue). I am not sure if it is related to the problem raised here, but it can be a potential place for investigation.

I will create a separate issue from this comment if they turn out to be unrelated.

PR #1102, opened 2022-07-22
Implements thread locking in OpenApiService during OpenAPI conversion

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1101
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1055
Potentially Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/851

This PR:
- Updates the OpenAPI conversion process by ensuring that only one thread converts an OpenAPI document. This will help prevent multiple processes trying to perform the conversion before the cache has been set. Having multiple requests attempting to perform the conversion increases both CPU and memory load.
- Introduces a _forceRefresh_ elapsed time (in minutes) check that will be used to determine the period in which `forceRefresh`  updates are allowed. This will help in mitigating potential DoS requests from either bad or legitimate actors. The current default value is 10 minutes. This value is configurable in the app service configuration settings.

### DIFF
--- a/OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
+++ b/OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
@@ -1,7 +1,8 @@
-// ------------------------------------------------------------------------------------------------------------------------------------------------------
+﻿// ------------------------------------------------------------------------------------------------------------------------------------------------------
 //  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 // ------------------------------------------------------------------------------------------------------------------------------------------------------
 
+using Microsoft.Extensions.Configuration;
 using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Models;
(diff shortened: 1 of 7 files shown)


MaggieKimani1 (dismissed):

MaggieKimani1 (approved):

OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
```diff
@@ -27,6 +29,15 @@ public OpenApiDocumentCreatorMock(IOpenApiService openApiService)
                 ?? throw new ArgumentNullException(nameof(openApiService), $"{ NullValueError }: { nameof(openApiService) }");
         }
 
+        public static IOpenApiService GetOpenApiService(string configFilePath)
+        {
+            IConfigurationRoot configuration = new ConfigurationBuilder()
+                                                    .AddJsonFile(configFilePath)
```
baywet, line 35: you might want to enable refresh on change so we don't need to kill the process on configuration chagnes)

OpenAPIService/OpenApiService.cs
```diff
@@ -500,19 +511,61 @@ public async Task<OpenApiDocument> GetGraphOpenApiDocumentAsync(string graphUri,
...
+                    _telemetryClient?.TrackTrace("OpenAPI document converted and cached by another thread." +
+                                                 "Retrieve the cached document.",
+                                                 SeverityLevel.Information,
+                                                 _openApiTraceProperties);
+
+                    return doc;
+                }
+
+                // Check whether forceRefresh is requested before the allowable refresh period has elapsed.
+                if (_OpenApiDocumentsDateCreated.TryGetValue(csdlHref, out DateTime dateCreated))
```
baywet, line 532: all the cache handling is mixing concerns which is going to make maintenance much harder....
I already provided a long explanation on the topic, we have three separate concerns to handle here, and it would be much easier to implement, test, maintain if they were separated https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/927#issuecomment-1095462453

baywet (commented):
Sorry I'm playing catch up after being on vacation. I do think we're trying to do everything at once in a single place here when separating concerns would make things easier.

PR #1104, opened 2022-07-22
Fix: Allow two threads to convert the OpenAPI document asynchronously

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1101
This PR:
- Allows only two threads to convert the OpenAPI document synchronously - provided they are converting different Graph versions of the OpenAPI.

### DIFF
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -553,7 +553,7 @@ namespace OpenAPIService
                                              _openApiTraceProperties);
                 _openApiTraceProperties.Remove(UtilityConstants.TelemetryPropertyKey_SanitizeIgnore);
 
-                OpenApiDocument source = await CreateOpenApiDocumentAsync(csdlHref);
+                OpenApiDocument source = CreateOpenApiDocumentAsync(csdlHref).GetAwaiter().GetResult();
                 _OpenApiDocuments[csdlHref] = source;
                 _OpenApiDocumentsDateCreated[csdlHref] = DateTime.UtcNow;
                 return source;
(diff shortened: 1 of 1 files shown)


zengin:
Curious to know, are we able to reproduce the service behavior locally? Do we observe similar resource exhaustion if we send multiple requests right after service startup?

Looks like there is some locking in place preventing multiple threads trying to initialize the document based on the comments in the code. So the question is:
1. Is the locking mechanism not working?
2. Is a single thread causing out of memory error?

If it is the case 1, then we should revisit the locking implementation.
If it is the case 2, then I don't think changing logic to convert OpenAPI doc synchronously won't solve the problem.

In any case, I think we need an isolated local repro of the issue. Please let me know if you want to take a stab at this together. Sorry if I missed any prior context and an earlier thorough investigation resulted in this PR as the solution.

irvinesunday:
> Curious to know, are we able to reproduce the service behavior locally? Do we observe similar resource exhaustion if we send multiple requests right after service startup?
> 
> Looks like there is some locking in place preventing multiple threads trying to initialize the document based on the comments in the code. So the question is:
> 
> 1. Is the locking mechanism not working?
> 2. Is a single thread causing out of memory error?
> 
> If it is the case 1, then we should revisit the locking implementation. If it is the case 2, then I don't think changing logic to convert OpenAPI doc synchronously won't solve the problem.
> 
> In any case, I think we need an isolated local repro of the issue. Please let me know if you want to take a stab at this together. Sorry if I missed any prior context and an earlier thorough investigation resulted in this PR as the solution.

Thank you for jumping in on this @zengin 

We are planning to load test the server with _/openapi_ requests to attempt to validate the high CPU/Memory usage. We will get back to you once this is successful. 

1. The locking mechanism is working, I have left the conversion to run asynchronously as before. I have only updated the mechanism so that we can allow two threads to perform the conversion (provided each is converting a different Graph version of the OpenAPI document). I have updated the PR title to reflect this. 

I shall reach out to you in case we come out stuck with something.

MaggieKimani1 (approved):

baywet:
FYI @zengin here is a related comment I left on another PR https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1110#issuecomment-1205124984

PR #1110, opened 2022-07-27
[Fix] Safely exit all recursive threads and avoid blocking the thread lock when converting the OpenAPI doc.

irvinesunday:
This PR:

- Ensures we safely exit all recursive threads by using a `Queue` to store and remove the graph versions that are used to check whether a thread will be recursed back to await a lock. This `Queue` is dequeued at the `finally` block, so as to ensure this happens even at the event of an exception. This will help avoid blocking the release of the lock by the recursed threads.
- Increases the OpenAPI document default refresh time to 15 mins. Our Azure availability tests run every 15 mins. This is to keep it in line with our availability tests frequency.

### DIFF
--- a/GraphWebApi/appsettings.json
+++ b/GraphWebApi/appsettings.json
@@ -57,6 +57,6 @@
     "TourSteps": "24"
   },
   "FileCacheRefreshTimeInMinutes": {
-    "OpenAPIDocuments": "10"
+    "OpenAPIDocuments": "15"
   }
 }
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------------------------------------------------------------------------------------------------
+// ------------------------------------------------------------------------------------------------------------------------------------------------------
 //  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 // -------------------------------------------------------------------------------------------------------------------------------------------------------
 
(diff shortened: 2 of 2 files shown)


MaggieKimani1 (approved):

baywet:
Queue is not thread safe, [concurrent queue](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentqueue-1?view=net-6.0) is.
I'm not sure this has much to do with the data structure (queue vs hashset) more than how concurrency is handled and the fact that entries were removed in the main block instead of the finally. (that change I believe has a positive impact).

What you're effectively trying to achieve is a memoization pattern where you want a single thread do to the works (for a specific entry) and all the other ones (for a specific entry) to wait for that thread to complete and use this result.
https://medium.com/nerd-for-tech/javascript-memoization-patterns-f0438f96a696

This could be achieved simply by keeping a reference to the Task that does the work (through the use of a concurrent dictionary), if it's completed, it'll yield the result right away, if it's still running, it'll await and block until the result is available.
I hope this helps.

irvinesunday:
> Queue is not thread safe, [concurrent queue](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentqueue-1?view=net-6.0) is. I'm not sure this has much to do with the data structure (queue vs hashset) more than how concurrency is handled and the fact that entries were removed in the main block instead of the finally. (that change I believe has a positive impact).
> 
> What you're effectively trying to achieve is a memoization pattern where you want a single thread do to the works (for a specific entry) and all the other ones (for a specific entry) to wait for that thread to complete and use this result. https://medium.com/nerd-for-tech/javascript-memoization-patterns-f0438f96a696
> 
> This could be achieved simply by keeping a reference to the Task that does the work (through the use of a concurrent dictionary), if it's completed, it'll yield the result right away, if it's still running, it'll await and block until the result is available. I hope this helps.

Thanks for your input @baywet, I will def. take a look at your suggestion.

PR #1124, opened 2022-08-04
Fix minor type format issue. (PHP, Go)

SilasKenneth:
## Overview

Captures more type formats for string values.
 - Boolean.
 - Double.
 - Duration.
It also simplifies the `evaluateStringProperty` method in `SnippetCodeGraph.cs`

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
+++ b/CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
@@ -272,6 +272,9 @@ namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators {
                     builder.AppendLine($"{indentManager.GetIndent()}{NormalizeJsonName(child.Name.ToFirstCharacterLowerCase())} := []byte(\"{child.Value.ToFirstCharacterLowerCase()}\")");
                     builder.AppendLine($"{propertyAssignment}.Set{propertyName.ToFirstCharacterUpperCase()}(&{propertyName}) ");
                     break;
+                case PropertyType.Duration:
+                    WriteStringProperty(propertyAssignment, codeProperty, builder, indentManager, child);
+                    break;
                 default:
                     builder.AppendLine($"{indentManager.GetIndent()}{NormalizeJsonName(child.Name.ToFirstCharacterLowerCase())} := {child.Value.ToFirstCharacterLowerCase()}");
                     builder.AppendLine($"{propertyAssignment}.Set{propertyName.ToFirstCharacterUpperCase()}(&{propertyName}) ");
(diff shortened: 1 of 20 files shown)


CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
```diff
@@ -272,6 +272,9 @@ private static void WriteCodeProperty(string propertyAssignment, StringBuilder b
                     builder.AppendLine($"{indentManager.GetIndent()}{NormalizeJsonName(child.Name.ToFirstCharacterLowerCase())} := []byte(\"{child.Value.ToFirstCharacterLowerCase()}\")");
                     builder.AppendLine($"{propertyAssignment}.Set{propertyName.ToFirstCharacterUpperCase()}(&{propertyName}) ");
                     break;
+                case PropertyType.Duration:
```
baywet, line 275: just call parse iso duration https://github.com/microsoft/kiota-abstractions-go/blob/1ecd925f435fb1aaf0174576a0ef222044c7217a/serialization/iso_duration.go#L85

CodeSnippetsReflection.OpenAPI/ModelGraph/SnippetCodeGraph.cs
```diff
@@ -22,6 +23,20 @@ public record SnippetCodeGraph
 
         private static readonly CodeProperty EMPTY_PROPERTY = new() { Name = null, Value = null, Children = null, PropertyType = PropertyType.Default };
 
+        private static Dictionary<string, PropertyType> _formatPropertyTypes = new ()
```
baywet, line 26:
```suggestion
        private static Dictionary<string, PropertyType> _formatPropertyTypes = new (StringComparer.OrdinalIgnoreCase)
```

baywet (changes requested):

CodeSnippetsReflection.OpenAPI/ModelGraph/SnippetCodeGraph.cs
SilasKenneth: I think given that the keys are in Lowercase and I'm doing case conversion on line https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/e948a26de6cc2c7ecf8d16f5ccc6c18dd3472aec/CodeSnippetsReflection.OpenAPI/ModelGraph/SnippetCodeGraph.cs#L281
we might not need to add the suggestion.

baywet: it is generally better to use the right comparer than forcing a casing before comparison.
https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
SilasKenneth: Done!

baywet (approved):

PR #1141, opened 2022-08-13
Improve performance and reduce memory usage of /openapi/tree

darrelmiller:
## Overview

This PR changes the way the `/openapi/tree` action returns the response payload to minimize memory usage.

### Notes
This change takes advantage of the fact that the ASP.NET Response object provides access to a BodyWriter instance that can be passed to the constructor of Urtf8JsonWriter.  This enables the server to construct the JSON response and have it written directly to the network stream without having to have the entire document buffered in memory.

A further enhancement could be made to cache the entire UrlNodeTree rather than reconstructing on every request.

One downside to this approach is that if an exception occurs whilst writing the tree, the HTTP response will already have returned a 200 response.

## Testing Instructions

Calling `https://localhost:5001/openapi/tree?graphVersions=v1.0` from multiple concurrent threads shows reduced memory usage on the server and shorter time to first byte.

### DIFF
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -129,10 +129,12 @@ namespace GraphWebApi.Controllers
                 sources.TryAdd(graphVersion, await _openApiService.GetGraphOpenApiDocumentAsync(graphUri, forceRefresh));
             }
 
+            var rootNode = _openApiService.CreateOpenApiUrlTreeNode(sources);
+            
             Response.ContentType = "application/json";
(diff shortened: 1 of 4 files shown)


baywet (approved):

irvinesunday:
Thank you @darrelmiller for this fix. From my local tests the tree doesn't render on the browser. The expectation is to be able to render the tree on the browser as well, as is currently in prod. 

image

baywet:
I do get the tree to show up once I fixed the launch configuration and made sure an exception pause wasn't being raised (I don't have the connection strings to the storage account)
image

PR #1143, opened 2022-08-13
Updated Permission Service to simplify caching

darrelmiller:
## Overview

To try and minimize memory usage, the caching of permission data has been simplified and no longer uses the recyclable memory stream.

### DIFF
--- a/GraphWebApi/appsettings.json
+++ b/GraphWebApi/appsettings.json
@@ -14,8 +14,8 @@
   },
   "AllowedHosts": "*",
   "GraphMetadata": {
-    "V1.0": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsv1.0.xml",
-    "Beta": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_beta_metadata/cleanMetadataWithDescriptionsbeta.xml"
+    "V1.0": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsv1.0.xml",
+    "Beta": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_beta_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsbeta.xml"
   },
   "BlobStorage": {
     "AzureConnectionString": "ENTER_AZURE_STORAGE_CONNECTION_STRING",
--- a/GraphWebApi/appsettings.json
+++ b/GraphWebApi/appsettings.json
@@ -14,8 +14,8 @@
   },
   "AllowedHosts": "*",
   "GraphMetadata": {
-    "V1.0": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsv1.0.xml",
-    "Beta": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_beta_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsbeta.xml"
+    "V1.0": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsv1.0.xml",
+    "Beta": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_beta_metadata/cleanMetadataWithDescriptionsAndAnnotationsbeta.xml"
   },
   "BlobStorage": {
     "AzureConnectionString": "ENTER_AZURE_STORAGE_CONNECTION_STRING",
--- a/PermissionsService/Services/PermissionsStore.cs
+++ b/PermissionsService/Services/PermissionsStore.cs
@@ -26,7 +26,7 @@ namespace PermissionsService
     public class PermissionsStore : IPermissionsStore
     {
         
-        private readonly IMemoryCache _Cache;
+        private readonly IMemoryCache _cache;
         private readonly IFileUtility _fileUtility;
         private readonly IHttpClientUtility _httpClientUtility;
         private readonly IConfiguration _configuration;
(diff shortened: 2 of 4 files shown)


TelemetryService/CustomPIIFilter.cs
```diff
@@ -161,6 +158,10 @@ private string SanitizeUrlQueryPath(string url)
                 return url;
             }
 
+            // use the service provider to lazily get an IPermissionsStore instance
```
darrelmiller, line 161: I moved this after the URL check to avoid loading the uriTemplateMatcher from the cache if we are not going to use it.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -50,36 +48,68 @@ public class PermissionsStore : IPermissionsStore
         private const string PermissionsContainerBlobConfig = "BlobStorage:Containers:Permissions";
         private const string NullValueError = "Value cannot be null";
 
+
+        private class PermissionsDataInfo
```
darrelmiller, line 52: I had to create this wrapper class so that LoadPermissionsData could return both the UriTemplateMatcher and the ScopesList.  I want to avoid using class level variables because we are going to store this in the cache.  This avoids having to keep track of booleans to tell us if the cache is fresh, and it reduces multi-threading issues.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -50,36 +48,68 @@ public class PermissionsStore : IPermissionsStore
...
-                ?? throw new ArgumentNullException(nameof(permissionsCache), $"{ NullValueError }: { nameof(permissionsCache) }");
+               ?? throw new ArgumentNullException(nameof(configuration), $"{NullValueError}: {nameof(configuration)}");
+            _Cache = permissionsCache
+                ?? throw new ArgumentNullException(nameof(permissionsCache), $"{NullValueError}: {nameof(permissionsCache)}");
             _httpClientUtility = httpClientUtility
-                ?? throw new ArgumentNullException(nameof(httpClientUtility), $"{ NullValueError }: { nameof(httpClientUtility) }");
+                ?? throw new ArgumentNullException(nameof(httpClientUtility), $"{NullValueError}: {nameof(httpClientUtility)}");
             _fileUtility = fileUtility
-                ?? throw new ArgumentNullException(nameof(fileUtility), $"{ NullValueError }: { nameof(fileUtility) }");
+                ?? throw new ArgumentNullException(nameof(fileUtility), $"{NullValueError}: {nameof(fileUtility)}");
```
darrelmiller, line 75: Not sure why my Visual Studio removed these spaces,  Are we using an editor.config in this project?

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -50,36 +48,68 @@ public class PermissionsStore : IPermissionsStore
...
             _defaultRefreshTimeInHours = FileServiceHelper.GetFileCacheRefreshTime(configuration[CacheRefreshTimeConfig]);
 
-            InitializePermissions();
         }
 
+        private PermissionsDataInfo PermissionsData
+        {
+            get
+            {
+                if (!_Cache.TryGetValue<PermissionsDataInfo>("PermissionsData", out var permissionsData))
```
darrelmiller, line 90: If anyone tries to get access to the PermissionsData we go get the reference from the cache.  We don't hold it as a field.  If we can't get it from the cache, we go load the data and put that in a lock so that a second caller will get blocked until the load is complete.
If the data expires it will be flushed from cache within a minute.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -50,36 +48,68 @@ public class PermissionsStore : IPermissionsStore
...
+
         /// <summary>
         /// Populates the template table with the request urls and the scopes table with the permission scopes.
         /// </summary>
-        private void SeedPermissionsTables()
+        private PermissionsDataInfo LoadPermissionsData()
         {
-            _urlTemplateMatcher = new UriTemplateMatcher();
-            _scopesListTable = new Dictionary<int, object>();
+            var urlTemplateMatcher = new UriTemplateMatcher();
```
darrelmiller, line 111: Class level variables in singleton services are a real pain for multi-threaded apps.  Avoid them where you can.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -25,23 +25,21 @@ namespace PermissionsService
...
-            new() { { UtilityConstants.TelemetryPropertyKey_Permissions, nameof(PermissionsStore)} };
+            new() { { UtilityConstants.TelemetryPropertyKey_Permissions, nameof(PermissionsStore) } };
         private readonly string _permissionsContainerName;
         private readonly List<string> _permissionsBlobNames;
         private readonly string _scopesInformation;
         private readonly int _defaultRefreshTimeInHours; // life span of the in-memory cache
         private const string DefaultLocale = "en-US"; // default locale language
         private readonly object _permissionsLock = new();
         private readonly object _scopesLock = new();
-        private static bool _permissionsRefreshed = false;
```
darrelmiller, line 44: This isn't needed anymore because we always get the instance from the cache.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -25,23 +25,21 @@ namespace PermissionsService
 {
     public class PermissionsStore : IPermissionsStore
     {
-        private UriTemplateMatcher _urlTemplateMatcher;
-        private IDictionary<int, object> _scopesListTable;
-        private readonly IMemoryCache _permissionsCache;
+        
+        private readonly IMemoryCache _Cache;
```
darrelmiller, line 29: The cache can store anything we need, so let's just call it what it is, a cache.

PermissionsService/Services/PermissionsStore.cs
MaggieKimani1: Yes we are.

GraphWebApi/appsettings.json
```diff
@@ -14,8 +14,8 @@
   },
   "AllowedHosts": "*",
   "GraphMetadata": {
-    "V1.0": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsv1.0.xml",
-    "Beta": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_beta_metadata/cleanMetadataWithDescriptionsbeta.xml"
+    "V1.0": "https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsAndErrorsv1.0.xml",
```
irvinesunday, line 17: This CSDL version supersedes `https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsv1.0.xml`?
We currently use the latter in production.

PermissionsService/Services/PermissionsStore.cs
irvinesunday, line 29:
```suggestion
        private readonly IMemoryCache _cache;
```

baywet (dismissed):

GraphWebApi/appsettings.json
darrelmiller: I'm not actually sure.  I think @baywet added the errors to help with Kiota generation.  Either one is probably fine for GE.  PowerShell might not want the error class.  I'll change my PR to not use the one with errors to match what you are using in prod.

baywet (approved):

irvinesunday (approved):

PR #1156, opened 2022-08-24
Have OpenApiConvertSettings configurable for each style

### LINKED ISSUE #956: Have `OpenApiConvertSettings` configurable for each style
Different styles (Plain, PowerShell, PowerPlatform. GEAutocomplete) have different OpenAPI document style requirements. What works for one doesn't work for another. We need to make the `OpenApiConvertSettings` configurable in the `OpenAPIStyleOptions` class so that each style can customize the way it wants the conversion of OpenAPI configured. We would also need to consider the below:
- We need to keep in mind that this might mean having different caches of OpenAPI documents for the different styles.
- We need to keep track of stale OpenAPI document caches and how to refresh them.
- Can we move this functionality to the OpenAPI.NET/Hidi project so that it is re-usable there as well? 

Related to: https://github.com/microsoft/OpenAPI.NET/issues/830

### LINKED ISSUE #1047: Use regexes in config to remove certain OpenAPI paths for PowerShell style
We need to consider moving the function of removing the problematic paths affecting PowerShell cmdlet generation via AutoREST from XSLT to DevX API (specifically in the `PowerShellFormatter.cs` class in the OpenAPIService). These paths can be removed when formatting the subset OpenAPI document for PowerShell. 
We can make use of a config file (probably _appsettings.json_) with a list of the target paths (regex formatted) that we want removed specifically for PowerShell and use the regex to match and remove them in DevX API before serving the PowerShell-styled OpenAPI documents. Using this alternative, we can go ahead and remove the transforms that were added for these paths in the [preprocess_csdl.xsl](https://github.com/microsoftgraph/msgraph-metadata/blob/master/transforms/csdl/preprocess_csdl.xsl) file in the metadata repo and just continue using the single (unedited) CSDL file to generate OpenAPI documents for all the other styles in DevX API (GEAutocomplete, PowerPlatform and PowerShell). This document should have all the expected paths that were intentionally removed via XSLT (due to one reason or another). This config with a list of the regex-formatted paths can easily be updated from the app service configuration without requiring any re-deployment of the API.

millicentachieng:
## Overview
Closes #956
This PR:
- Reads convert settings specific to each Open API style e.g. PowerShell, PowerPlatform etc, from appsettings.json
- Apply settings when converting CSDL to Open API document

### DIFF
--- a/GraphWebApi/appsettings.json
+++ b/GraphWebApi/appsettings.json
@@ -50,6 +50,18 @@
     "RequestUrl": "https://cdn.graph.office.net/en-us/graph/api/proxy/endpoint",
     "Authorization": "ENTER_BEARER_TOKEN"
   },
+  "OpenAPI": {
+    "ConvertSettings": {
+      "PowerShell": {
+        "ShowLinks":  false,
+        "DeclarePathParametersOnPathItem": true,
+        "EnableODataAnnotationReferencesForResponses": false
+      },
+      "GEAutocomplete": {
+        "ShowRootPath": true
+      }
+    }
+  },
   "FileCacheRefreshTimeInHours": {
     "SampleQueries": "24",
     "Permissions": "24",
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -44,8 +44,8 @@ namespace OpenAPIService
 
     public class OpenApiService : IOpenApiService
     {
-        private static readonly ConcurrentDictionary<Uri, OpenApiDocument> _OpenApiDocuments = new();
-        private static readonly ConcurrentDictionary<Uri, DateTime> _OpenApiDocumentsDateCreated = new();
+        private static readonly ConcurrentDictionary<string, OpenApiDocument> _OpenApiDocuments = new();
+        private static readonly ConcurrentDictionary<string, DateTime> _OpenApiDocumentsDateCreated = new();
         private static readonly ConcurrentDictionary<string, string> _openApiTraceProperties =
                         new();
         private readonly TelemetryClient _telemetryClient;
@@ -64,6 +64,7 @@ namespace OpenAPIService
             _telemetryClient = telemetryClient;
             _openApiTraceProperties.TryAdd(UtilityConstants.TelemetryPropertyKey_OpenApi, nameof(OpenApiService));            
         }
+
         /// <summary>
         /// Create partial OpenAPI document based on the provided predicate.
         /// </summary>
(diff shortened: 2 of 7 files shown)


GraphWebApi/appsettings.json
```diff
@@ -50,6 +50,22 @@
     "RequestUrl": "https://cdn.graph.office.net/en-us/graph/api/proxy/endpoint",
     "Authorization": "ENTER_BEARER_TOKEN"
   },
+  "OpenAPI": {
+    "ConvertSettings": {
+      "PowerShell": {
```
baywet, line 55: It feels like some of those profiles should have more settings defined here, can you circle back with the different owners to understand better their settings requirements please?

OpenAPIService/OpenApiService.cs
```diff
@@ -738,6 +723,51 @@ public OpenApiDocument CloneOpenApiDocument(OpenApiDocument openApiDocument)
...
+                EnableDerivedTypesReferencesForRequestBody = false,
+                EnableDerivedTypesReferencesForResponses = false,
+                ShowRootPath = true,
+                ShowLinks = true,
+                ExpandDerivedTypesNavigationProperties = false
+            };
+
+            try
+            {
+                var convertSettingsForStyle = _configuration.GetSection($"OpenAPI:ConvertSettings:{style}")
```
baywet, line 747: use the configuration binder instead of building it manually.

https://github.com/microsoft/kiota/blob/3cc6161845aef7276de44757aead6e1820dc6599/src/kiota/Handlers/BaseKiotaCommandHandler.cs#L25

baywet (changes requested):

GraphWebApi/appsettings.json
irvinesunday: I know PowerShell does not like `ShowRootPath` and doesn't care about `ShowLinks`. These are GE-specific settings. And we def. want all styles to implement `ExpandDerivedTypesNavigationProperties=false` for now.
`DeclarePathParametersOnPathItem=true` and `EnableODataAnnotationReferencesForResponses=false` can be implemented for PowerShell.

millicentachieng: Thanks for the info @irvinesunday. I'll update the settings.

OpenAPIService/OpenApiService.cs
millicentachieng: @baywet I could bind the configuration to `OpenApiConvertSettings` but the class defines some default values and it would be impossible to tell which values came from the configuration.

baywet: If the default is acceptable for the profile, it's fine and no need to change it in the configuration. If the default is not suitable for the profile, just update the configuration. I'm not sure why we need to know where it was set from?

baywet: I'm not saying we should change that part of the logic.
You can perfectly bind the object once with the "global" entry of the configuration, and rebind it with the "styled" entry of the configuration to override some of the settings as needed and without having to implement that manually.

millicentachieng: I see. Thanks @baywet 😀

OpenAPIService/OpenApiService.cs
```diff
@@ -739,6 +723,29 @@ public OpenApiDocument CloneOpenApiDocument(OpenApiDocument openApiDocument)
             return doc;
         }
 
+        public OpenApiConvertSettings GetOpenApiConvertSettings(OpenApiStyle style = OpenApiStyle.Plain)
+        {
+            var globalConvertSettings = new OpenApiConvertSettings()
+            {
+                AddSingleQuotesForStringParameters = true,
```
baywet, line 730: I'd advocate to even have the baseline to the configuration file so we can change the settings without having to redeploy the code. And bind twice

baywet (changes requested):

baywet (approved):

PR #1197, opened 2022-09-27
Revert Microsoft.OpenApi.OData to 1.0.11

peombwa:
## Overview
This PR reverts `Microsoft.OpenApi.OData` to `1.0.11` due to a breaking change in `1.1.0`. See https://github.com/microsoft/OpenAPI.NET.OData/issues/282 for more info on the breaking change.


baywet:
I'd prefer to fail forward, adding a convert setting in the conversion lib, releasing a preview and using the preview + setting here. I believe some unit tests have been updated for other aspects with this newer version and I'd hate to have to roll everything back.

peombwa:
I agree with @baywet, let's work on a preview to add the option to the converter settings.

Closing PR.

PR #1233, opened 2022-11-07
Update pipeline tasks

MaggieKimani1:
Adds credscan and ESRP codesigning tasks to scan source code for committed credentials and signing .dlls


MaggieKimani1:
> Not sure we need code-sign since this is only going to be deployed to environments we own.

We're doing this as part of a security and compliance requirement for maintaining integrity of executables.

PR #1254, opened 2022-11-16
Start work on snippets generation (CLI)

calebkiage:
## Overview

Adds support for [msgraph-cli](https://github.com/microsoftgraph/msgraph-cli) code snippets generation

## Testing Instructions

* `dotnet test`

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/LanguageGenerators/GraphCliGenerator.cs
+++ b/CodeSnippetsReflection.OpenAPI/LanguageGenerators/GraphCliGenerator.cs
@@ -120,16 +120,35 @@ public partial class GraphCliGenerator : ILanguageGenerator<SnippetModel, OpenAp
         var matchingParams = operation.Parameters
                     .Where(p => p.In != ParameterLocation.Path && splitQueryString
                         .Any(s => s.Item1
-                            .Equals(p.Name, StringComparison.OrdinalIgnoreCase)));
+                            .Equals(p.Name, StringComparison.OrdinalIgnoreCase)))
+                    .Where(p => !string.IsNullOrWhiteSpace(p.Name));
 
         foreach (var param in matchingParams)
         {
-            AddParameterToDictionary(ref parameters, param.Name);
+            AddParameterToDictionary(ref parameters, $"{{{param.Name}}}");
         }
     }
 
     /// <summary>
     /// Adds a new parameter to the dictionary or replaces an existing one.
+    /// This function handles surrounding braces in the following ways:
+    /// <list type="number">
+    /// <item>
+    /// <description>
+    /// If the name has surrounding braces, then the dictionary entry's key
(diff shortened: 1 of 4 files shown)


CodeSnippetsReflection.OpenAPI/LanguageGenerators/GraphCliGenerator.cs
```diff
@@ -0,0 +1,254 @@
...
+
+        // Gets the request payload
+        var payload = GetRequestPayLoad(snippetModel);
+        if (!string.IsNullOrWhiteSpace(payload))
+        {
+            commandSegments.Add(payload);
+        }
+    }
+
+    private static void ProcessQueryParameters([NotNull] in SnippetModel snippetModel, [NotNull] in OpenApiOperation operation, [NotNull] ref Dictionary<string, string> parameters)
```
andrueastman, line 105: I believe it may be possible to take advantage of the `SnippetCodeGraph` to avoid duplication of the query parameter parsing logic shared by the other generators.

```cs
if(codeGraph.HasParameters())
{
      // process the parameters parsed ....
}
```
This will also give access to other properties like headers out of the box in a fashion similar to Kiota CodeDOM abstractions.

CodeSnippetsReflection.OpenAPI/LanguageGenerators/GraphCliGenerator.cs
```diff
@@ -0,0 +1,254 @@
...
+        {
+            paramName = paramName.Remove(0, 1);
+            addBraces.Item1 = false;
+        }
+
+        if (paramName.EndsWith('}'))
+        {
+            paramName = paramName.Remove(paramName.Length - 1, 1);
+            addBraces.Item2 = false;
+        }
```
andrueastman, line 162: It probably may be simpler to do
```cs
var paramName = name.TrimStart('{').TrimEnd('}');
var value = $"{{{name}}}";
```

calebkiage: I've had to also update the `ProcessQueryParameters` function as well but the resulting code is still simpler. Thanks

CodeSnippetsReflection.OpenAPI/LanguageGenerators/GraphCliGenerator.cs
calebkiage: I can add an issue for this. The reason is to give time to read the source and understand the API since the Snippet code graph isn't documented. I also notice that PowerShell doesn't use that API.

andrueastman: I believe the Powershell snippet generation may have come into play before the `SnippetCodeGraph`.  Will resolve this one so that we can add an issue to track.

andrueastman (approved):

Issue #1263, opened 2022-11-25
Investigate intermittent permission information unavailability issue


irvinesunday:
When permission for a given endpoint is requested, the `consentDisplayName` and `consentDescription` information gets shown intermittently, example: https://graphexplorerapi.azurewebsites.net/permissions?requestUrl=/groups&method=GET

![image](https://user-images.githubusercontent.com/40403681/203987526-7fd4e6df-bd8b-4ad5-a030-eba6cf1ff7eb.png)
_information unavailable_

 vs

![image](https://user-images.githubusercontent.com/40403681/203987607-9ea9941c-0b44-4fb7-b0b4-9944f320bc52.png)
_information available_

We need to investigate why this happens and fix this.

One area where we can immediately try and plug a potential issue is in making sure that we use the supported locale variants when creating in-memory caches for permissions i.e., `en-US`, `de-DE`, `ja-JP`, `zh-CN`, `pt-BR`, `ru-RU`, `es-ES`, `fr-FR`. Currently we create caches for any locale variant that we get from a client, even though we serve the supported permissions. This means we create several caches of different variants with potentially the same localized permissions.

Related to https://github.com/microsoftgraph/microsoft-graph-explorer-v4/issues/2243

irvinesunday:
Re-opening this issue as the issue still persists.

irvinesunday:
The issue doesn't get repro'd in the staging endpoint however, https://graphexplorerapi-staging.azurewebsites.net/permissions?requestUrl=/groups&method=GET and neither when tested locally. Technically both staging and prod have the same permissions code, with the exception of some of the telemetry and exception code removed in prod. with this [PR](https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1286)
Could this potentially be an infrastructure issue with prod.?

zengin:
quick note: this doesn't seem limited to the localized fields, `isAdmin` is also fetched as default `false` value of the boolean type in the error case, even though the expected value is `true`.

irvinesunday:
> quick note: this doesn't seem to be limited to the localized fields, `isAdmin` is also fetched as default `false` value of the boolean type in the error case, even though the expected value is `true`.

That's because all properties are set to their defaults when there is no scopes information data
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/a85d5cc01d0f8591d0712fce88106ba63be94689/PermissionsService/Models/ScopeInformation.cs#L14-L21

[...]

PR #1302, opened 2022-12-22
additional telemetry to diagnose permission caching issue

zengin:
## Overview

- Additional telemetry to diagnose whether seeding JSON object is not read correctly.
- Additional telemetry to watch changing dictionary location
- partial #1263

### DIFF
--- a/PermissionsService/Services/PermissionsStore.cs
+++ b/PermissionsService/Services/PermissionsStore.cs
@@ -316,6 +316,18 @@ namespace PermissionsService
                                          SeverityLevel.Information,
                                          _permissionsTracePropertiesWithSanitizeIgnore);
 
+            if (delegatedScopesInfoTable.Count == 0 && applicationScopesInfoTable.Count == 0)
+            {
+                // sneak-peak into the scopes JSON if we didn't get any scopes
+                const int maxCharsToLog = 100;
+                var scopesInfoJsonToLog = scopesInfoJson[0..int.Max(maxCharsToLog, scopesInfoJson.Length)]
+                                                .Replace("\r", string.Empty, StringComparison.OrdinalIgnoreCase)
+                                                .Replace("\n", string.Empty, StringComparison.OrdinalIgnoreCase);
+                _telemetryClient?.TrackTrace($"Scopes information is empty. ScopesInfoJson start: {scopesInfoJsonToLog}",
+                                             SeverityLevel.Error,
+                                             _permissionsTracePropertiesWithSanitizeIgnore);
+            }
+
             return new Dictionary<string, IDictionary<string, ScopeInformation>>
             {
                 { Delegated, delegatedScopesInfoTable },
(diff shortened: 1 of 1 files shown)


PermissionsService/Services/PermissionsStore.cs
```diff
@@ -302,6 +302,11 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
                                          _permissionsTraceProperties);
 
             var scopesInformationList = JsonConvert.DeserializeObject<ScopesInformationList>(scopesInfoJson);
+
+            _telemetryClient?.TrackTrace($"DelegatedScopesList count: {scopesInformationList.DelegatedScopesList.Count}, ApplicationScopesList count: {scopesInformationList.ApplicationScopesList.Count}",
```
irvinesunday, line 306: We already have telemetry to capture the count: https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/7a07d785cf323a913ef1e3bd690f97b88fbe7bb9/PermissionsService/Services/PermissionsStore.cs#L308-L312

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -302,6 +302,11 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
...
+                                         SeverityLevel.Information,
+                                         _permissionsTraceProperties);
```
irvinesunday, line 308: Also, this will sanitize the telemetry data and remove the numerical values.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -493,12 +510,24 @@ private static string CleanRequestUrl(string requestUrl)
...
+                }
+                else
+                {
+                    scopeInfoMissing = true;
+                    return new() { ScopeName = scope };
+                }
             }).ToList();
 
+            _telemetryClient?.TrackTrace($"scopesInformationDictionary hash code is: {scopesInformationDictionary.GetHashCode()}",
+                                                scopeInfoMissing ? SeverityLevel.Error : SeverityLevel.Information,
```
irvinesunday, line 528: Since when we send a request to [/permissions](https://graphexplorerapi.azurewebsites.net/permissions) to fetch all permissions there are always some scopes with no information, this telemetry will always be logging the information as an error. Perhaps we just want to capture the info as error when requesting for permissions for single urls. If so, then we can we add `(scopeInfoMissing && !getAllPermissions)` to this conditional.

PermissionsService/Services/PermissionsStore.cs
zengin: This was to capture if for some reason list to dictionary conversion was failing.

PermissionsService/Services/PermissionsStore.cs
zengin: makes sense, thanks for the catch!

PermissionsService/Services/PermissionsStore.cs
zengin: Thanks for the context, I updated with `SanitizeIgnore`

PermissionsService/Services/PermissionsStore.cs
zengin: Please let me know if you still want me to remove.

MaggieKimani1 (approved):

Issue #1350, opened 2023-02-01
Incorrect generation of PowerShell snippets that have nested values with 'OR' separators


timayabi2020:
**Describe the bug**
PowerShell snippets with 'OR' separators are incorrectly generated making the separator appear as a positional separator during execution. [Reference](https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1770)

**To Reproduce**
1. Go to [GE](https://developer.microsoft.com/en-us/graph/graph-explorer)
2. Input this request URL "https://graph.microsoft.com/v1.0/users?$search="displayName%3Adi" OR "displayName%3Aal"&$count=true&$orderby=displayname"
3. Login to a Tenant and run query
4. Click on code snippets and select powershell
5. Get the snippet and run it on a powershell session ( Ensure that you run ```Connect-MgGraph``` first to connect to a Tenant)
6. You will get the below error
<img width="849" alt="image" src="https://user-images.githubusercontent.com/10947120/216078763-66589df0-30c6-4cc1-bf38-f105e0865fe4.png">

**Expected behavior**
"+OR+" should not appear as a positional parameter when the snippet was generated.
The snippet should be generated and successfully executed as below.
<img width="861" alt="image" src="https://user-images.githubusercontent.com/10947120/216079737-0500e5b8-41df-4f36-842c-ecf5b50a5893.png">

PR #1351, opened 2023-02-01
Resolve PowerShell Snippet 'OR' positional parameter issue

timayabi2020:
## Overview

This PR prevents the OR separator inside parameter values from appearing as a positional parameter when executing a PowerShell snippet. The issue is described [here](https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1350).

### Demo
image

## Testing Instructions

* Check out the branch associated with this PR.
* Build and run 'GraphWebApi' project
* Use [this postman collection](https://api.postman.com/collections/3669426-ac10c1c8-d60f-43b5-8d1e-3f9a14a8b90c?access_key=PMAT-01GR6SKZW1HCKYN490M9ZH7CB3) to run the request.
* Copy and execute the snippet in a PowerShell session
* image

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/LanguageGenerators/PowerShellGenerator.cs
+++ b/CodeSnippetsReflection.OpenAPI/LanguageGenerators/PowerShellGenerator.cs
@@ -176,13 +176,21 @@ namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators
                 return originalValue.ToLowerInvariant();
             else if (int.TryParse(originalValue, out var intValue))
                 return intValue.ToString();
-            else {
+            else if (originalValue.Contains("+OR+"))
+                return NestedValueWithORSeparator(originalValue);
+            else
+            {
                 var valueWithNested = originalValue.Split(',')
                                                     .Select(v => replacements.ContainsKey(v) ? v + replacements[v] : v)
                                                     .Aggregate((a, b) => $"{a},{b}");
                 // Replace '$' with '`$' since '$' is a reserved character in powershell.
                 return $"\"{valueWithNested.Replace("$", "`$")}\"";
             }
+        }
+        private static string NestedValueWithORSeparator(string originalValue)
+        {
+            originalValue = originalValue.Replace("+", " ");
+            return $"'\"{originalValue.Replace("$", "`$")}\"'";
         }
         private static string NormalizeQueryParameterName(string queryParam)
         {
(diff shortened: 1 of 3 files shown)


CodeSnippetsReflection.OpenAPI/LanguageGenerators/PowerShellGenerator.cs
```diff
@@ -176,14 +176,22 @@ private static string GetQueryParameterValue(string normalizedParameterName, str
                 return originalValue.ToLowerInvariant();
             else if (int.TryParse(originalValue, out var intValue))
                 return intValue.ToString();
-            else {
+            else if (originalValue.Contains("+OR+"))
```
andrueastman, line 179: I think the right thing to do here is simply to match the search parameter and surround the original value with single quotes. 
```cs
            if (normalizedParameterName.Equals("Search"))
                return $"'{originalValue.Replace("+", " ")}'";
```

This is because the http sample could change with different operators like AND which would not work with the hardcoded operator. 
https://learn.microsoft.com/en-us/graph/search-query-parameter?tabs=http#using-search-on-message-collections

andrueastman (commented):
Any chance we can add a test like this to validate the changes?

https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/3b086c7d115fb507af8559f55facbf34745b912c/CodeSnippetsReflection.OpenAPI.Test/PowerShellGeneratorTests.cs#L236

I suspect the assumption in the test is also incorrect as the double quotes may not be passed over to the request on execution as they are needed.

timayabi2020:
> Any chance we can add a test like this to validate the changes?
> 
> https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/3b086c7d115fb507af8559f55facbf34745b912c/CodeSnippetsReflection.OpenAPI.Test/PowerShellGeneratorTests.cs#L236
> 
> I suspect the assumption in the test is also incorrect as the double quotes may not be passed over to the request on execution as they are needed.

I have added tests to validate the search query options scenarios and they are all passing and the generated snippets are executable on a PowerShell session.

timayabi2020:
> > Any chance we can add a test like this to validate the changes?
> > https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/3b086c7d115fb507af8559f55facbf34745b912c/CodeSnippetsReflection.OpenAPI.Test/PowerShellGeneratorTests.cs#L236
> > 
> > I suspect the assumption in the test is also incorrect as the double quotes may not be passed over to the request on execution as they are needed.
> 
> I have added tests to validate the search query options scenarios and they are all passing and the generated snippets are executable on a PowerShell session.

There is assumption is Ok, If you execute the command with debug option you'll see that the double quotes are being passed

andrueastman (dismissed):

andrueastman (approved):

PR #1380, opened 2023-02-21
Update PHP Snippet generation.

SilasKenneth:
## Overview

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/CodeSnippetsReflection.OpenAPI.Test/PhpGeneratorTests.cs
+++ b/CodeSnippetsReflection.OpenAPI.Test/PhpGeneratorTests.cs
@@ -83,8 +83,8 @@ public class PhpGeneratorTests
             };
         var snippetModel = new SnippetModel(requestPayload, ServiceRootUrl, await GetV1TreeNode());
         var result = _generator.GenerateCodeSnippet(snippetModel);
-        Assert.Contains("$requestConfiguration = new UsersRequestBuilderPostRequestConfiguration(\n", result);
-        Assert.Contains("queryParameters: UsersRequestBuilderPostRequestConfiguration::addQueryParameters(", result);
+        Assert.Contains("$requestConfiguration = new UsersRequestBuilderPostRequestConfiguration();", result);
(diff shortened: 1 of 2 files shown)


Ndiritu (commented):
Would it confuse customers if query parameters snippets only work with PHP 8 while the others work across both PHP 7 and 8?

SilasKenneth:
> Would it confuse customers if query parameters snippets only work with PHP 8 while the others work across both PHP 7 and 8?

That makes sense. What about having something like this? Knowing these other query parameters that are supposed to be set to null is the problem, at least on the devx side

```php
...addQueryParameters(
null,
null,
'age eq 20',
['email', 'principalName']);

```
Won't that be confusing, too?

Ndiritu:
> 

I think for PHP 7 devs the existing snippets seem like the best experience. PHP 8, constructors would be ideal.
Unless we add both snippets? Or only add a comment in the snippet about a better experience for PHP 8 using the constructors?

CodeSnippetsReflection.OpenAPI.Test/PhpGeneratorTests.cs
```diff
@@ -83,8 +83,8 @@ public async Task GeneratesObjectInitializationWithCallToSetters()
             };
         var snippetModel = new SnippetModel(requestPayload, ServiceRootUrl, await GetV1TreeNode());
         var result = _generator.GenerateCodeSnippet(snippetModel);
-        Assert.Contains("$queryParameters = new UsersRequestBuilderPostQueryParameters();", result);
-        Assert.Contains("$requestConfiguration->queryParameters = $queryParameters;", result);
+        Assert.Contains("$requestConfiguration = new UsersRequestBuilderPostRequestConfiguration();", result);
+        Assert.Contains("$queryParameters = UsersRequestBuilderPostRequestConfiguration::createQueryParameters();", result);
```
Ndiritu, line 87: 👍🏾

PR #1388, opened 2023-02-23
[dotnet] Removes curly braces from generated snippets

andrueastman:
Updates .Net snippet generation to fix for Odata functions with no function parameters to remove the curly braces.

previously
```cs
await graphClient.Users.MicrosoftGraphDelta().GetAsync();
```

which should now be
```cs
await graphClient.Users.MicrosoftGraphDelta.GetAsync();
```


baywet (approved):

baywet:
@andrueastman this is probably an issue for java, python, php, typescript (but not go) too, would you mind addressing it for those languages as well please while at it?

andrueastman:
> python

No worries. Will create a separate PR to resolve this soon.

andrueastman:
Just to follow up here @baywet

PHP uses the curly needs the braces - https://github.com/microsoftgraph/msgraph-sdk-php/blob/feat/kiota-preview/src/Generated/Users/UsersRequestBuilder.php#L38

As well as Typescript - https://github.com/microsoftgraph/msgraph-sdk-typescript/blob/d27d32590585c2d6e7e2e76df5e9a400b3878a46/src/users/usersRequestBuilder.ts#L24

We still need to build generators for Java and Python. So, we should be okay for now.

PR #1393, opened 2023-02-28
Feature: change open api gen process

### LINKED ISSUE #1392: Use pre-generated OpenAPI files

thewahome:
## Overview

Closes #1392 

### DIFF
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -55,6 +55,7 @@ namespace OpenAPIService
         private readonly int _defaultForceRefreshTime; // time span for allowable forceRefresh of the OpenAPI document
         private readonly Queue<string> _graphUriQueue = new();
         private static readonly RecyclableMemoryStreamManager _streamManager = new();
+        Dictionary<OpenApiStyle, string> fileNames = new Dictionary<OpenApiStyle, string>();
 
         public OpenApiService(IConfiguration configuration, TelemetryClient telemetryClient = null)
         {
@@ -62,7 +63,12 @@ namespace OpenAPIService
                ?? throw new ArgumentNullException(nameof(configuration), $"Value cannot be null: {nameof(configuration)}");
             _defaultForceRefreshTime = FileServiceHelper.GetFileCacheRefreshTime(_configuration[CacheRefreshTimeConfig]);
             _telemetryClient = telemetryClient;
-            _openApiTraceProperties.TryAdd(UtilityConstants.TelemetryPropertyKey_OpenApi, nameof(OpenApiService));            
+            _openApiTraceProperties.TryAdd(UtilityConstants.TelemetryPropertyKey_OpenApi, nameof(OpenApiService));
+
+            fileNames.Add(OpenApiStyle.GEAutocomplete, "graphexplorer");
+            fileNames.Add(OpenApiStyle.Plain, "openapi");
+            fileNames.Add(OpenApiStyle.PowerShell, "powershell");
+            fileNames.Add(OpenApiStyle.PowerPlatform, "openapi");
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


baywet (dismissed):
My only concern is if somebody else than us has taken a dependency on the API /enum values as changing those will break them. Trusting this will at least be coordinated with GE if there are direct dependencies. Otherwise we might want to consider a compatibility mapping.

CC @estruyf if you still have a dependency on devx api for your extensions

estruyf:
@baywet indeed, it is used in the following extension: https://marketplace.visualstudio.com/items?itemName=eliostruyf.vscode-msgraph-autocomplete - one of your colleagues is working on this extension to allow snippets to work as well.

thewahome:
Aah makes sense @baywet ... I will just make a dictionary to map the enum values to the file names ..

baywet:
Thanks for reverting that fast @estruyf It seems the extension is relying on a static zip file instead of calling out to the devx API, did I get this right? https://github.com/estruyf/vscode-msgraph-autocomplete/blob/main/apis.json.gz

estruyf:
@baywet yes, that is the cached version of the API, so that it doesn't always have to call the API, but if it doesn't find it, or there is a new endpoint, it will still call the API.

https://github.com/estruyf/vscode-msgraph-autocomplete/blob/ced76a29162db1f1089f009db3cec95f98c4f8ae/src/utils/AutoComplete.ts#L4

baywet:
Ah great, you're not making use of the style parameter and anyway we're going to be reverting the changes to the API surface. Thanks again for all that input!

OpenAPIService/OpenApiService.cs
```diff
@@ -55,14 +55,20 @@ public class OpenApiService : IOpenApiService
         private readonly int _defaultForceRefreshTime; // time span for allowable forceRefresh of the OpenAPI document
         private readonly Queue<string> _graphUriQueue = new();
         private static readonly RecyclableMemoryStreamManager _streamManager = new();
+        Dictionary<OpenApiStyle, string> fileNames = new Dictionary<OpenApiStyle, string>();
```
baywet, line 58:
```suggestion
        private static readonly Dictionary<OpenApiStyle, string> fileNames = new Dictionary<OpenApiStyle, string>() {
            { OpenApiStyle.GEAutocomplete, "graphexplorer"},
            { OpenApiStyle.Plain, "openapi"},
            { OpenApiStyle.PowerShell, "powershell"},
            { OpenApiStyle.PowerPlatform, "openapi"},
         };
```

OpenAPIService/OpenApiService.cs
```diff
@@ -55,14 +55,20 @@ public class OpenApiService : IOpenApiService
...
                ?? throw new ArgumentNullException(nameof(configuration), $"Value cannot be null: {nameof(configuration)}");
             _defaultForceRefreshTime = FileServiceHelper.GetFileCacheRefreshTime(_configuration[CacheRefreshTimeConfig]);
             _telemetryClient = telemetryClient;
-            _openApiTraceProperties.TryAdd(UtilityConstants.TelemetryPropertyKey_OpenApi, nameof(OpenApiService));            
+            _openApiTraceProperties.TryAdd(UtilityConstants.TelemetryPropertyKey_OpenApi, nameof(OpenApiService));
+
+            fileNames.Add(OpenApiStyle.GEAutocomplete, "graphexplorer");
+            fileNames.Add(OpenApiStyle.Plain, "openapi");
+            fileNames.Add(OpenApiStyle.PowerShell, "powershell");
+            fileNames.Add(OpenApiStyle.PowerPlatform, "openapi");
```
baywet, line 71:
```suggestion
```

baywet (changes requested):

baywet (dismissed):

andrueastman (approved):

Issue #1355, opened 2023-02-02
[Feature Request] DevX API should return permission types in the permissions response object


peombwa:
DevX API should return `PermissionType`\ `ScopeType` as a property in the permissions response object. The `PermissionType` should have a value of `DelegatedWork`, `DelegatedPersonal`, and `Application`. Something similar to `ScopeType`.

The current permissions response only returns the properties declared in `ScopeInformation` model class https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/3b086c7d115fb507af8559f55facbf34745b912c/PermissionsService/Models/ScopeInformation.cs#L14-L21

`ScopeInformation` should have property that holds a collection of `PermissionType`.

Issue #1318, opened 2023-01-18
Update permissions endpoint to add support for returning permissions for all scope types for a given endpoint


irvinesunday:
Currently, when requesting for permissions, the client has to provide a value for the `scopeType` in the query parameter. If none is provided, the default `DelegatedWork` is provided. This means it is currently not possible for a client to retrieve all the scope types (_DelegatedPersonal,_ _DelegatedWork_, _Application_) in one call.
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/fa0aca54e933b54f68f0259d26065e3974540fef/GraphWebApi/Controllers/PermissionsController.cs#L42-L46
We need to update the query parameter to not have a default value and add support for clients requiring retrieval of all the scope types for a given endpoint.

adhiambovivian:
Related to #1337

Issue #1253, opened 2022-11-16
Hidden permissions on Graph Explorer are not hidden


RabebOthmani:
The list of permissions Graph Explorer is consuming, is not filtered resulting in showing permissions publicly that should not be shown.

PR #1402, opened 2023-03-03
Support processing of permission file in new format

### LINKED ISSUE #1261: Convert the current permissions model being used by DevX API to the new permissions model from Kibali.
the current permissions model here(https://github.com/microsoftgraph/microsoft-graph-devx-content/tree/dev/permissions) should be converted to the new model that looks like this (https://github.com/microsoftgraph/kibali/blob/main/test/kibaliTests/GraphPermissions.json).

This will allow us to see how DevX API will handle new permissions when Identity team start using the new model. It will also help us start tagging hidden permissions as so and thus filter them out to avoid displaying them on GE

millicentachieng:
## Overview

Fixes #1261, #1253, #1318, #1337, #1338, #1355, 
This PR
- Updates DevX API to use permissions file in new format
- Updates permissions endpoint to add support for returning permissions for all scope types for a given endpoint
- Updates DevX API to return scope types in the permissions response object
- Updates DevX API to return a payload indicating which permission has the least privilege
- Updates permissions endpoint to add support for filtering out hidden permissions
- Adds an endpoint that to return least privilege permissions given a set of URLs


GraphWebApi/wwwroot/OpenApi.yaml
```diff
@@ -256,17 +262,128 @@ paths:
...
                   consentDescription: "Allows the app to read email in your mailbox. "
                   isAdmin: false
+                  isLeastPrivilege: false
                 - value: Mail.ReadWrite
+                  scopeType: Application
                   consentDisplayName: "Read and write access to your mail "
                   consentDescription: "Allows the app to read, update, create and delete email in your mailbox. Does not include permission to send mail. "
                   isAdmin: false
+                  isLeastPrivilege: false
+        "207":
```
baywet, line 365: 207 is a webdav status code. Why do we need both the scopes and the errors at the same time? shouldn't it be one or the other?

PermissionsService/Interfaces/IPermissionStore.cs
```diff
@@ -17,19 +17,40 @@ public interface IPermissionsStore
...
+        /// </summary>
+        /// <param name="requestUrls">The list of request URLs to fetch permissions for.</param>
+        /// <param name="scopeType">Optional: The type of scope to be retrieved for the target request url.</param>
+        /// <param name="locale">Optional: The language code for the preferred localized file.</param>
+        /// <param name="org">Optional: The name of the org/owner of the repo.</param>
+        /// <param name="branchName">Optional: The name of the branch containing the files</param>
+        /// <param name="includeHidden">Optional: Whether to include hidden permissions or not: Defaults to false.</param>
+        /// <param name="isLeastPrivilege">Optional: Whether to only return least privilege permissions on not. Defaults to true.</param>
+        /// <returns></returns>
+        Task<PermissionResult> GetScopesForMultipleResourcesAsync(List<string> requestUrls = null,
```
baywet, line 47: instead of introducing a new method, since we own both the interface and the implementation, why don't we evolve the existing one?

PermissionsService/Models/PathPermissions.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace PermissionsService.Models
+{
+    internal class PathPermissions
+    {
+        public string Url
```
baywet, line 8:
```suggestion
        public required string Url
```

(this way you can get rid of the constructor)

PermissionsService/Models/PathSet.cs
```diff
@@ -0,0 +1,132 @@
...
+        [JsonPropertyName("paths")]
+        [JsonConverter(typeof(PathsLeastPrivilegeSchemeKeyConverter))]
+        public Dictionary<string, List<ScopeType>> Paths
+        {
+            get; set;
+        } = new();
+    }
+
+
+    internal class PathsLeastPrivilegeSchemeKeyConverter : JsonConverter<Dictionary<string, List<ScopeType>>>
```
baywet, line 47: why do we need custom converters, couldn't we adapt our models to avoid needing them?

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -94,79 +95,108 @@ private PermissionsDataInfo PermissionsData
...
                         }
                     }
                 }
             }
 
-            _telemetryClient?.TrackTrace("Finished seeding permissions tables",
-                                         SeverityLevel.Information,
-                                         _permissionsTraceProperties);
+            var hiddenPermissions = permissionsDocument.Permissions
+                .Where(x => x.Value.ProvisioningInfo.IsHidden).Select(x => x.Key).ToHashSet();
```
baywet, line 193:
```suggestion
                .Where(static x => x.Value.ProvisioningInfo.IsHidden).Select(static x => x.Key).ToHashSet(StringComparer.OrdinalIgnoreCase);
```

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -319,111 +349,147 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
...
+                                                           string branchName = null,
+                                                           bool includeHidden = false,
+                                                           bool isLeastPrivilege = true)
+        {
+            if (PermissionsData?.PathPermissions == null)
+                throw new InvalidOperationException("Failed to fetch permissions");
 
-                scopesInfo = GetScopesInformation(scopesInformationDictionary, scopes, scopeType, true);
+            // Create a dict of scopes information from GitHub files or cached files
+            var scopesInformationDictionary = (!string.IsNullOrEmpty(org) && !string.IsNullOrEmpty(branchName))
```
baywet, line 396:
```suggestion
            var scopesInformationDictionary = !(string.IsNullOrEmpty(org) || string.IsNullOrEmpty(branchName))
```

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -319,111 +349,147 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
...
+                Errors = errors
+            };
+        }
 
-                if (permissionsData.ScopesListTable[key] is not JToken resultValue)
+        private IEnumerable<ScopeInformation> GetAllScopes(ScopeType? scopeType, bool includeHidden)
+        {
+            var scopes = PermissionsData.PathPermissions.Values
+                .SelectMany(x => x.SupportedMethodPermissions.Values)
+                .SelectMany(x => x)
```
baywet, line 438:
```suggestion
                .SelectMany(static x => x.SupportedMethodPermissions.Values)
                .SelectMany(static x => x)
```

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -319,111 +349,147 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
...
+                .Where(x => scopeType == null || x.Key == scopeType)
+                .SelectMany(x => x.Value.AllPermissions, (x, permission) => new ScopeInformation
```
baywet, line 440:
```suggestion
                .SelectMany(static x => x.Value.AllPermissions, static (x, permission) => new ScopeInformation
```

baywet (changes requested):

GraphWebApi/wwwroot/OpenApi.yaml
millicentachieng: Noted. I've updated the status code to 400. We need to let the client know that there were some URLs that did not yield  any scopes and they can proceed with the result if they choose to.

PermissionsService/Interfaces/IPermissionStore.cs
millicentachieng: Done. I've merged the two methods.

PermissionsService/Models/PathSet.cs
millicentachieng: I've removed the custom converters.

GraphWebApi/wwwroot/OpenApi.yaml
```diff
@@ -745,18 +859,24 @@ components:
...
+        - scopeType
         - consentDisplayName
         - consentDescription
         - isAdmin
+        - isLeastPrivilege
       properties:
         value:
           type: string
+        scopeType:
+          type: string
```
baywet, line 871: this could use an enum type instead

GraphWebApi/wwwroot/OpenApi.yaml
```diff
@@ -256,23 +262,130 @@ paths:
...
+      operationId: permissions.ListPermissionsSet
+      parameters:
+        - name: scopeType
+          in: query
+          schema:
+            type: string
+            enum:
+              - DelegatedWork
+              - DelegatedPersonal
+              - Application
```
baywet, line 302: why not move this definition to the schema + parameters components and refer to it here instead to enable reusability?

GraphWebApi/wwwroot/OpenApi.yaml
```diff
@@ -256,23 +262,130 @@ paths:
...
+          required: false
+          description: If set to true, allows fetching of any available hidden permissions
+          default: false
+        - name: isLeastPrivilege
+          in: query
+          schema:
+            type: boolean
+          required: false
+          description: If set to true, fetches minimum permissions for a set of resources
+          default: true
```
baywet, line 330: parameters components for all of them so they can be reused between the get and post operations

PermissionsService/Models/PathPermissions.cs
```diff
@@ -0,0 +1,31 @@
...
+            get; set;
+        } = new(StringComparer.OrdinalIgnoreCase);
+    }
+
+    internal class SchemePermissions
+    {
+        public HashSet<string> LeastPrivilegePermissions
+        {
+            get; set;
+        } = new();
```
baywet, line 24:
```suggestion
        } = new(StringComparer.OrdinalIgnoreCase);
```

PermissionsService/Models/PathPermissions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public HashSet<string> AllPermissions
+        {
+            get; set;
+        } = new();
```
baywet, line 29:
```suggestion
        } = new(StringComparer.OrdinalIgnoreCase);
```

PermissionsService/Models/PathSet.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace PermissionsService.Models
+{
+    internal class PathSet
+    {
+        [JsonPropertyName("schemeKeys")]
+        public HashSet<string> SchemeKeys
+        {
+            get; set;
+        } = new();
```
baywet, line 13:
```suggestion
        } = new(StringComparer.OrdinalIgnoreCase);
```

PermissionsService/Models/PathSet.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        [JsonPropertyName("methods")]
+        public HashSet<string> Methods
+        {
+            get; set;
+        } = new(StringComparer.OrdinalIgnoreCase);
+
+        [JsonPropertyName("excludedProperties")]
+        public List<string> ExcludedProperties
```
baywet, line 22: any specific reason why we're using a list and not a hashset for those two?

PermissionsService/Models/PermissionResult.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Net;
+
+namespace PermissionsService.Models
+{
+    public class PermissionResult
+    {
+        public List<ScopeInformation> Results
+        {
+            get; set;
+        }
```
baywet, line 11:
```suggestion
        } = new();
```

PermissionsService/Models/PermissionResult.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public List<PermissionError> Errors
+        {
+            get; set;
+        }
```
baywet, line 16:
```suggestion
        } = new();
```

PermissionsService/Models/PermissionResult.cs
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public class PermissionError
+    {
+        public string Url
+        {
+            get; set;
+        }
```
baywet, line 24:
```suggestion
        } = string.Empty;
```

PermissionsService/Models/PermissionResult.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public string Message
+        {
+            get; set;
+        }
```
baywet, line 29:
```suggestion
        } = string.Empty;
```

Should we decide to enable NRT for this project, we won't have to revisit this class this way

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -47,23 +45,27 @@ public class PermissionsStore : IPermissionsStore
...
+            public IDictionary<string, PathPermissions> PathPermissions
             {
-                get;set;
+                get; set;
             }
 
+            public HashSet<string> HiddenPermissions
+            {
+                get; set;
+            }
```
baywet, line 66: set default values?

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -94,79 +93,121 @@ private PermissionsDataInfo PermissionsData
                 {
                     lock (_permissionsLock)
                     {
-                        permissionsData = LoadPermissionsData();
-                        _cache.Set("PermissionsData", permissionsData, new MemoryCacheEntryOptions
+                        permissionsData = LoadPermissionsDataAsync().Result;
```
baywet, line 96: instead of using a lock and blocking everyone, use [GetOrCreateAsync](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.cacheextensions.getorcreateasync?view=dotnet-plat-ext-7.0) and change the property type to be a Task<PermissionsDataInfo> so everyone can await, which will be deferred in most cases and avoid blocking everything, everywhere.

baywet (changes requested):

PermissionsService/Models/PathSet.cs
millicentachieng: Updating to HashSet

PermissionsService/Models/ScopeType.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
```
irvinesunday, line 1: Remove unnecessary usings

irvinesunday: Add license info.

PermissionsService/Models/ScopeType.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace PermissionsService.Models
+{
+    public enum ScopeType
```
irvinesunday, line 9: Include public XML summary

GraphWebApi/Controllers/PermissionsController.cs
```diff
@@ -39,53 +42,78 @@ public PermissionsController(IPermissionsStore permissionsStore, TelemetryClient
...
+                                                             [FromQuery]string branchName = null,
+                                                             [FromQuery]bool includeHidden = false,
+                                                             [FromQuery]bool isLeastPrivilege = false)
         {
-            string localeCode = RequestHelper.GetPreferredLocaleLanguage(Request) ?? Constants.DefaultLocale;
-            _telemetryClient?.TrackTrace($"Request to fetch permissions for locale '{localeCode}'",
-                                            SeverityLevel.Information,
-                                            _permissionsTraceProperties);
+            if (!string.IsNullOrEmpty(requestUrl) && string.IsNullOrEmpty(method))
+                return BadRequest("The HTTP method value cannot be null or empty.");
```
irvinesunday, line 54: Exception message does not indicate whether the `requestUrl` could be null or empty

GraphWebApi/Controllers/PermissionsController.cs
```diff
@@ -39,53 +42,78 @@ public PermissionsController(IPermissionsStore permissionsStore, TelemetryClient
...
-                                                             [FromQuery]string branchName = null)
+                                                             [FromQuery]string branchName = null,
+                                                             [FromQuery]bool includeHidden = false,
+                                                             [FromQuery]bool isLeastPrivilege = false)
         {
-            string localeCode = RequestHelper.GetPreferredLocaleLanguage(Request) ?? Constants.DefaultLocale;
-            _telemetryClient?.TrackTrace($"Request to fetch permissions for locale '{localeCode}'",
-                                            SeverityLevel.Information,
-                                            _permissionsTraceProperties);
+            if (!string.IsNullOrEmpty(requestUrl) && string.IsNullOrEmpty(method))
```
irvinesunday, line 53: What if the `requestUrl` is empty? Do we want to throw an error?

GraphWebApi/Controllers/PermissionsController.cs
```diff
@@ -39,53 +42,78 @@ public PermissionsController(IPermissionsStore permissionsStore, TelemetryClient
...
+            return result == null || !result.Results.Any() ? NotFound() : Ok(result.Results);
+        }
 
-            if (!string.IsNullOrEmpty(org) && !string.IsNullOrEmpty(branchName))
-            {
-                // Fetch permissions descriptions file from Github
-                result = await _permissionsStore.GetScopesAsync(scopeType: scopeType,
+        [HttpPost]
+        [Produces("application/json")]
+        public async Task<IActionResult> GetPermissionScopes([FromBody] List<string> requestUrls,
```
irvinesunday, line 78: Check against nullability of `requestUrls`.

GraphWebApi/Controllers/PermissionsController.cs
```diff
@@ -39,53 +42,78 @@ public PermissionsController(IPermissionsStore permissionsStore, TelemetryClient
...
+                                                             [FromQuery] ScopeType? scopeType = null,
```
irvinesunday, line 79: Same as above, check for nullability.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -94,79 +93,121 @@ private PermissionsDataInfo PermissionsData
...
             {
-                _telemetryClient?.TrackTrace($"Seeding permissions table from file source '{permissionFilePath}'",
+                var permissionsSourceUrl = _configuration.GetValue<string>("PermissionsSourceUrl");
+                _telemetryClient?.TrackTrace($"Fetching permissions from file source '{permissionsSourceUrl}'",
                                              SeverityLevel.Information,
                                              _permissionsTracePropertiesWithSanitizeIgnore);
 
-                string relativePermissionPath = FileServiceHelper.GetLocalizedFilePathSource(_permissionsContainerName, permissionFilePath);
-                string jsonString = _fileUtility.ReadFromFile(relativePermissionPath).GetAwaiter().GetResult();
+                using var client = new HttpClient();
```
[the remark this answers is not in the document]
irvinesunday: Or use the already declared http client private field.
```CSharp 
private readonly IHttpClientUtility _httpClientUtility;
```

GraphWebApi/Controllers/PermissionsController.cs
millicentachieng: If `requestUrl` is null or empty then we return all permissions. We won't need to throw an error.

GraphWebApi/Controllers/PermissionsController.cs
```diff
@@ -39,53 +42,78 @@ public PermissionsController(IPermissionsStore permissionsStore, TelemetryClient
         // Gets the permissions scopes
         [HttpGet]
         [Produces("application/json")]
-        public async Task<IActionResult> GetPermissionScopes([FromQuery]string scopeType = "DelegatedWork",
+        public async Task<IActionResult> GetPermissionScopes([FromQuery]ScopeType? scopeType = null,
```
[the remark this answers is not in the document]
millicentachieng: We want it to be nullable so that if `scopeType` is not specified then we return all scope types.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -319,112 +360,133 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
...
-                // Creates a dict of scopes information from cached files
-                scopesInformationDictionary = await GetOrCreatePermissionsDescriptionsAsync(locale);
-            }
-
-            var scopesInfo = new List<ScopeInformation>();
+            List<ScopeInformation> scopesInfo = new List<ScopeInformation>();
+            List<PermissionError> errors = new List<PermissionError>();
 
-            if (string.IsNullOrEmpty(requestUrl))  // fetch all permissions
+            if (!requestUrls.Any())
```
irvinesunday, line 383: `requestUrls` could be be null here.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -319,112 +360,133 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
...
 
-                    return null;
-                }
+        private IEnumerable<ScopeInformation> GetScopesForRequestUrl(string requestUrl,
+                                                              string method = null,
+                                                              ScopeType? scopeType = null,
+                                                              bool includeHidden = false,
+                                                              bool isLeastPrivilege = false)
+        {
+            if (string.IsNullOrEmpty(requestUrl))
```
irvinesunday, line 452: Consider moving this check to the calling method and check whether the `IEnumerable` of `requestUrls` is null.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -319,112 +360,133 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
...
+                // Get all scopes
+                var allScopes = GetAllScopes(scopeType, includeHidden);
+                scopesInfo.AddRange(allScopes);
             }
-            else // fetch permissions for a given request url and method
+            else
             {
-                if (string.IsNullOrEmpty(method))
+                // Get scopes for multiple request URLs
+                foreach (var url in requestUrls)
```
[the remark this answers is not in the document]
millicentachieng: The thrown exception is handled within the catch block and execution continues to the next URL. And what are the benefits of using a select statement instead?

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -94,79 +93,121 @@ private PermissionsDataInfo PermissionsData
...
+            catch (Exception exception)
+            {
+                _telemetryClient?.TrackException(exception);
+                permissionsDocument = null;
+            }
+            return permissionsDocument;
+        }
 
-                    JToken apiPermissions = permissionsObject.First.First;
+        private async Task<PermissionsDataInfo> LoadPermissionsDataAsync()
```
zengin, line 141: I think we should run this conversion once at the time of permission file deployment and generate the reverse lookup data structure on file.

DevX API should simply load that reverse lookup file as a data structure and run lookup logic per request.

Current architecture has a very long latency for the initial request which can be avoided.

zengin (changes requested):

GraphWebApi/Controllers/PermissionsController.cs
irvinesunday: Resolving this as it has been explained above.

PermissionsService/Services/PermissionsStore.cs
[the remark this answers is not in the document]
baywet: @irvinesunday is there a specific reason why we rolled out our "own version" of the http client factory?

PermissionsService/Services/PermissionsStore.cs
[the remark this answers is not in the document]
baywet: I might be missing some of the UX context here. My assumption was that as an application developer, I want to see all the permissions I might need for my application to run. If there's is a failure, instead of getting only the ones that didn't fail until the first failure, wouldn't it be better to get all the ones that didn't fail at least?

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -94,79 +93,121 @@ private PermissionsDataInfo PermissionsData
...
+        {
+            var permissionsDocument = await LoadPermissionsDocumentAsync();
+            if (permissionsDocument == null)
+                return null;
 
-                    foreach (JProperty property in apiPermissions.OfType<JProperty>())
+            var uriTemplateMatcher = new UriTemplateMatcher();
+            var pathPermissions = new Dictionary<string, PathPermissions>();
+            foreach (var permission in permissionsDocument.Permissions)
```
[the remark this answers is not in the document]
millicentachieng: For performance improvement, we will implement this [idea](https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1402#discussion_r1128667816) from @zengin.

PermissionsService/Services/PermissionsStore.cs
millicentachieng: Here's the link to the reverse lookup file I'll be using 
https://raw.githubusercontent.com/microsoftgraph/microsoft-graph-devx-content/task/ma/add-reverse-lookup-permission-file/permissions/permissions-all.json

PermissionsService/Services/PermissionsStore.cs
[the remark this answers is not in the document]
millicentachieng: Here's a snippet of more of the code
```csharp
foreach (var url in requestUrls)
{
    try
    {
        if (string.IsNullOrEmpty(url))
            throw new InvalidOperationException("The request URL cannot be null or empty.");

        var scopesForUrl = GetScopesForRequestUrl(requestUrl: url,
                                                scopeType: scopeType,
                                                includeHidden: includeHidden,
                                                leastPrivilegeOnly: leastPrivilegeOnly);

        if (scopesForUrl == null)
            throw new InvalidOperationException($"Permissions information for {url} were not found.");

        scopesInfo.AddRange(scopesForUrl);
    }
    catch (Exception exception)
    {
        errors.Add(new PermissionError()
        {
            Url = url,
            Message = exception.Message
        });
    }
}
```
If there's an exception while fetching scopes for any of the Urls, it's caught, added to the errors list and we move on to process the next Url.

PermissionsService/Services/PermissionsStore.cs
millicentachieng: Updated to use reverse lookup file

PermissionsService/Services/PermissionsStore.cs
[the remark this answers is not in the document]
baywet: thanks for clarifying/updating.

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -319,111 +312,146 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
...
+                    }
                 }
+            }
 
-                requestUrl = CleanRequestUrl(requestUrl);
-
-                // Check if requestUrl is contained in our Url Template table
-                TemplateMatch resultMatch = permissionsData.UriTemplateMatcher.Match(new Uri(requestUrl, UriKind.RelativeOrAbsolute));
+            // Get consent display name and description
+            var scopesInfo = GetAdditionalScopesInformation(scopesInformationDictionary, scopes.DistinctBy(x => new { x.ScopeName, x.ScopeType }));
```
baywet, line 373:
```suggestion
            var scopesInfo = GetAdditionalScopesInformation(scopesInformationDictionary, scopes.DistinctBy(x => $"{x.ScopeName}{x.ScopeType}", StringComparison.OrdinalIgnoreCase));
```

Otherwise you're doing reference comparisons on anonymous objects which will always be different

GraphWebApi/Controllers/PermissionsController.cs
```diff
@@ -39,53 +42,81 @@ public PermissionsController(IPermissionsStore permissionsStore, TelemetryClient
...
+
+            _permissionsTraceProperties.Add(UtilityConstants.TelemetryPropertyKey_SanitizeIgnore, nameof(PermissionsController));
+            _telemetryClient?.TrackTrace($"Fetched {result?.Results.Count ?? 0} permissions",
                                             SeverityLevel.Information,
                                             _permissionsTraceProperties);
-                localeCode = supportedLocaleCode;
-            }
 
-            List<ScopeInformation> result = null;
+            return result == null || !result.Results.Any() ? NotFound() : Ok(result.Results);
```
baywet, line 73: why is there a difference in returning status based on the result and presence of errors or not between the two methods?

baywet (changes requested):
Thanks for making the changes, two minor things on my side

PermissionsService/Services/PermissionsStore.cs
baywet: right, this ends up calling [Object.Equals()](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals?view=net-7.0) which can have a bunch of size effects and relies on reflection (bad perf) which recursively does reference comparisons on the object structure.
Whereas a string is deterministic and will provide better perf

GraphWebApi/Controllers/PermissionsController.cs
millicentachieng: Implemented the above

baywet (dismissed):
Thanks for making the changes

PermissionsService.Test/PermissionsService.Test.csproj
```diff
@@ -35,15 +34,12 @@
...
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </Content>
-    <Content Include="TestFiles\permissions-test-file-v1.0_ver1.json">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </Content>
-    <Content Include="TestFiles\permissions-test-file-v1.0_ver2.json">
+    <Content Include="TestFiles\permissions-test-file-empty.json">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
+      <Content Include="TestFiles\permissions-test-file.json">
```
zengin, line 40: nit: you can use wildcards to copy files, e.g. `TestFiles\*`

PermissionsService/Interfaces/IPermissionStore.cs
```diff
@@ -15,21 +15,25 @@ namespace PermissionsService.Interfaces
...
-        /// <returns>A list of <see cref="ScopeInformation"/>.</returns>
-        Task<List<ScopeInformation>> GetScopesAsync(string scopeType = "DelegatedWork",
-                                                    string locale = null,
-                                                    string requestUrl = null,
-                                                    string method = null,
-                                                    string org = null,
-                                                    string branchName = null);
+        /// <param name="branchName">Optional: The name of the branch containing the files</param>
+        /// <returns></returns>
+        Task<PermissionResult> GetScopesAsync(List<string> requestUrls = null,
```
zengin, line 29: I think requestUrls should not be optional.

zengin (approved):

PermissionsService/Interfaces/IPermissionStore.cs
millicentachieng: We want to fetch all permissions when requestUrls is null

PermissionsService.Test/PermissionsService.Test.csproj
millicentachieng: Updated

baywet (approved):

baywet (changes requested):

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -479,33 +504,32 @@ private static string CleanRequestUrl(string requestUrl)
                 throw new ArgumentNullException(nameof(scopes));
             }
 
-            if (string.IsNullOrEmpty(scopeType))
-            {
-                throw new ArgumentException($"'{nameof(scopeType)}' cannot be null or empty.", nameof(scopeType));
-            }
+            var schemeKeys = scopes.Select(x => x.ScopeType).Distinct()
+                .Select(x => x.ToString().Contains(Delegated, StringComparison.InvariantCultureIgnoreCase) ? Delegated : Application)
```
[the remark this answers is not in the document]
millicentachieng: The descriptions list is grouped into two lists: "Delegated" and "Application". We are comparing an enum and a string that do not exactly match.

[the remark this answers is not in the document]
baywet: So you're mapping two values to two other values? A dictionary would be much faster at doing that.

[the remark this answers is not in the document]
baywet: or more accurately 3 enum values to 2 strings. so a Dictionary<ScopeType, string> with the 3 entries as a static readonly field shuold do the trick

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -78,97 +79,83 @@ public UriTemplateMatcher UriTemplateMatcher
...
             _scopesInformation = configuration[ScopesInfoBlobConfig]
                 ?? throw new ArgumentNullException(nameof(ScopesInfoBlobConfig), $"Config path missing: {ScopesInfoBlobConfig}");
             _defaultRefreshTimeInHours = FileServiceHelper.GetFileCacheRefreshTime(configuration[CacheRefreshTimeConfig]);
-
         }
 
-        private PermissionsDataInfo PermissionsData
-        {
-            get
+        private Task<PermissionsDataInfo> PermissionsData =>
```
baywet, line 89: could you log an issue about multiple threads potentially populating the value concurrently and the impact on the initial request. And the suggested solution to load the cache on app startup instead + have something else re-populate it on a schedule.

baywet (approved):

PermissionsService/Services/PermissionsStore.cs
millicentachieng: https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1456

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -479,40 +507,45 @@ private static string CleanRequestUrl(string requestUrl)
...
+                }
+            });
 
             var scopesInfo = scopes.Select(scope =>
             {
-                scopesInformationDictionary[key].TryGetValue(scope, out var scopeInfo);                
-                return scopeInfo ?? new() { ScopeName = scope };
+                permissionDescriptionGroups.TryGetValue((ScopeType)scope.ScopeType, out var schemeKey);
+                scopesInformationDictionary[schemeKey].TryGetValue(scope.ScopeName, out var scopeInfo);
+                if (scopeInfo != null)
```
baywet, line 526:
```suggestion
                if (scopesInformationDictionary[schemeKey].TryGetValue(scope.ScopeName, out var scopeInfo))
```

baywet (changes requested):

PR #1406, opened 2023-03-06
Feature: change open api gen process [Restore]

### LINKED ISSUE #1392: Use pre-generated OpenAPI files

thewahome:
## Overview

Closes #1392 

It is a restoration of the work done in #1393

### DIFF
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -33,7 +33,7 @@ namespace GraphWebApi.Controllers
     {
         private readonly IConfiguration _configuration;
         private readonly IOpenApiService _openApiService;
- 
+
         public OpenApiController(IConfiguration configuration, IOpenApiService openApiService)
         {
             ArgumentNullException.ThrowIfNull(openApiService, nameof(openApiService));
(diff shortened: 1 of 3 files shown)


baywet (approved):

irvinesunday:
Sorry for jumping into this late. I feel like there are a few things that needed to be addressed in this PR. We need to modify some of the telemetry trace messages that were capturing the conversion process. Some of the method functions have evolved past their initial names, e.g.: 
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/ca8a345c89b6dbbb42de3c92afcaec70d7191b7f/OpenAPIService/OpenApiService.cs#L663-L681

We also need to refactor this method and update it to reflect the changed process of retrieving OpenAPI document streams and not converting CSDL streams to OpenAPI docs. The telemetry trace messaging also need to change likewise:
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/ca8a345c89b6dbbb42de3c92afcaec70d7191b7f/OpenAPIService/OpenApiService.cs#L525-L602

We might also need to consider removing the [POST method](https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/ca8a345c89b6dbbb42de3c92afcaec70d7191b7f/GraphWebApi/Controllers/OpenApiController.cs#L143-L163) of the _/openapi_ endpoint. This is not used, and Hidi is doing the same thing anyway. 

Consequently, we will need to remove the below functionalities accordingly:
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/ca8a345c89b6dbbb42de3c92afcaec70d7191b7f/OpenAPIService/OpenApiService.cs#L735-L756

This will only be used by the [OpenAPIDocumentCreatorMock.cs](https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/dev/OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs) class and it can hence be moved there: 
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/ca8a345c89b6dbbb42de3c92afcaec70d7191b7f/OpenAPIService/OpenApiService.cs#L714-L733

Then with the above removal, we can replace this: https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/ca8a345c89b6dbbb42de3c92afcaec70d7191b7f/OpenAPIService/OpenApiService.cs#L157
with 
```CSharp
return new OpenApiDocument(subset);
```

[Nitpick]: Naming convention is to prefix fields with `_`
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/ca8a345c89b6dbbb42de3c92afcaec70d7191b7f/OpenAPIService/OpenApiService.cs#L58

PR #1411, opened 2023-03-08
Feature: change open api gen process 

### LINKED ISSUE #1392: Use pre-generated OpenAPI files

thewahome:
## Overview

Closes #1392 

It is a restoration of the work done in #1393

## Notes

Makes sure all the comments on this [PR comment ](https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1406#issuecomment-1456813052) by @irvinesunday  have been addressed

### DIFF
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -142,6 +142,7 @@ namespace GraphWebApi.Controllers
 
         [Route("openapi")]
         [HttpPost]
+        [Obsolete("No longer supported")]
         public async Task<IActionResult> Post([FromQuery] string operationIds = null,
                                               [FromQuery] string tags = null,
                                               [FromQuery] string url = null,
@@ -153,14 +154,8 @@ namespace GraphWebApi.Controllers
                                               [FromQuery] bool forceRefresh = false,
                                               [FromQuery] bool includeRequestBody = false)
         {
-            var styleOptions = new OpenApiStyleOptions(style, openApiVersion, graphVersion, format);
-
-            var openApiConvertSettings = _openApiService.GetOpenApiConvertSettings(style);
-
-            var source = await _openApiService.ConvertCsdlToOpenApiAsync(Request.Body, openApiConvertSettings);
-
-            return CreateSubsetOpenApiDocument(operationIds, tags, url, source, title, styleOptions, forceRefresh, includeRequestBody);
-        }
+            throw new NotImplementedException();
+        }        
 
         private FileStreamResult CreateSubsetOpenApiDocument(string operationIds, string tags,
                                                              string url, OpenApiDocument source,
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -33,7 +33,7 @@ namespace GraphWebApi.Controllers
     {
         private readonly IConfiguration _configuration;
         private readonly IOpenApiService _openApiService;
- 
+
         public OpenApiController(IConfiguration configuration, IOpenApiService openApiService)
         {
             ArgumentNullException.ThrowIfNull(openApiService, nameof(openApiService));
@@ -46,16 +46,16 @@ namespace GraphWebApi.Controllers
         [Route("$openapi")]
         [HttpGet]
         public async Task<IActionResult> Get(
-                                    [FromQuery]string operationIds = null,
-                                    [FromQuery]string tags = null,
-                                    [FromQuery]string url = null,
-                                    [FromQuery]string openApiVersion = null,
-                                    [FromQuery]string title = "Partial Graph API",
-                                    [FromQuery]OpenApiStyle style = OpenApiStyle.Plain,
-                                    [FromQuery]string format = null,
-                                    [FromQuery]string graphVersion = null,
-                                    [FromQuery]bool includeRequestBody = false,
-                                    [FromQuery]bool forceRefresh = false)
+                                    [FromQuery] string operationIds = null,
+                                    [FromQuery] string tags = null,
+                                    [FromQuery] string url = null,
+                                    [FromQuery] string openApiVersion = null,
+                                    [FromQuery] string title = "Partial Graph API",
+                                    [FromQuery] OpenApiStyle style = OpenApiStyle.Plain,
+                                    [FromQuery] string format = null,
+                                    [FromQuery] string graphVersion = null,
+                                    [FromQuery] bool includeRequestBody = false,
+                                    [FromQuery] bool forceRefresh = false)
         {
             var styleOptions = new OpenApiStyleOptions(style, openApiVersion, graphVersion, format);
 
@@ -72,11 +72,11 @@ namespace GraphWebApi.Controllers
 
         [Route("openapi/operations")]
         [HttpGet]
-        public async Task<IActionResult> Get([FromQuery]string graphVersion = null,
-                                             [FromQuery]string openApiVersion = null,
-                                             [FromQuery]OpenApiStyle style = OpenApiStyle.Plain,
-                                             [FromQuery]string format = null,
-                                             [FromQuery]bool forceRefresh = false)
+        public async Task<IActionResult> Get([FromQuery] string graphVersion = null,
+                                             [FromQuery] string openApiVersion = null,
+                                             [FromQuery] OpenApiStyle style = OpenApiStyle.Plain,
+                                             [FromQuery] string format = null,
+                                             [FromQuery] bool forceRefresh = false)
         {
             var styleOptions = new OpenApiStyleOptions(style, openApiVersion, graphVersion, format);
 
@@ -129,11 +129,11 @@ namespace GraphWebApi.Controllers
             }
 
             var rootNode = _openApiService.CreateOpenApiUrlTreeNode(sources);
-            
+
             Response.ContentType = "application/json";
             Response.StatusCode = 200;
             await Response.StartAsync();
-            
+
             var writer = new Utf8JsonWriter(Response.BodyWriter, new JsonWriterOptions() { Indented = false });
             OpenApiService.ConvertOpenApiUrlTreeNodeToJson(writer, rootNode);
             await writer.FlushAsync();
--- a/OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
+++ b/OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
@@ -2,15 +2,20 @@
 //  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 // ------------------------------------------------------------------------------------------------------------------------------------------------------
 
+using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.Extensions.Configuration;
 using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Readers;
+using Microsoft.OpenApi.Writers;
 using OpenAPIService.Interfaces;
 using System;
 using System.Collections.Concurrent;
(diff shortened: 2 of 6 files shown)


GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -153,14 +154,8 @@ public OpenApiController(IConfiguration configuration, IOpenApiService openApiSe
...
         {
-            var styleOptions = new OpenApiStyleOptions(style, openApiVersion, graphVersion, format);
-
-            var openApiConvertSettings = _openApiService.GetOpenApiConvertSettings(style);
-
-            var source = await _openApiService.ConvertCsdlToOpenApiAsync(Request.Body, openApiConvertSettings);
-
-            return CreateSubsetOpenApiDocument(operationIds, tags, url, source, title, styleOptions, forceRefresh, includeRequestBody);
-        }
+            throw new NotImplementedException();
```
baywet, line 157: I'd support either:
- removing the method all together so callers get a 400 class response and they know they need to update their code
- still replying but setting a [deprecation or sunset header](https://nordicapis.com/how-to-smartly-sunset-and-deprecate-apis/) + adding an artificial delay of 5 seconds so we don't break people right away, they'll have a performance issue, and know they need to upgrade their code.

But throwing an error is just bad API practices.

Do we know from telemetry what applications depend on that endpoint today?

OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
```diff
@@ -38,6 +43,17 @@ public static IOpenApiService GetOpenApiService(string configFilePath)
             return new OpenApiService(configuration);
         }
 
+        private OpenApiDocument CloneOpenApiDocument(OpenApiDocument openApiDocument)
```
baywet, line 46: this method should be async as it calls into a bunch of APIs that could be replaced by async equivalents

OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
```diff
@@ -38,6 +43,17 @@ public static IOpenApiService GetOpenApiService(string configFilePath)
...
+        {
+            var stream = new MemoryStream();
```
baywet, line 48:
```suggestion
            await using var stream = new MemoryStream();
```

OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
```diff
@@ -38,6 +43,17 @@ public static IOpenApiService GetOpenApiService(string configFilePath)
...
+            var writer = new OpenApiYamlWriter(new StreamWriter(stream));
```
baywet, line 49: get the stream writer out and dispose it (await using)

OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
```diff
@@ -38,6 +43,17 @@ public static IOpenApiService GetOpenApiService(string configFilePath)
...
+            openApiDocument.SerializeAsV3(writer);
+            writer.Flush();
```
baywet, line 51: should use flushasync here instead

OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
```diff
@@ -38,6 +43,17 @@ public static IOpenApiService GetOpenApiService(string configFilePath)
...
+            stream.Position = 0;
+            var reader = new OpenApiStreamReader();
+            return reader.Read(stream, out _); ;
```
baywet, line 54: use the read async method instead

OpenAPIService/OpenApiService.cs
```diff
@@ -585,7 +594,7 @@ public async Task<OpenApiDocument> GetGraphOpenApiDocumentAsync(string graphUri,
                     _graphUriQueue.Dequeue();
                 }
                 
-                _openApiDocumentConversionAccess.Release();
+                _openApiDocumentAccess.Release();
                 _telemetryClient?.TrackTrace("Conversion lock successfully released.",
```
baywet, line 598:
```suggestion
                _telemetryClient?.TrackTrace("Fetch lock successfully released.",
```

OpenAPIService/OpenApiService.cs
```diff
@@ -517,19 +523,19 @@ public async Task<OpenApiDocument> GetGraphOpenApiDocumentAsync(string graphUri,
             }
 
             /* OpenAPI document not cached;
-             * Ensure only one thread is running the conversion process.
+             * Ensure only one thread is running the fetch process.
              */
-            await _openApiDocumentConversionAccess.WaitAsync();
+            await _openApiDocumentAccess.WaitAsync();
```
baywet, line 528: don't implement locking/caching yourself. Just use an InMemoryCache and use the [GetOrCreateAsync](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.caching.memory.cacheextensions.getorcreateasync?view=dotnet-plat-ext-7.0) method.
It's thread safe because designed for aspnet scenarios.
We could even imagine using a background service populating the cache so requests don't hang.
https://learn.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-7.0

OpenAPIService/OpenApiService.cs
```diff
@@ -651,101 +660,26 @@ public OpenApiDocument ApplyStyle(OpenApiStyle style, OpenApiDocument subsetOpen
             return subsetOpenApiDocument;
         }
 
-        private async Task<OpenApiDocument> CreateOpenApiDocumentAsync(Uri csdlHref, OpenApiStyle openApiStyle)
+        private async Task<OpenApiDocument> GetOpenApiDocumentAsync(Uri openAPIHref)
         {
             var stopwatch = new Stopwatch();
             var httpClient = CreateHttpClient();
```
baywet, line 666: Instantiating http client randomly without disposing them WILL lead to port exhaustion.
Please have the constructor accept and set as a field an IHttpClientFactory instead and configure the dependency injection to setup the client for you.
https://learn.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-7.0

Please coordinate with @millicentachieng on that work since I've made a similar comment in #1402

OpenAPIService/OpenApiService.cs
```diff
@@ -651,101 +660,26 @@ public OpenApiDocument ApplyStyle(OpenApiStyle style, OpenApiDocument subsetOpen
...
 
             _openApiTraceProperties.TryAdd(UtilityConstants.TelemetryPropertyKey_SanitizeIgnore, nameof(OpenApiService));
-            _telemetryClient?.TrackTrace($"Success getting CSDL for {csdlHref} in {stopwatch.ElapsedMilliseconds}ms",
+            _telemetryClient?.TrackTrace($"Success getting OpenAPI document for {openAPIHref} in {stopwatch.ElapsedMilliseconds}ms",
                                          SeverityLevel.Information,
                                          _openApiTraceProperties);
             _openApiTraceProperties.TryRemove(UtilityConstants.TelemetryPropertyKey_SanitizeIgnore, out string _);
 
-            OpenApiDocument document = await ConvertCsdlToOpenApiAsync(csdl, GetOpenApiConvertSettings(openApiStyle));
+            OpenApiDocument document = new OpenApiStreamReader().Read(stream, out var diagnostic);
```
baywet, line 678: use the read async alternative

OpenAPIService/OpenApiService.cs
```diff
@@ -651,101 +660,26 @@ public OpenApiDocument ApplyStyle(OpenApiStyle style, OpenApiDocument subsetOpen
...
 
-        private async Task<OpenApiDocument> CreateOpenApiDocumentAsync(Uri csdlHref, OpenApiStyle openApiStyle)
+        private async Task<OpenApiDocument> GetOpenApiDocumentAsync(Uri openAPIHref)
         {
             var stopwatch = new Stopwatch();
             var httpClient = CreateHttpClient();
 
             stopwatch.Start();
-            Stream csdl = await httpClient.GetStreamAsync(csdlHref.OriginalString);
+            using Stream stream = await httpClient.GetStreamAsync(openAPIHref.OriginalString);
```
baywet, line 669:
```suggestion
            await using Stream stream = await httpClient.GetStreamAsync(openAPIHref.OriginalString);
```

baywet (changes requested):

GraphWebApi/Controllers/OpenApiController.cs
irvinesunday: We can't tell the applications calling the endpoint from the telemetry, but there are about 78 calls to this endpoint in the last 30 days, most of them ending with errors 500 or 404:

image

The last one being on 8th March 2023

baywet: so in other words, we're pretty sure GE is not depending on this anymore at this point? other experiences might be impacted but their are outside of our control and impact should be low/non-existent

OpenAPIService/OpenApiService.cs
thewahome: This, though beneficial, is outside of the scope of this PR since the work was targetted to fetching OpenAPI documents from their locations. Embarking on this would need it's own issue and I have created it here: https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1420

OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
thewahome: Similar to: https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1411#discussion_r1133771346

OpenAPIService.Test/OpenAPIDocumentCreatorMock.cs
thewahome: Similar to: https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1411#discussion_r1133771346

baywet (approved):

OpenAPIService/OpenApiService.cs
```diff
@@ -50,19 +50,25 @@ public class OpenApiService : IOpenApiService
                         new();
         private readonly TelemetryClient _telemetryClient;
         private readonly IConfiguration _configuration;
```
baywet, line 52:
```suggestion
```

(this is only used to read a single value in the constructor, update the constructor too)

OpenAPIService/OpenApiService.cs
```diff
@@ -651,99 +660,23 @@ public OpenApiDocument ApplyStyle(OpenApiStyle style, OpenApiDocument subsetOpen
             return subsetOpenApiDocument;
         }
 
-        private async Task<OpenApiDocument> CreateOpenApiDocumentAsync(Uri csdlHref, OpenApiStyle openApiStyle)
+        private async Task<OpenApiDocument> GetOpenApiDocumentAsync(Uri openAPIHref)
```
baywet, line 663:
```suggestion
        private static async Task<OpenApiDocument> GetOpenApiDocumentAsync(Uri openAPIHref)
```

baywet (changes requested):

OpenAPIService/OpenApiService.cs
thewahome: The configuration field is needed so that the service can surface the contents of the CacheRefreshTime setting. 
Is the ask to remove the whole process completely?

baywet: no, simply to remove the field, and the field affectation in the constructor.
the value can then be read directly from the field.

OpenAPIService/OpenApiService.cs
baywet: right, I'm not sure why sonarcloud has been suggesting that.

OpenAPIService/OpenApiService.cs
thewahome: I am not sure what you mean by read directly from the parameter. Might this be another overoptimisation from sonarcloud?

thewahome: done

baywet (approved):
thanks for making the changes. There are conflicts to resolve before this can be merged.

PR #1454, opened 2023-03-24
Adds query param to override the default OpenAPI file name for a given style

irvinesunday:
Currently PowerShell uses the endpoint http://devxapitest.azurewebsites.net/ to test for v2. This endpoint uses [powershell_v2.yaml](https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/openapi/v1.0/powershell_v2.yaml) for its OpenAPI document. This document has been converted using PowerShell v2 specific [OpenAPIConvert settings.](https://github.com/microsoftgraph/msgraph-metadata/blob/master/conversion-settings/powershell_v2.json) DevX API prod uses [powershell.yaml](https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/openapi/v1.0/powershell.yaml). 
Instead of supporting 2 different endpoints for PowerShell cmdlet generation, this query param `fileName` can be used for both scenarios:
- PowerShell v1: https://graphexplorerapi.azurewebsites.net/openapi/operations?style=PowerShell
- PowerShell v2: https://graphexplorerapi.azurewebsites.net/openapi/operations?style=PowerShell&fileName=powershell_v2

Both documents will also be cached separately in the DevX API. 
This fix will allow us to switch off the devxapi test endpoint when not in use and help reduce COGS.

### DIFF
--- a/GraphWebApi/Controllers/OpenApiController.cs
+++ b/GraphWebApi/Controllers/OpenApiController.cs
@@ -1,4 +1,4 @@
-// ------------------------------------------------------------------------------------------------------------------------------------------------------
+﻿// ------------------------------------------------------------------------------------------------------------------------------------------------------
 //  Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 // ------------------------------------------------------------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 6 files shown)


baywet (changes requested):
letting people set the filename through API parameters without doing any validation on the filename is a risky thing security wise. We should at least validate for slash, backslash and other characters that are invalid for URLs...

thewahome (dismissed):

GraphWebApi/Controllers/OpenApiController.cs
```diff
@@ -44,17 +44,18 @@ public OpenApiController(IConfiguration configuration, IOpenApiService openApiSe
...
+                                    [FromQuery] string url = null,
+                                    [FromQuery] string openApiVersion = null,
+                                    [FromQuery] string title = "Partial Graph API",
+                                    [FromQuery] OpenApiStyle style = OpenApiStyle.Plain,
+                                    [FromQuery] string format = null,
+                                    [FromQuery] string graphVersion = null,
+                                    [FromQuery] bool includeRequestBody = false,
+                                    [FromQuery] bool forceRefresh = false,
+                                    [FromQuery] bool singularizeOperationIds = false,
+                                    [FromQuery] string fileName = null)
```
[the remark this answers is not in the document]
irvinesunday: I've added URL-safety validation as suggested by Vincent.

UtilityService/UtilityFunctions.cs
```diff
@@ -33,5 +34,21 @@ public static string CheckArgumentNullOrEmpty(string value, string parameterName
...
+        /// <param name="input">The input string.</param>
+        /// <returns>True whether the input string does not contain URL-safe characters.</returns>
+        public static bool IsUrlSafe(string input)
+        {
+            if (string.IsNullOrEmpty(input))
+            {
+                return true;
+            }
+
+            const string pattern = @"[^a-zA-Z0-9\-._~]";
```
baywet, line 50: move the regex as a static readonly regex property with a compiled flag or as a generated regex so we don't pay the cost of parsing the regex for each call. And include a timeout for the parsing.

baywet (changes requested):

UtilityService/UtilityFunctions.cs
irvinesunday: Went with option 1

baywet (approved):
Thanks for making the changes

Issue #1458, opened 2023-03-28
`/openapi/operations` has inconsistencies in the output when `style=PowerShell`


irvinesunday:
The endpoint `/openapi/operations` experiences inconsistent results.

Whenever you call the endpoint initially, the operations load as shown below:

![image](https://user-images.githubusercontent.com/40403681/228380189-66caed88-a6a0-480b-a647-aa590b54c1ca.png)

However, when you select the OpenAPI link for `admin.serviceAnnouncement.CreateHealthOverviews` for example and reload the url: https://graphexplorerapi.azurewebsites.net/openapi/operations?style=powerShell, you notice the operation id of the above link changes to: `admin.serviceAnnouncement_CreateHealthOverviews` as illustrated below:

![image](https://user-images.githubusercontent.com/40403681/228380097-dc24bf40-ea83-41d6-b243-b8ceaf0b2ff4.png)

This is because we are using copy constructors [here](https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/c87b77c4fcd46019fdc341afafeb4461ffd5196d/OpenAPIService/OpenApiService.cs#L156), however, the copy constructors has a missing capability of copying over (creating a new instance) of the `OpenApiPathItems` for the `OpenApiPaths` object. This issue is being tracked [here](https://github.com/microsoft/OpenAPI.NET/issues/1193). 
We need to use cloning to support creating new instances of an existing OpenAPI document in the interim. 
Cloning needs to be re-introduced as we will be returning copies of the doc. which could be altered at the calling code like [here](https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/c87b77c4fcd46019fdc341afafeb4461ffd5196d/OpenAPIService/OpenApiService.cs#L156) or we could be adding style-specific modifications that may alter the original doc. like in [here](https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/c87b77c4fcd46019fdc341afafeb4461ffd5196d/OpenAPIService/OpenApiService.cs#L622).

irvinesunday:
Resolved with PR: https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1463

PR #1461, opened 2023-03-29
Replaces copy constructors with cloning and adds missing param

irvinesunday:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1458

This PR:
- Replaces use of copy constructors with cloning of an OpenAPI document. Copy constructors has a missing capability that will be addressed in this [issue](https://github.com/microsoft/OpenAPI.NET/issues/1193).
- Adds missing parameter `fileName` to the html output of `/openapi/operations`.


OpenAPIService/OpenApiService.cs
```diff
@@ -801,6 +799,19 @@ private static void RemoveContent(OpenApiDocument target)
...
+        
+        /// <summary>
+        /// Creates a clone of an OpenAPI document
+        /// </summary>
+        /// <param name="document">The source document to clone.</param>
+        /// <returns>A clone of the source document.</returns>
+        public OpenApiDocument CloneOpenApiDocument(OpenApiDocument document)
+        {
+            var sb = new StringBuilder();
+            document.SerializeAsV3(new OpenApiYamlWriter(new StringWriter(sb)));
```
baywet, line 811: the string writer needs to be disposed even though it maps to a string builder.
Also why use strings when we could directly use streams?

irvinesunday:
Need to abandon this PR and create a new one that merges into `master` branch. `dev` has some permissions work that is still not ready to get into production.

OpenAPIService/OpenApiService.cs
irvinesunday: Resolved it as such:

https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/e47bbf257fc6c021e4c17f5c98d94d0e3007e168/OpenAPIService/OpenApiService.cs#L808-L817

Issue #1453, opened 2023-03-24
Remap trimmed /drive paths in snippet generation


andrueastman:
Related to https://github.com/microsoftgraph/msgraph-metadata/issues/248 and https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1762

As the openAPI metadata does not have child paths related to /drive paths we should probably remap known paths to known alternatives during snippet generation. 

e.g. `me/drive/root/delta` to `drives/{drive-id}/items/root/delta` 

This will reduce missing snippets due to the failed path matching and should provide benefit for all languages.

PR #1468, opened 2023-04-03
Remap known trimmed paths in snippet generation

andrueastman:
This PR closes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1453

It introduces a KnownReMappings collection that remaps paths trimmed from the original openApi description to alternate paths so that inputs to the snippet generation result in outputs and not failures and should result in increase numbers for snippet generation across all openAPI based languages.

e.g. a snippet with the path `/me/drive/root/` will be remapped to the path `/drives/driveId/items/root/` path to ensure doc examples and sample responses on GE.

Tests have been added to validate this

Other changes include 
- Fixes path matching for functions with parameters `/range(address='A1:B2')` for excel functions
- Fixes a bug in dotnet generation where empty collections in the payload would result in failed type lookup. 

Sample update generation can viewed at 
https://github.com/microsoftgraph/microsoft-graph-docs/compare/main...preview-snippet-generation/111959 

and run at https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=111959&view=ms.vss-test-web.build-test-results-tab

Missing snippets/failed generation should reduce by up to 50% in V1 353 -> 171 while 515 ->308 in beta in comparison with the latest run.

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/SnippetModel.cs
+++ b/CodeSnippetsReflection.OpenAPI/SnippetModel.cs
@@ -36,7 +36,8 @@ namespace CodeSnippetsReflection.OpenAPI
         {
             if (treeNode == null) throw new ArgumentNullException(nameof(treeNode));
 
-            var splatPath = ReplaceIndexParametersByPathSegment(requestPayload.RequestUri
+            var remappedPayload = RemapKnownPathsIfNeeded(requestPayload);
+            var splatPath = ReplaceIndexParametersByPathSegment(remappedPayload.RequestUri
                                         .AbsolutePath
                                         .TrimStart(pathSeparator))
                                         .Split(pathSeparator, StringSplitOptions.RemoveEmptyEntries)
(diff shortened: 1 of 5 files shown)


CodeSnippetsReflection.OpenAPI/SnippetModel.cs
```diff
@@ -36,14 +36,44 @@ public SnippetModel(HttpRequestMessage requestPayload, string serviceRootUrl, Op
...
                                         .TrimStart(pathSeparator))
                                         .Split(pathSeparator, StringSplitOptions.RemoveEmptyEntries)
                                         .Skip(1); //skipping the version
             LoadPathNodes(treeNode, splatPath);
             InitializeModel(requestPayload);
         }
+
+        private static readonly Dictionary<Regex, string> KnownReMappings = new()
+        {
+            { new Regex(@"/me/drive/root/",RegexOptions.Compiled, TimeSpan.FromMilliseconds(200)), "/drives/driveId/items/root/" },
```
baywet, line 51: would you like to use capturing groups to be able to conserve the Ids?

baywet: ah, right and because you're coming from /me you don't have an ID so you're adding the generic one. Got it.

PR #1479, opened 2023-04-07
Updates operationIds deduplication logic

### LINKED ISSUE #1475: [PowerShell] Review `operationId`s deduplication logic
The deduplication logic for `operationId`s wrongly considers certain words as duplicates of the previous segments. For example, `print.taskDefinitions.tasks_GetTrigger` is resolved as `print.taskDefinition_GetTrigger` instead of `print.taskDefinition.task_GetTrigger`.

Let's consider doing a whole word lookup for repeating words instead of a regex match. Alternatively, we can consider excluding duplicate words removal since this is handled by AutoREST. We only need to singularize the segments.

### LINKED ISSUE #1476: [PowerShell] Add singularization exception for `data`
We should consider adding a singularization exception for `data`. `Data` is currently singularized as `datum`, which causes command name change (breaking change). For example, operationId `external.industryData_ListDataConnectors` is singularized as `external.industryDatum_ListDataConnectors`.

The exception can be added via:
``` c#
Vocabularies.Default.AddSingular("(data)$", "$1");
```

peombwa:
This PR fixes #1475 and fixes #1476 by:
- Updates deduplication logic to use whole word lookup instead of contains when looking for repeating words.
- Updates deduplication logic to only consider a segment name duplicate if exists in the previous segment.
- Excludes `data` from being singularized.


OpenAPIService/PowershellFormatter.cs
```diff
@@ -219,28 +219,25 @@ private static string SingularizeAndDeduplicateOperationId(string operationId)
 
             // drives does not properly singularize to drive.
             Vocabularies.Default.AddSingular("(drive)s$", "$1");
-            
+            // exclude the following from singularization.
+            Vocabularies.Default.AddSingular("(data)$", "$1");
+
             var segments = operationId.Split('.').ToList();
+            var segementsCount = segments.Count;
```
irvinesunday, line 226: [Nitpick]
```suggestion
            var segmentsCount = segments.Count;
```

irvinesunday: And update the other places referenced

irvinesunday:
Otherwise, it LGTM.

OpenAPIService/PowershellFormatter.cs
peombwa: Thanks for taking a look. I traced this to a vs extension that wrongly "auto-corrects" the word. Let me correct the PR.

irvinesunday (approved):

PR #1496, opened 2023-04-12
Omitted unsupported workbook paths from powershell snippet generation…

timayabi2020:
## Overview

This PR removes unsupported workbook Api paths from PowerShell snippet generation report.  Support was intentionally removed because of AutoREST timeouts caused by expansion of workbook navigation property on driveItem.  Having those paths reduces PowerShell snippet generation pass rates.

### Before omission
**Beta:** 791/3305 failed
**Pass rate:** 76% 

**V1:** 486/2152 failed
 **Pass rate:** 77% 

image

### After omission
**Beta:** 611/3305 failed 
**Pass rate:** 81.5% 

**V1:** 301/2152 failed 
**Pass rate:** 86% 
image

## Testing Instructions

* Run snippet generation pipeline on this branch and compare results with main branch.

### DIFF
--- a/CodeSnippetsPipeline.Test/CodeSnippetsPipeline.cs
+++ b/CodeSnippetsPipeline.Test/CodeSnippetsPipeline.cs
@@ -56,7 +56,17 @@ public class CodeSnippetsPipeline
                     var message = "Original HTTP Snippet:" + Environment.NewLine + 
                         File.ReadAllText(httpSnippetFilePath).TrimStart() + Environment.NewLine +
                         File.ReadAllText(expectedLanguageSnippetErrorFileFullPath);
-                    Assert.Fail(message);
+                    if(language.Equals("powershell", StringComparison.OrdinalIgnoreCase))
+                    {
+                        if(!message.Contains("workbook"))
+                        {
+                            Assert.Fail(message);
+                        }
+                    }
+                    else
+                    {
+                        Assert.Fail(message);
+                    }
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


peombwa (approved):

zengin:
@peombwa @timayabi2020

This will show in our reports as if we are generating PowerShell snippets for these paths, which is not correct. We need an alternative approach.

We are also building a funnel graph, so this will fatten the snippet generation output in that funnel, but now the diff between available snippets and executable snippets will be bigger.

We should figure out a systematic way to highlight this scenario differently.

peombwa:
@zengin, I thought Raptor was calling the unit test in this PR when generating the reports as suggested by the PR description.

I do agree with finding an alternative approach. Raptor should consider having an exclude list which excludes known paths from tests. This can be a Regex based json config that maps to OpenAPI paths/tags/operationIds.

timayabi2020:
@zengin Maybe I'm missing something because the idea behind this PR was to exclude those workbook paths that are not supported from the snippet generation report for PowerShell.  Shouldn't we report on the paths that an SDK supports?

zengin:
@timayabi2020, if we want to add those paths eventually, I am not sure if we should only report on the paths that SDK supports. It is better to highlight the debt instead of hiding it. Maybe we need a way to mark them (similar to known issue categories in Raptor tests).

If we think that we will never add those paths, then maybe we can modify the baseline. However, this PR does not exclude the test cases from the baseline. The test cases still take HTTP snippets as the base. The test cases for workbook are generated, and they pass, which implies that we are able to generate those snippets for PowerShell.

One option we have is that we can remove them from the test discovery, however, I still believe that having a separate known issue category is a good idea if we would like to add those paths later.

https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/4250e240f9d7d4f5b8b8002ab0633a9b6f8706a8/CodeSnippetsPipeline.Test/CodeSnippetsPipeline.cs#L18-L24

PR #1537, opened 2023-05-05
Move singularization rules to static constructor

millicentachieng:
## Overview
This is to fix timeout issue when singularizing operation Ids.

### DIFF
--- a/OpenAPIService/PowershellFormatter.cs
+++ b/OpenAPIService/PowershellFormatter.cs
@@ -33,6 +33,12 @@ namespace OpenAPIService
             _singularizeOperationIds = singularizeOperationIds;
         }
 
+        static PowershellFormatter()
+        {
+            Vocabularies.Default.AddSingular("(drive)s$", "$1"); // drives does not properly singularize to drive.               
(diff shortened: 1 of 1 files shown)


OpenAPIService/PowershellFormatter.cs
```diff
@@ -33,6 +33,12 @@ public PowershellFormatter(bool singularizeOperationIds)
             _singularizeOperationIds = singularizeOperationIds;
         }
 
+        static PowershellFormatter()
+        {
+            Vocabularies.Default.AddSingular("(drive)s$", "$1"); // drives does not properly singularize to drive.               
+            Vocabularies.Default.AddSingular("(data)$", "$1"); // exclude the following from singularization.
+        }
```
peombwa, line 40: Can we also add exclusions to the singularization rule for the following words:
```suggestion
            Vocabularies.Default.AddSingular("(delta)$", "$1");
            Vocabularies.Default.AddSingular("(quota)$", "$1");
            Vocabularies.Default.AddSingular("(statistics)$", "$1");
        }
```

peombwa (approved):

irvinesunday (approved):

PR #1559, opened 2023-05-17
Update response for endpoint used to fetch permissions for a set of request URLs

millicentachieng:
## Overview
The update is to ensure consistent behaviour regardless of the content of the response.

```
{
  "results": [
    {
      "value": "string",
      "scopeType": "ScopeType",
      "consentDisplayName": "string",
      "consentDescription": "string",
      "isAdmin": "string",
      "leastPrivilegeOnly": "bool"
    }
  ],
  "errors": [
    {
      "requestUrl": "string",
      "message": "string"
    }
  ]
}
```

### DIFF
--- a/PermissionsService/Services/PermissionsStore.cs
+++ b/PermissionsService/Services/PermissionsStore.cs
@@ -89,8 +89,9 @@ namespace PermissionsService
         private Task<PermissionsDataInfo> PermissionsData =>
             _cache.GetOrCreateAsync("PermissionsData", async entry =>
             {
-                entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromHours(_defaultRefreshTimeInHours);
-                return await LoadPermissionsDataAsync();
+                var permissionsData = await LoadPermissionsDataAsync();
+                entry.AbsoluteExpirationRelativeToNow = permissionsData is not null ? TimeSpan.FromHours(_defaultRefreshTimeInHours) : TimeSpan.FromSeconds(1);
+                return permissionsData;
             });
 
         /// <summary>
@@ -127,7 +128,7 @@ namespace PermissionsService
                 foreach (var key in fetchedPermissions.Keys)
                 {
                     // Remove any '(...)' from the request url and set to lowercase for uniformity
-                    string requestUrl = key.RemoveParentheses().RenameDuplicatePlaceholders().ToLower();
+                    string requestUrl = key.RemoveParentheses().ToLower();
 
                     count++;
 
--- a/PermissionsService/Services/PermissionsStore.cs
+++ b/PermissionsService/Services/PermissionsStore.cs
@@ -127,7 +127,7 @@ namespace PermissionsService
                 foreach (var key in fetchedPermissions.Keys)
                 {
                     // Remove any '(...)' from the request url and set to lowercase for uniformity
-                    string requestUrl = key.RemoveParentheses().ToLower();
+                    string requestUrl = key.RemoveParentheses().RenameDuplicatePlaceholders().ToLower();
 
                     count++;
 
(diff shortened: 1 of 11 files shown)


PermissionsService/Services/PermissionsStore.cs
```diff
@@ -89,8 +89,9 @@ public UriTemplateMatcher UriTemplateMatcher
         private Task<PermissionsDataInfo> PermissionsData =>
             _cache.GetOrCreateAsync("PermissionsData", async entry =>
             {
-                entry.AbsoluteExpirationRelativeToNow = TimeSpan.FromHours(_defaultRefreshTimeInHours);
-                return await LoadPermissionsDataAsync();
+                var permissionsData = await LoadPermissionsDataAsync();
+                entry.AbsoluteExpirationRelativeToNow = permissionsData is not null ? TimeSpan.FromHours(_defaultRefreshTimeInHours) : TimeSpan.FromSeconds(1);
```
andrueastman, line 93: This means that we may occasionally be putting null values into the cache, which may be hit (as much as short expiration time is set) in that short duration.
Any reason we can't use a `TryGetValue` then call `LoadPermissionsDataAsync` and `cache. Set` when it returns false?

UriMatchService/UriTemplateMatcher.cs
```diff
@@ -122,6 +136,36 @@ public TemplateMatch Match(Uri uri)
             return match.Success ? parameters : null;
         }
 
+
+        /// <summary>
+        /// Renames placeholder in path by appending number to occurrence
+        /// </summary>
+        /// <param name="value">The target string value.</param>
+        /// <returns>The string without duplicated placeholder names.</returns>
+        private string RenameDuplicatePlaceholders(string value)
```
andrueastman, line 145:
```suggestion
        private static string RenameDuplicatePlaceholders(string value)
```

UtilityService/UtilityExtensions.cs
```diff
@@ -3,6 +3,8 @@
 // -------------------------------------------------------------------------------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
```
andrueastman, line 6: Unused using statements added.

UriMatchService/UriTemplateMatcher.cs
```diff
@@ -122,6 +136,36 @@ public TemplateMatch Match(Uri uri)
...
+            MatchCollection matches = placeholderRegex.Matches(value);
+
+            foreach (var match in matches.Cast<Match>())
+            {
+                var placeholder = match.Groups[1].Value;
+                if (!placeholderCounts.ContainsKey(placeholder))
+                    placeholderCounts.Add(placeholder, 0);
+                int count = ++placeholderCounts[placeholder];
+                var newPlaceholder = $"{{{placeholder}{count}}}";
+                Regex regex = new Regex(Regex.Escape($"{{{placeholder}}}"));
```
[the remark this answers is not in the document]
millicentachieng: I've removed the method.

UriMatchService/UriTemplateMatcher.cs
```diff
@@ -31,6 +31,8 @@ public class UriTemplateMatcher
     {
         private readonly Dictionary<string, string> _templates = new Dictionary<string, string>();
 
+        private static readonly Regex placeholderRegex = new Regex(@"\{(.*?)\}", RegexOptions.Compiled);
```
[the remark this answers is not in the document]
millicentachieng: Resolved.

UriMatchService/UriTemplateMatcher.cs
```diff
@@ -17,6 +17,7 @@
 using System;
 #pragma warning restore S125 // Sections of code should not be commented out
 using System.Collections.Generic;
+using System.Drawing;
```
irvinesunday, line 20: Unecessary using

PermissionsService/Services/PermissionsStore.cs
andrueastman: Just checking on this one @millicentachieng.
Otherwise, the PR looks good to me.

millicentachieng: @andrueastman We will need to call `LoadPermissionsDataAsync` synchronously if we choose that path, which is something we want to avoid. I've set the cache expiry to 1 millisecond.

andrueastman (approved):

PR #1603, opened 2023-06-19
Fix: Use pascal case for issue property key

### LINKED ISSUE #1602: Use PascalCase case for knownissues
The content returned from the endpoint is camelCased, and we expect it to be PascalCased on the portal

Onokaev:
## Overview
Fixes #1602 

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/GraphWebApi/Controllers/KnownIssuesController.cs
+++ b/GraphWebApi/Controllers/KnownIssuesController.cs
@@ -44,7 +44,7 @@ namespace GraphWebApi.Controllers
 
             List<KnownIssue> result = await _knownIssuesService.QueryBugsAsync();
 
-            return result == null ? NotFound() : Ok(JsonConvert.SerializeObject(result));
+            return result == null ? NotFound() : Ok(result);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


GraphWebApi/Controllers/KnownIssuesController.cs
```diff
@@ -42,7 +43,8 @@ public async Task<IActionResult> GetKnownIssues()
                                             _knownIssuesTraceProperties);
 
             List<KnownIssue> result = await _knownIssuesService.QueryBugsAsync();
-            return result == null ? NotFound() : Ok(result);
+
+            return result == null ? NotFound() : Ok(JsonConvert.SerializeObject(result));
```
andrueastman, line 47: Because of this line in the project config
https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/ff1f170bd821fddf4f6800f537e01a8a9ec6c677/GraphWebApi/Startup.cs#L85

I suspect this file can remain unchanged after adding the property attributes and there is no need to call `JsonConvert.SerializeObject`. 
```suggestion
            return result == null ? NotFound() : Ok(result);
```

baywet (approved):

andrueastman (approved):

PR #1611, opened 2023-06-21
Use parallel forEach to fetch permissions for a set of request URLs

millicentachieng:
## Overview
This PR attempts to reduce the time it takes to fetch permissions for a large set of request URLs but doing parallel lookup.

### DIFF
--- a/PermissionsService/Services/PermissionsStore.cs
+++ b/PermissionsService/Services/PermissionsStore.cs
@@ -5,7 +5,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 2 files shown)


PermissionsService/Services/PermissionsStore.cs
```diff
@@ -3,7 +3,9 @@
 // -------------------------------------------------------------------------------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
```
andrueastman, line 8: Unused import

PermissionsService/Services/PermissionsStore.cs
```diff
@@ -332,8 +334,10 @@ private async Task<string> FetchHttpSourceDocument(string sourceUri)
             }
             else
             {
-                var scopesByRequestUrl = new Dictionary<string, IEnumerable<ScopeInformation>>();
-                foreach (var request in requests)
+                var scopesByRequestUrl = new ConcurrentDictionary<string, IEnumerable<ScopeInformation>>();
+                var uniqueRequests = requests.DistinctBy(static x => $"{x.HttpMethod}{x.RequestUrl}", StringComparer.OrdinalIgnoreCase);
+
+                Parallel.ForEach(uniqueRequests, async request =>
```
andrueastman, line 340:
```suggestion
                ```suggestion
                await Parallel.ForEachAsync(uniqueRequests, async (request,_) =>
```
As the lambda passed is `async`

andrueastman (approved):

PR #1634, opened 2023-06-27
Handle URL template matching exceptions

### LINKED ISSUE #1633: Wrap URL template matching within try catch block to gracefully handle errors in permissions file

millicentachieng:
## Overview
Fixes #1633

### DIFF
--- a/UriMatchService/UriTemplateMatcher.cs
+++ b/UriMatchService/UriTemplateMatcher.cs
@@ -69,13 +69,20 @@ namespace UriMatchingService
             var templateMatches = new Dictionary<KeyValuePair<string, string>, int>(); // {{ templateKey, foundParameterCount }}
             foreach (var template in _templates)
             {
-                var parameters = GetParameters(absolutePath, template.Value);
-                if (parameters != null)
+                try
                 {
-                    if (parameters.Count == 0)
-                        return new TemplateMatch() { Key = template.Key, Template = template.Value }; // exact match, no ids
-                    else 
-                        templateMatches.Add(template, parameters.Count);
+                    var parameters = GetParameters(absolutePath, template.Value);
+                    if (parameters != null)
+                    {
+                        if (parameters.Count == 0)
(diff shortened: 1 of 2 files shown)


UriMatchService/UriTemplateMatcher.cs
```diff
@@ -69,13 +69,20 @@ public TemplateMatch Match(Uri uri)
...
-                        return new TemplateMatch() { Key = template.Key, Template = template.Value }; // exact match, no ids
-                    else 
-                        templateMatches.Add(template, parameters.Count);
+                    var parameters = GetParameters(absolutePath, template.Value);
+                    if (parameters != null)
+                    {
+                        if (parameters.Count == 0)
+                            return new TemplateMatch() { Key = template.Key, Template = template.Value }; // exact match, no ids
+                        else
+                            templateMatches.Add(template, parameters.Count);
```
rkodev, line 80: we should log this exception as a warning if its a none critical error

andrueastman, line 80:
```suggestion
                        else if(!templateMatches.ContainsKey(template))
                            templateMatches.Add(template, parameters.Count);
```
If we are simply avoiding the throwing of the `ArgumentExcpetion` we could check if the key exists to avoid the exception handling overhead here.

irvinesunday: @millicentachieng what example scenarios throw exceptions?

millicentachieng: @irvinesunday Like this one `/employeeexperience/learningproviders/{id}/learningcourseactivities(extern`. There's no closing bracket.

irvinesunday: Then this makes sense. 
Perhaps we could incorporate the other suggested additions to make this more robust?

andrueastman: Just to confirm, do we still need the try catch block after the change? @millicentachieng

PR #1644, opened 2023-06-27
Filter out private issues

### LINKED ISSUE #1643: Private known issues should be filtered out
Private issues are showing on the knownissues page

Onokaev:
## Overview
Fixes #1643 

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description

### DIFF
--- a/KnownIssuesService/Models/KnownIssue.cs
+++ b/KnownIssuesService/Models/KnownIssue.cs
@@ -81,5 +81,13 @@ namespace KnownIssuesService.Models
         {
             get; set;
         }
+
+        /// <summary>
+        /// Microsoft Graph Issue Visibility
+        /// </summary>
+        public Boolean IsPublicIssue
+        {
+            get; set;
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


KnownIssuesService/Models/KnownIssue.cs
```diff
@@ -64,8 +64,8 @@ public record KnownIssue
         /// <summary>
         /// Determines if the Last Update is current compared to the Created date
         /// </summary>
-        [JsonProperty(nameof(IsUpdated))]
-        public bool IsUpdated => LastUpdatedDateTime > CreatedDateTime;
+        [JsonProperty(nameof(IsDateUpdated))]
```
baywet, line 67: why would you want to serialize a computed property?

baywet (approved):

KnownIssuesService/Models/KnownIssue.cs
Onokaev: This was a mistake. I'll create a new PR to fix it

PR #1681, opened 2023-07-24
Fixes functions with parameters in Go

rkodev:
## Overview

Fixes generation fluent path in go for functions with paramerts e.g https://learn.microsoft.com/en-us/graph/api/reportroot-getyammeractivitycounts?view=graph-rest-1.0&tabs=http

Current behaviour
```go
graphClient.Reports().GetYammerActivityCounts(period='{period}')().Get(context.Background(), nil)
```

expected behaviour

```go
period = '{period}'
graphClient.Reports().GetYammerActivityCountsWithPeriod(&period).Get(context.Background(), nil)
```

## Testing Instructions

* Unit tests

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
+++ b/CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
@@ -7,6 +7,7 @@ using CodeSnippetsReflection.OpenAPI.ModelGraph;
 using CodeSnippetsReflection.StringExtensions;
 using Microsoft.OpenApi.Services;
 using System.Text.RegularExpressions;
+using System.Collections;
 
 namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators
 {
@@ -79,7 +80,7 @@ namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators
             var models = getModelsPaths(codeGraph);
             foreach (var path in models)
             {
-                builder.AppendLine($"\t  graph{path.Replace(".","").ToLowerInvariant()} \"github.com/microsoftgraph/{apiVersion}/{path.Replace(".", "/").ToLowerInvariant()}\"");
+                builder.AppendLine($"\t  graph{path.Replace(".", "").ToLowerInvariant()} \"github.com/microsoftgraph/{apiVersion}/{path.Replace(".", "/").ToLowerInvariant()}\"");
             }
 
             builder.AppendLine("\t  //other-imports"); // models version
@@ -97,7 +98,7 @@ namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators
             }
             return nameSpaces;
         }
-        
+
         /// <summary>
         /// Returns a list of all the namespaces that are referenced in the body.
         /// </summary>
@@ -292,7 +293,7 @@ namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators
             // if the first element is a collection index skip it
             var isCollection = nodes.First().Segment.IsCollectionIndex();
             var isSingleElement = nodes.Count() == 1;
-            
+
             var filteredNodes = (isCollection && !isSingleElement) ? nodes.Skip(2) : isCollection ? nodes.Skip(1) : nodes; // skip first element if its not only element
             if (!(filteredNodes?.Any() ?? false)) return string.Empty;
             return filteredNodes.Select(static x =>
@@ -327,7 +328,7 @@ namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators
             if ((!String.IsNullOrWhiteSpace(Name) && !Name.Substring(1).Equals("\"", StringComparison.OrdinalIgnoreCase)) && (Name.Contains('.') || Name.Contains('-')))
             {
                 var propertyMatch = PropertyNameRegex.Match(Name);
-                return propertyMatch.Success ? string.Join("",propertyMatch.Groups[1].Value.Split(".").Select(static x => x.ToFirstCharacterUpperCase())).ToFirstCharacterLowerCase() : $"\"{Name}\"";
+                return propertyMatch.Success ? string.Join("", propertyMatch.Groups[1].Value.Split(".").Select(static x => x.ToFirstCharacterUpperCase())).ToFirstCharacterLowerCase() : $"\"{Name}\"";
             }
 
             return Name;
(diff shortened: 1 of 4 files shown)


CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
```diff
@@ -559,6 +567,14 @@ private static void WriteCodePropertyObject(string propertyAssignment, StringBui
                 WriteCodeProperty(propertyAssignment, builder, codeProperty, child, indentManager, childPosition++);
         }
 
+        private static Dictionary<String, String> GetPathVariables(IEnumerable<OpenApiUrlTreeNode> nodes)
+        {
+            return nodes.Select(x => x.Segment.GetFunctionWithParameters())
```
baywet, line 572:
```suggestion
            return nodes.Select(static x => x.Segment.GetFunctionWithParameters())
```

CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
```diff
@@ -559,6 +567,14 @@ private static void WriteCodePropertyObject(string propertyAssignment, StringBui
...
+                .Aggregate(new Dictionary<String, String>(), (first, second) => first.Concat(second.Item3)
+                  .GroupBy(kvp => kvp.Key, kvp => kvp.Value)
+                  .ToDictionary(g => g.Key, g => g.Last()));
```
baywet, line 575:
```suggestion
                .Aggregate(new Dictionary<String, String>(), static (first, second) => first.Concat(second.Item3)
                  .GroupBy(static kvp => kvp.Key, static kvp => kvp.Value)
                  .ToDictionary(static g => g.Key, static g => g.Last()));
```

baywet (changes requested):

CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
```diff
@@ -573,7 +589,16 @@ private static string GetFluentApiPath(IEnumerable<OpenApiUrlTreeNode> nodes)
                     return x.Segment.Split('.')
                                     .Select(static s => s.ToFirstCharacterUpperCase())
                                     .Aggregate(static (a, b) => $"{a}{b}") + "().";
-                return x.Segment.ToFirstCharacterUpperCase() + "().";
+                else
+                {
+                    var funcWithParams = x.Segment.GetFunctionWithParameters();
```
andrueastman, line 594: I think all you need to do here is call `x.Segment.IsFunctionWithParameters()`(already defined) then look up the parameters and values in `snippetCodeGraph.PathParameters` as 
- Functions with multiple parameters could exists
- The `snippetCodeGraph.PathParameters` in contains type information that you can use to know for scenarios where the parameter is a number/Date type
- Casing is consistent with the metadata therefore generating names that are aligned with the generated SDK.

rkodev: Thanks for this. One concern with the Path Parameter is we can have an example using 1 or 2 params instead of all the path parameters. Thats the reason I opted for extracting the parameters from the snippet. I however do agree that the path param will give better type information

CodeSnippetsReflection.OpenAPI/StringExtensions.cs
```diff
@@ -1,14 +1,19 @@
...
 using System.Text.RegularExpressions;
 using CodeSnippetsReflection.StringExtensions;
 
 namespace CodeSnippetsReflection.OpenAPI;
 
-internal static class StringExtensions 
+internal static class StringExtensions
 {
+
+    private static readonly Regex FunctionWithParams = new(@"^(\w+)\(([^)]+)\)$", RegexOptions.Compiled, TimeSpan.FromMilliseconds(200));
```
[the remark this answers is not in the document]
rkodev: The need here was to extract the parameters used in the example

CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
andrueastman: If a path had 3 parameters, and the example would use only 2 parameters, that would mean that one of the parameters is optional or has a pre-defined default. In such a case i believe the metadata would either 
- have two distinct paths (one with the optional param and the other without). 
- one path but the default would still need to be generated/passed as Kiota will still generate the parameter name in the function https://github.com/microsoft/kiota/issues/2584

At this point of the snippet generation, the example path would already have been matched to the path that is relevant to it (when the snippet model is built) and every path parameter (including the optional ones if matched with their defaults) would be needed to be generated as only one url template is present and a parameter must be passed/included in function name. 

Using the http URL would generate a function that does not exist in the SDK in such a scenario as the default parameter would be missing/not passed.

CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
rkodev: done

CodeSnippetsReflection.OpenAPI/StringExtensions.cs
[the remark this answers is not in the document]
rkodev: I've combined these 2 since the existing regex both identify a function syntax. The difference is the new regex splits the function into "name" + "params" and make it simpler to return function names.

CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
rkodev: The only left concern would be is it a given than the function with params is always the last segment of a path? This approach ensures that the params retain the position in the segment

andrueastman (changes requested):
Latest updates break some tests in PHP/C#/Poweshell.

Any chance we can 
- Fix the broken test.
- Add a test for Go for the test thats broken. Should point to the snippet at https://learn.microsoft.com/en-us/graph/api/callrecords-callrecord-getpstncalls?view=graph-rest-1.0&tabs=http

CodeSnippetsReflection.OpenAPI/LanguageGenerators/GoGenerator.cs
```diff
@@ -573,7 +581,17 @@ private static string GetFluentApiPath(IEnumerable<OpenApiUrlTreeNode> nodes)
...
+                    if (!x.Segment.IsFunctionWithParameters())
```
andrueastman, line 587: This can be done first before `var funcWithParams = x.Segment.GetFunctionWithParameters();` as it assures us that results exist.

andrueastman (approved):

PR #1690, opened 2023-07-27
Fixes missing snippets using alternate alternate keys.

andrueastman:
Examples with alternate paths fail to get generates or are not correctly generates as the alternate path keys are not added to the `pathParameters` collection in the Snippet code graph. The snippets are therefore generated with incorrect/missing arguments or fail.

This PR unblocks snippet generation in the doc pages such as 
https://learn.microsoft.com/en-us/graph/api/application-get?view=graph-rest-1.0&tabs=csharp#example-2-retrieve-an-application-by-its-appid-and-only-specific-properties


CodeSnippetsReflection.OpenAPI/ModelGraph/SnippetCodeGraph.cs
```diff
@@ -226,8 +226,13 @@ private static List<CodeProperty> parsePathParameters(SnippetModel snippetModel)
                                              .SelectMany(static pathItem => pathItem.Value.Operations)
                                              .Where(operation => operation.Key.ToString().Equals(snippetModel.Method.ToString(), StringComparison.OrdinalIgnoreCase)) // get the operations that match the method
                                              .SelectMany(static operation => operation.Value.Parameters)
-                                             .Where(static parameter => parameter.In == ParameterLocation.Path); // find the parameters in the path
+                                             .Where(static parameter => parameter.In == ParameterLocation.Path).ToList(); // find the parameters in the path
```
baywet, line 229: you could use a union instead

baywet (approved):

PR #1759, opened 2023-09-19
chore: enable CodeQL, update branch protection policy, add issue automation

MIchaelMainer:
## Overview

Improve branch protection by requiring conversation resolution (signoff), require strict status checks including CodeQL, and add issue automation. 

## Testing Instructions

After merge into dev:
1. Open a new test PR and witness that there are now required status checks, all PR comments require resolution.
2. Create test issue and set the Resolution: Duplicate label on it. It should be automatically closed. This only tests the one action and the bot integration is working.

### DIFF
--- a/.github/policies/microsoft-graph-devx-api-branch-protection.yml
+++ b/.github/policies/microsoft-graph-devx-api-branch-protection.yml
@@ -40,7 +40,7 @@ configuration:
     - DevX API Validate PR
     - CodeQL
     # Require branches to be up to date before merging. Requires requiredStatusChecks. boolean
-    requiresStrictStatusChecks: true
+    requiresStrictStatusChecks: false
     # Indicates whether there are restrictions on who can push. boolean. Requires whoCanPush.
     restrictsPushes: false
     # Restrict who can dismiss pull request reviews. boolean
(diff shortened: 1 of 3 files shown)


.github/policies/microsoft-graph-devx-api-branch-protection.yml
```diff
@@ -31,15 +31,16 @@ configuration:
...
     # Are merge commits prohibited from being pushed to this branch. boolean
     requiresLinearHistory: true
     # Required status checks to pass before merging. Values can be any string, but if the value does not correspond to any existing status check, the status check will be stuck on pending for status since nothing exists to push an actual status
     requiredStatusChecks:
     - license/cla
     - DevX API Validate PR
+    - CodeQL
     # Require branches to be up to date before merging. Requires requiredStatusChecks. boolean
-    requiresStrictStatusChecks: false
+    requiresStrictStatusChecks: true
```
baywet, line 43: can we pause the strictChecks roll out until we figure out a model for dependabot please?

baywet (changes requested):

.github/policies/microsoft-graph-devx-api-branch-protection.yml
MIchaelMainer: Thank you for reminding me about this! Yes, paused. Let me get back to my write-up......

MIchaelMainer: We don't need to have this set to true. I think we should make this setting true by default, and if the maintainer finds that this setting is more trouble than it is worth, then we can disable. For dependabot scenarios, we may want to consider a less frequent update cycle for non-securoty updates.

baywet (approved):

PR #1764, opened 2023-09-29
Rsh/flagged enums for java snippets

ramsessanchez:
Flagged Enum support for Java Snippets. 
``` java 
//Previously setting a flagged enum property would be generated as:
setProperty(EnumClass.Enum1, EnumClass.Enum2);
//This fix ensures that flagged enums are handled as sets so we generate the following instead:
setProperty(EnumSet.of(EnumClass.Enum1, EnumClass.Enum2));
```
Added a isFlagsEnum boolean property to the CodeProperty class in order to aid in generation. This also adds a check in the SnippetCodeGraph class in order to set this boolean in the CodeProperty instance it creates. 

Also fixes a bug where sub-namespaces of models would be incorrectly generated on return types. Sub-namespaces need to be fully qualified to disambiguate.
``` Java
//In the prior snippets return types of models, under the security namespace for example, would be shown as:
security.Alert result = graphClient.something().get();
//The fix ensures that model subnamespaces are fully qualified so now we generate:
com.microsoft.graph.models.security.Alert result = graphClient.something().get();
```

Latest raptor run shows 94% pass rate:
https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=125851&view=ms.vss-test-web.build-test-results-tab

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1764)


CodeSnippetsReflection.OpenAPI/LanguageGenerators/JavaGenerator.cs
```diff
@@ -5,6 +5,8 @@
 using CodeSnippetsReflection.OpenAPI.ModelGraph;
 using CodeSnippetsReflection.StringExtensions;
 using Microsoft.OpenApi.Services;
+using Microsoft.VisualBasic;
+using static System.Runtime.InteropServices.JavaScript.JSType;
```
andrueastman, line 9:
```suggestion
```

CodeSnippetsReflection.OpenAPI/LanguageGenerators/JavaGenerator.cs
```diff
@@ -142,6 +146,25 @@ private static string GetRequestConfiguration(SnippetCodeGraph snippetCodeGraph,
...
+            {
+                var namespaceSegments = codeGraph.ResponseSchema?.Reference?.Id?.Replace($"{DefaultNamespace}", "com.microsoft.graph.models").Split('.', StringSplitOptions.RemoveEmptyEntries);
+
+                if (namespaceSegments != null)
+                {
+                    for (int i = 0; i < namespaceSegments.Length; i++)
+                    {
+                        namespaceSegments[i] = (i < namespaceSegments.Length - 1) ? namespaceSegments[i].ToLowerInvariant() : namespaceSegments[i].ToPascalCase();
+                    }
+                    return namespaceSegments?.Length > 5 ? $"{string.Join(".", namespaceSegments)}" : $"{namespaceSegments?.Last()}";
```
andrueastman, line 161:
```suggestion
                    return namespaceSegments.Length > 5 ? $"{string.Join(".", namespaceSegments)}" : $"{namespaceSegments.Last()}";
```

We already know `namespaceSegments` is not null

andrueastman (approved):
👍🏼

PR #1905, opened 2024-01-18
Add Generic and python imports

shemogumbe:
## Overview

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1905)

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
+++ b/CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
@@ -17,7 +17,7 @@ namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators
         private const string RequestBodyVarName = "request_body";
         private const string RequestConfigurationVarName = "request_configuration";
         private const string RequestParametersPropertyName = "query_params";
-        
+
         private static readonly HashSet<string> ReservedTypeNames = new(StringComparer.OrdinalIgnoreCase)
         {
             "and",
@@ -59,7 +59,7 @@ namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators
             "yield",
             "property",
         };
-        
+
         public string GenerateCodeSnippet(SnippetModel snippetModel)
         {
             var indentManager = new IndentManager();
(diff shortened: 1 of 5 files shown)


CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
```diff
@@ -69,8 +69,49 @@ public string GenerateCodeSnippet(SnippetModel snippetModel)
...
+        {
+            const string modelImportPrefix = "from msgraph.generated.models";
+            const string requestBuilderImportPrefix = "from msgraph.generated";
+
+            HashSet<string> snippetImports = new HashSet<string>();
+
+            snippetImports.Add("from msgraph import GraphServiceClient");
+
+            var importsGenerator = new ImportsGenerator();
+            var imports = importsGenerator.GenerateImportTemplates(snippetModel);
```
andrueastman, line 86: The call to `GenerateImportTemplates` ends up creating another `SnippetCodeGraph` instance internally. Since this is already available, it may make sense to change the function signature to take in `SnippetCodeGraph` as parameter and pass the value avaialble here directly.
Alternatively, we can call `GenerateImportTemplates` internally in the `SnippetCodeGraph` and have the results available as property of the `SnippetCodeGraph`

CodeSnippetsReflection.OpenAPI/LanguageGenerators/SnippetImports.cs
```diff
@@ -0,0 +1,80 @@
...
+using CodeSnippetsReflection.OpenAPI.ModelGraph;
+using CodeSnippetsReflection.StringExtensions;
+using Microsoft.OpenApi.Services;
+
+namespace CodeSnippetsReflection.OpenAPI.LanguageGenerators;
+
+public class ImportsGenerator
+{
+    public List<Dictionary<string, string>> imports = new();
+    public List<Dictionary<string, string>> GenerateImportTemplates(SnippetModel snippetModel)
```
andrueastman, line 14: I think we should create a class(model) with the known keys(Path,RequestBuilderName, Name,TypeDefinition ) as properties rather than using `Dictionary<string, string>` here to avoid the key checking in the caller.

CodeSnippetsReflection.OpenAPI/LanguageGenerators/SnippetImports.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                var path = codeGraph.Nodes.Last().Path;
+                var pathParts = path.Split('/');
+                var lastPathPart = pathParts.Last();
+                if (lastPathPart != null)
+                {
+                    requestBuilderName = $"{requestBuilderName}";
+                    imports.Add(new Dictionary<string, string>
+                    {
+                        { "Path", Regex.Replace(path.Replace("\\", ".").Replace("()", ""), @"\{[^}]*-id\}", "item", RegexOptions.None, TimeSpan.FromSeconds(60))},
```
andrueastman, line 36: We should probably create a static class property(so we don't keep instantiating a new one) for the regex and pass the compile flag too. 
Similar to https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/723913bef8fed8a109697ec51e367ea5d3c2f1fb/TelemetryService/CustomPIIFilter.cs#L30
```suggestion
                        { "Path", Regex.Replace(path.Replace("\\", ".").Replace("()", ""), @"\{[^}]*-id\}", "item", RegexOptions.Compiled, TimeSpan.FromSeconds(60))},
```

CodeSnippetsReflection.OpenAPI/LanguageGenerators/SnippetImports.cs
shemogumbe: The keys differ for models/ request builder/ request_body etc.
With the class approach how do we achieve this, coz at the point of calling, its the keys that help us know if we  have a model, request builder etc

CodeSnippetsReflection.OpenAPI/LanguageGenerators/SnippetImports.cs
shemogumbe: This is a one off instance, its not used elsewhere, would that still require a static class?

CodeSnippetsReflection.OpenAPI/LanguageGenerators/SnippetImports.cs
shemogumbe: For models - we return a different set of key value pairs from what we return for Request_builder, same for request_body

CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
```diff
@@ -69,8 +69,47 @@ public string GenerateCodeSnippet(SnippetModel snippetModel)
...
+            foreach (var import in imports)
+            {
+                if (import.TryGetValue("NamespaceName", out var namespaceName) && string.IsNullOrEmpty(namespaceName))
+                    continue;
+                if (import.TryGetValue("NamespaceName", out var nameSpaceName) && nameSpaceName.Contains("models") && import.TryGetValue("TypeDefinition", out var typeDefinition) && typeDefinition != null)                {
+                    snippetImports.Add($"{modelImportPrefix}.{import["TypeDefinition"].ToSnakeCase()} import {import["TypeDefinition"]}");
```
SilasKenneth, line 92:
```suggestion
                    snippetImports.Add($"{modelImportPrefix}.{typeDefinition.ToSnakeCase()} import {typeDefinition}");
```

CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
```diff
@@ -69,8 +69,47 @@ public string GenerateCodeSnippet(SnippetModel snippetModel)
...
+                    if (import.TryGetValue("Name", out var Name) && !string.IsNullOrEmpty(Name))
+                    {
+                        var namespaceParts = NamespaceName.Split('.').Select((s, i) => i == NamespaceName.Split('.').Length - 1 ? s.ToSnakeCase() : s.ToLowerInvariant());
+                        var importString = $"{requestBuilderImportPrefix}.{string.Join(".", namespaceParts)}.{Name.ToSnakeCase()} import {Name}";
+                        snippetImports.Add(importString.Replace(".me.", ".users.item."));
+                    }
+                }
+                if (import.TryGetValue("Path", out var path) && path != null && import.TryGetValue("RequestBuilderName", out var requestBuilderName) && requestBuilderName != null)                {
+                    //construct path to request builder
+                    snippetImports.Add($"{requestBuilderImportPrefix}{import["Path"].Replace(".me.", ".users.item.")}.{import["RequestBuilderName"].ToSnakeCase()} import {import["RequestBuilderName"]}");
```
SilasKenneth, line 105:
```suggestion
                    snippetImports.Add($"{requestBuilderImportPrefix}{path.Replace(".me.", ".users.item.")}.{requestBuilderName.ToSnakeCase()} import {requestBuilderName}");
```

CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
```diff
@@ -69,8 +69,47 @@ public string GenerateCodeSnippet(SnippetModel snippetModel)
...
+            var importStatements = GetImportStatements(snippetModel);
+            snippetBuilder.Insert(0, string.Join(Environment.NewLine, importStatements));
             return snippetBuilder.ToString();
         }
+        private static HashSet<string> GetImportStatements(SnippetModel snippetModel)
+        {
+            const string modelImportPrefix = "from msgraph.generated.models";
+            const string requestBuilderImportPrefix = "from msgraph.generated";
+
+            HashSet<string> snippetImports = new HashSet<string>();
```
SilasKenneth, line 81:
```suggestion
           var snippetImports = new HashSet<string>();
```

CodeSnippetsReflection.OpenAPI/LanguageGenerators/SnippetImports.cs
```diff
@@ -0,0 +1,102 @@
...
+                    { "TypeDefinition", node.TypeDefinition },
+                    { "NamespaceName", node.NamespaceName },
+                    {"PropertyType", node.PropertyType.ToString()},
+                    {"Value", node.Value}
+
+                };
+            imports.Add(import);
+        }
+
+        if (node.Children != null && node.Children.Count > 0)
```
SilasKenneth, line 89:
```suggestion
        if (node.Children is { Count: > 0 })
```

SilasKenneth (changes requested):

SilasKenneth (commented):
Looks good just a few suggestions.

samwelkanda (approved):

PR #2047, opened 2024-05-14
Fix model import paths + Add support for enum imports

SilasKenneth:
## Overview

### Demo

### Notes

## Testing Instructions

* How to test this PR
* Prefer bulleted description
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/2047)

### DIFF
--- a/CodeSnippetsReflection.OpenAPI/LanguageGenerators/PhpGenerator.cs
+++ b/CodeSnippetsReflection.OpenAPI/LanguageGenerators/PhpGenerator.cs
@@ -138,20 +138,34 @@ public class PhpGenerator : ILanguageGenerator<SnippetModel, OpenApiUrlTreeNode>
             switch (import.Kind)
             {
                 case ImportKind.Model:
+                    if (import.ModelProperty.PropertyType is PropertyType.DateOnly or PropertyType.TimeOnly)
+                    {
+                        snippetImports.Add($"{customTypesPrefix}\\{GetPropertyTypeName(import.ModelProperty)};");
+                        continue;
+                    }
                     var typeDefinition = import.ModelProperty.TypeDefinition;
                     const string modelsNamespaceName = "models.microsoft.graph";
                     var modelNamespaceStringLen = modelsNamespaceName.Length;
+                    var modelNamespace = import.ModelProperty.NamespaceName;
+                    var inModelsNamespace =
+                        modelNamespace.Equals(modelsNamespaceName,
+                            StringComparison.OrdinalIgnoreCase);
+                    var nested = !inModelsNamespace && modelNamespace.StartsWith(modelsNamespaceName);
                     // This takes care of models in nested namespaces inside the model namespace for instance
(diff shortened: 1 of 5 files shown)


CodeSnippetsReflection.OpenAPI/LanguageGenerators/PhpGenerator.cs
```diff
@@ -567,4 +612,14 @@ private static string GetFluentApiPath(IEnumerable<OpenApiUrlTreeNode> nodes, Sn
 
         return result.EndsWith("()()") ? result[..^2] : result;
     }
+
+    private static string GetPropertyTypeName(CodeProperty property)
+    {
```
shemogumbe, line 617: Looks good, How will this work for Languages that do not import time, I think PHP does not need to import time and date classes

CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
```diff
@@ -78,32 +78,65 @@ public string GenerateCodeSnippet(SnippetModel snippetModel)
         }
         private static HashSet<string> GetImportStatements(SnippetModel snippetModel)
         {
-            const string modelImportPrefix = "from msgraph.generated.models";
-            const string requestBuilderImportPrefix = "from msgraph.generated";
+            var packageName = snippetModel.ApiVersion switch
+            {
+                "v1.0" => "msgraph",
```
shemogumbe, line 83: Wouldn't this be confusing for users in graph explorer and docs, I think leaving it as it was is a better idea, then for our case in Raptor, we use GraphAPI version to get the prefix.
This is because there is no docs/GE snippets for beta/v1, Will GE and docs automatically fall back to V1 prefix?

CodeSnippetsReflection.OpenAPI/LanguageGenerators/PhpGenerator.cs
SilasKenneth: I believe it's just Python that doesn't use a class to handle dates and instead uses strings. For PHP, we have custom `DateOnly` and `TimeOnly` types in the abstractions package.

CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
SilasKenneth: This shouldn't be an issue for GE since the snippets shown there are generated by DevX API based on the url the user has passed in Graph Explorer ( _try checking `go` snippets when you change API version on graph explorer_), which always contains the ApiVersion.
Just always showing imports from v1.0 won't always be correct since some classes exist in the Beta API but not in v1.0 API.

shemogumbe (approved):

Issue #1937, opened 2024-02-08
[Python] Align snippet generation to specification.


andrueastman:
Specification for snippets should align to the specification available at https://gist.github.com/sebastienlevert/ae40eb334c3d5a9d4cd3f4a06d2b09be and shared document.

In summary, we should update snippets to

- [x]  Have necessary import statements
- [x]  Remove client initialization code
- [x]  Add comments with links
- [x]  Display version of SDK the snippet supports.

shemogumbe:
This was related to work on snippets on DevX api which is done:
Request adapter was removed
Snippets were added
comments and version were added

PR #2173, opened 2024-08-24
Add  comments on client initialization and supported Version

shemogumbe:
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1937

## Overview

graph_client is currently referenced in snippets with no information on how to initialize it.

### Demo

before the line referencing graphcliet, a python comment is added saying

```python
# To initialize your graph_client, see https://learn.microsoft.com/en-us/graph/sdks/create-client?from=snippets&tabs=python

```
## Snippet Update to:
```python

# Code snippets are only available for the latest version. Current version is 1.x
from msgraph import GraphServiceClient
from msgraph.generated.users.users_request_builder import UsersRequestBuilder
from kiota_abstractions.base_request_configuration import RequestConfiguration
# To initialize your graph_client, see https://learn.microsoft.com/en-us/graph/sdks/create-client?from=snippets&tabs=python


query_params = UsersRequestBuilder.UsersRequestBuilderGetQueryParameters(
		filter = "accountEnabled eq false",
)

request_configuration = RequestConfiguration(
query_parameters = query_params,
)

result = await graph_client.users.get(request_configuration = request_configuration)

```


andrueastman (dismissed):

CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
```diff
@@ -67,13 +71,16 @@ public string GenerateCodeSnippet(SnippetModel snippetModel)
         {
             var indentManager = new IndentManager();
             var codeGraph = new SnippetCodeGraph(snippetModel);
-            var snippetBuilder = new StringBuilder($"{Environment.NewLine}{Environment.NewLine}" +
+            var snippetBuilder = new StringBuilder();
+            snippetBuilder.AppendLine(InitializationInfoString);
+            snippetBuilder.AppendLine($"{Environment.NewLine}{Environment.NewLine}" +
                                                    $"{ClientVarName} = {ClientVarType}({CredentialVarName}, {ScopesVarName}){Environment.NewLine}{Environment.NewLine}");
```
Ndiritu, line 77: Should we remove the graph client instantiation line here?

andrueastman: You're right. This should be dropped to match the spec.

CodeSnippetsReflection.OpenAPI/LanguageGenerators/PythonGenerator.cs
```diff
@@ -67,13 +71,14 @@ public string GenerateCodeSnippet(SnippetModel snippetModel)
         {
             var indentManager = new IndentManager();
             var codeGraph = new SnippetCodeGraph(snippetModel);
-            var snippetBuilder = new StringBuilder($"{Environment.NewLine}{Environment.NewLine}" +
-                                                   $"{ClientVarName} = {ClientVarType}({CredentialVarName}, {ScopesVarName}){Environment.NewLine}{Environment.NewLine}");
```
andrueastman, line 71: Couple of things
- Some of these variables are no longer used.
- Check the tests that are broken due to this change.

andrueastman: Could we also remove the unused variables like `ClientVarName` and `ClientVarType` as well?

shemogumbe: ClientVarName is referenced in the request execution path :
```
            string pathSegment = $"{ClientVarName}.{GetFluentApiPath(codeGraph.Nodes, codeGraph)}";
```


I have removed scopes, credentials and varType

andrueastman (approved):
This may also break raptor validation as it may expect the client initialization code to build/validate. 
We should probably confirm that before merging this. Otherwise, looks good from my end.

shemogumbe:
> This may also break raptor validation as it may expect the client initialization code to build/validate. We should probably confirm that before merging this. Otherwise, looks good from my end.

Thought the same too, I will need to check if Raptor has a client initiated, if not, instatiate one within the snippets in Raptor, will keep this unmerged till I fix that

apidoctor
Ndiritu, line 1: nit; conflicts

Ndiritu, line 1: Ideally we shouldn't have changes to this submodule in this PR.

Ndiritu (changes requested):

apidoctor
andrueastman: +1

Ndiritu (approved):

PR #2372, opened 2025-05-03
Locking optimizations

MarkCiliaVincenti:
Using a dictionary-based approach to locking with hardcoded strings is not optimal, hence switched to use `AsyncNonKeyedLocker`.

However in the case of GetOrCreatePermissionsDescriptionsAsync, it feels like two concurrent threads can exist as long as they have a different locale, hence the change to this method too.


MarkCiliaVincenti:
Set lock at locale level also on SamplesStore.

PR #2375, opened 2025-05-20
fix: a bug where slicing would lead to invalid documents due to missing tag definitions

baywet:
related https://github.com/microsoft/kiota/pull/6481

### DIFF
--- a/OpenAPIService/OpenApiService.cs
+++ b/OpenAPIService/OpenApiService.cs
@@ -144,6 +144,12 @@ namespace OpenAPIService
                 throw new ArgumentException("No paths found for the supplied parameters.");
             }
 
+            var tagIds = subset.Paths.Values.SelectMany(static x => (x.Operations ?? []).Values.SelectMany(static y => y.Tags ?? []))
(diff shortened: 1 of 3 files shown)


OpenAPIService/OpenApiService.cs
```diff
@@ -144,6 +144,12 @@ public OpenApiDocument CreateFilteredDocument(OpenApiDocument source, string tit
                 throw new ArgumentException("No paths found for the supplied parameters.");
             }
 
+            var tagIds = subset.Paths.Values.SelectMany(static x => (x.Operations ?? []).Values.SelectMany(static y => y.Tags ?? []))
```
MaggieKimani1, line 147:
```suggestion
            var tagIds = subset.Paths.Values.SelectMany(static x => (x.Operations ?? new Dictionary<OperationType, OpenApiOperation>()).Values.SelectMany(static y => y.Tags ?? []))
```

MaggieKimani1 (changes requested):

baywet:
nevermind, this is actually valid.

